PR #78, opened 2022-03-11
[safety] Reject high rounds in block headers

### LINKED ISSUE #46: [safety] Reject high rounds in block headers
The `sanitize_header` function processes all headers received by a primary:
https://github.com/MystenLabs/narwhal/blob/03e91f82c358c09990fd282bd0dc3eceec86292d/primary/src/core.rs#L328-L340
It should reject blocks with a claimed round much higher than the current block round.

This block round is managed using an `AtomicU64` in the core's fields:
https://github.com/MystenLabs/narwhal/blob/03e91f82c358c09990fd282bd0dc3eceec86292d/primary/src/core.rs#L44

A simple but large integer interval between a received header's claimed round and the current round should be enough to avoid spam (which is the main goal of this issue). Determining what "large enough" looks like will require analyzing how fast the block round evolves over time. The idea is that a small gap in round numbers is acceptable (though the header will not be voted on, it is not invalid, and post verification should contribute to an increase of the node's own round number).

The error produced here should clearly indicate the cause of the rejection, and produce a new variant in `DagError`:
https://github.com/MystenLabs/narwhal/blob/03e91f82c358c09990fd282bd0dc3eceec86292d/primary/src/error.rs#L27
Indeed at worst, this could indicate the local authority is an out-of-sync node that hasn't caught up with the network.

It should also emit logs, ideally with context allowed since the switch to tracing (#26).

akichidis:
resolves: MystenLabs/narwhal#46 

This PR introduces a check on the Core component when the header is sanitised. It basically checks whether a received header has a round greater than a defined threshold when compared to local round value. This threshold is defined via the introduced parameter `max_header_round_offset` . I chose to make this a parameter to be easily configurable since this could be different across deployments.


asonnino (commented):
I am not sure why the changes affect safety (indicated in the title). I however do not understand why these changes do not break liveness

akichidis:
We discussed with @asonnino offline about the liveness lock concerns and it seems that we might want to tackle this problem in a different manner to ensure we don't break liveness assumptions that research guarantees. On high level, an adversary that can slow down an honest node which implements the proposed fix, can bring the honest node to a point that will stop operating (acknowledging headers) and basically break the liveness assumptions and make our system susceptible to attacks.

Instead, we reduced the problem to receiving headers for which we don't manage to get their parents certificates (because the header is basically junk). Keep receiving such headers can lead to getting swamped by messages from potentially dishonest nodes and practically slow down our node. In this case we might want to throttle the peer node and ensure that we put a cap on the number of pending headers (for which we don't have their partner certificates). So if the received headers are junk (very high round number without 2f+1 parents) , we'll eventually stop receiving headers from that peer node until the buffer is cleaned up (this won't affect our ability to receive and process headers from rest of peers). We can probably have mechanisms to eventually clean up / GC those buffers as well. The upside of this method is that we don't break liveness guarantees only because we observe headers with high round numbers (which could be our node catching up) and we limit suspicious activity. If the peer node is honest and we happen to eventually receive those certificates, the buffer will get cleared up and we'll continue processing headers from that node again.

Opened issue MystenLabs/sui#5197

PR #102, opened 2022-03-27
Clients support

asonnino:
This PR is part of our story to allow Tusk/Bullshark to support light clients.

### What this PR does:
* Implementation of persistent storage for the consensus. Consensus should now be able to crash-recover safely, but there is still an issue to fix to make it live (https://github.com/MystenLabs/sui/issues/5200). We should support crash-recover for both the consensus node and the light clients.
* Add a subscriber module accepting connections from clients and serving them the consensus output sequence.
* Implement a synchroniser helping slow/crash-recovered clients to get up to speed.

### In a next PR:
* Limit the number of subscribers a node can handle (DoS resistance): https://github.com/MystenLabs/narwhal/issues/109
* Rather than dropping connections with clients when things go wrong, engage in a helpful dialogue: https://github.com/MystenLabs/narwhal/issues/107
* Finish the last bit to make consensus' crash-recovery live: https://github.com/MystenLabs/sui/issues/5200
* Make a facility to serve block contents to subscribers (not only certificates)

PR #106, opened 2022-03-28
Clients support

asonnino:
This PR is part of our story to allow Tusk/Bullshark to support light clients.

### What this PR does:
* Implementation of persistent storage for the consensus. Consensus should now be able to crash-recover safely, but there is still an issue to fix to make it live (https://github.com/MystenLabs/sui/issues/5200). We should support crash-recover for both the consensus node and the light clients.
* Add a subscriber module accepting connections from clients and serving them the consensus output sequence.
* Implement a synchroniser helping slow/crash-recovered clients to get up to speed.

### In a next PR:
* Limit the number of subscribers a node can handle (DoS resistance): https://github.com/MystenLabs/narwhal/issues/109
* Rather than dropping connections with clients when things go wrong, engage in a helpful dialogue: https://github.com/MystenLabs/narwhal/issues/107
* Finish the last bit to make consensus' crash-recovery live:https://github.com/MystenLabs/sui/issues/5200


huitseeker (changes requested):
See https://github.com/MystenLabs/narwhal/pull/102#pullrequestreview-923142657

PR #136, opened 2022-04-18
Always output tx

asonnino:
Tiny PR ensuring consensus always returns the transaction to the end user after processing (even if it encountered an error).


huitseeker (commented):
Now this looks good to me from the `SerializedTransactionDigest` side, but since you renamed `client -> executor` (which I think is a good move), I'm gonna need at least a paragraph doc comment in `executor/lib.rs` (or possibly a Readme) that explains what execution is (and isn't) in this code base and what justifies the name, written so that somebody used to working with a classic blockchain architecture (Eth, Celo, Tendermint ...) can understand.

executor/src/core.rs
```diff
@@ -38,7 +38,7 @@ pub struct Core<State: ExecutionState, PublicKey: VerifyingKey> {
     /// Receive ordered consensus output to execute.
     rx_subscriber: Receiver<ConsensusOutput<PublicKey>>,
     /// Outputs executed transactions.
-    tx_output: Sender<SubscriberResult<SerializedTransactionDigest>>,
+    tx_output: Sender<(SubscriberResult<()>, SerializedTransactionDigest)>,
```
asonnino, line 41: Revering this for the moment but will fix it later

asonnino: Made an issue here: #1489 (it needs to be fixed along with Sui)

laura-makdah (approved):
I'm glad to have one less thing named "Client".

PR #140, opened 2022-04-19
Add gRPC Server + get_collections endpoint to Narwhal

### LINKED ISSUE #195: [get_collections] introduce gRPC endpoint & first server setup
### Story

Parent story: https://github.com/MystenLabs/narwhal/issues/67

### Description

As part of this task a new gRPC endpoint should be introduce to retrieve collections. More specifically the endpoint description can be found [here](https://www.notion.so/mystenlabs/Narwhal-Reference-Mempool-API-6e6268556f1249bc958d32f4bd1d5b77#e6dd07a1e8924b9c9bcb9b09bd702fb8).

To achieve that the [block_waiter](https://github.com/MystenLabs/narwhal/blob/fecbfc9385e5599aed504d2b63de4b8763d4b6a2/primary/src/block_waiter.rs#L138) should be used to fetch the blocks/collections.

arun-koshy:
Embedded gRPC server into Primary. For `get_collections` endpoint hooked up `BlockWaiter` channels to gRPC server for communication. Channel data flow can be seen in diagram below. 

Narwhal Channel Flow drawio

Full API documentation can be found [here](https://docs.google.com/document/d/1ogieB-Y-PkvINJLfpS9G5Tga4RrmBef3OO0RSHa1fyU/edit#)

`get_collections` takes a list of `CollectionDigest` and will return a list of `CollectionRetrievalResults`. These resulting objects will be of either type `BatchMessage` or `BlockError`. `BatchMessage` indicating a successful fetch of a `Batch` where `BlockError` indicates a failure during fetch. Open to feedback here but seemed like the best route to take so the consensus layer can handle retrying if needed for just the failed `collection_ids` instead of everything. 

Also followed @bmwill example in his gRPC work in [PR#157](https://github.com/MystenLabs/narwhal/pull/157) and used `bootstrap` test for codegen instead of a `build.rs` file. 

The other piece that I am looking for feedback is seeing if I can replace the existing narwhal types to just use the generated rust types from narwhal.proto. Would we lose the traits/macros on the existing narwhal types or is there a way to add them? 

Fixes MystenLabs/narwhal#195


primary/tests/integration_tests.rs
```diff
@@ -0,0 +1,215 @@
...
+        None,
+        Some("payloads"),
+    )
+    .unwrap();
+    let payload_store = Store::new(payload_map);
+
+    // Create a new test batch store.
+    let batch_map = rocks::DBMap::<BatchDigest, SerializedBatchMessage>::open(
+        temp_dir(),
+        None,
```
akichidis, line 64: I believe you can use instead the NodeStorage https://github.com/MystenLabs/narwhal/blob/fecbfc9385e5599aed504d2b63de4b8763d4b6a2/node/src/lib.rs#L38 to initialise all the required stores.

primary/tests/integration_tests.rs
```diff
@@ -0,0 +1,215 @@
...
+        worker_id,
+        committee.clone(),
+        parameters.clone(),
+        batch_store,
+    );
+
+    // Test gRPC server with client call
+    let mut client = ValidatorClient::connect("http://127.0.0.1:50052")
+        .await
+        .unwrap();
```
akichidis, line 147: I see that you put a `sleep` above to ensure that you give some time for the primary to spin up but there is always the possibility for this to take more time. Maybe instead we want to retry here in case where the client can not connect successfully? So in this case instead of adding a sleep on line 133, we do a few retries (with some wait before each retry step) and fail permanently when maxed the retries?

types/tests/bootstrap.rs
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
```
akichidis, line 1: ok ,so I assume we'll have one `bootstrap` file that will take care of any proto build across the entire codebase, correct?

primary/src/grpc_server.rs
```diff
@@ -0,0 +1,182 @@
...
+impl Narwhal {
+    fn new(tx_get_block_commands: Sender<BlockCommand>) -> Self {
+        Self {
+            tx_get_block_commands,
+        }
+    }
+}
+
+#[tonic::async_trait]
+impl Validator for Narwhal {
```
akichidis, line 32: Assuming that we move the gRPC implementation under its own module, then it might worth having each service (`Validator`, `Proposer` etc) on their own files as well

primary/tests/integration_tests.rs
```diff
@@ -0,0 +1,215 @@
...
+};
+use tokio::sync::mpsc::channel;
+use types::{
+    BatchDigest, Certificate, CertificateDigest, CertificateDigestProto, CollectionRetrievalResult,
+    GetCollectionsRequest, Header, HeaderDigest, ValidatorClient,
+};
+use worker::{SerializedBatchMessage, Worker, WorkerMessage};
+
+#[tokio::test]
+async fn test_get_collections() {
```
akichidis, line 24: ❤️

primary/tests/integration_tests.rs
```diff
@@ -0,0 +1,215 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
```
akichidis, line 1: Is there also a way to unit test the implementation as well?

akichidis (commented):
Thanks so much for this @arun-koshy ! Also so happy to see the end-to-end test here on this! Please see the comments and we can catch up as well.

primary/tests/integration_tests.rs
bmwill: Ideally we'd bind on port zero, query what port we actually are bound to and then use that port to completely avoid port conflicts as well as sleeps in tests

primary/src/block_waiter.rs
```diff
@@ -73,6 +76,24 @@ pub struct BatchMessage {
...
+                    .iter()
+                    .map(|transaction| TransactionProto {
+                        f_bytes: transaction.to_vec(),
+                    })
+                    .collect::<Vec<TransactionProto>>(),
+            }),
+        }
+    }
+}
+
```
bmwill, line 96: Yeah as was mentioned in another comment, if possible I think it would be great if we can centralize where these types are defined (eg in the `types` crate) and then include the from impls there.

primary/src/grpc_server.rs
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
```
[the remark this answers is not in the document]
bmwill: Actually the way that Arun has set this up is exactly how I would do it. Basically the way that the gRPC code gen works is that we should have the proto types in the `types` crate which generate the client implementation as well as the server trait `Validator` etc. Then the crate that wants to implement that interface can pull in that trait and implement it given the implementation requires details only available inside the `primary` crate

primary/src/grpc_server.rs
```diff
@@ -0,0 +1,182 @@
...
+    tx_get_block_commands: Sender<BlockCommand>,
+}
+
+impl Narwhal {
+    fn new(tx_get_block_commands: Sender<BlockCommand>) -> Self {
+        Self {
+            tx_get_block_commands,
+        }
+    }
+}
```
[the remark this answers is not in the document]
bmwill: Yeah we may want to provide a more descriptive name for this as this is the component that implements the Validator interface.

primary/src/grpc_server.rs
bmwill: @akichidis what do you mean by `own module`? This implementation is already inside its own module named `primary::grpc_server`. Are you instead asking if this should go inside its own crate?

primary/src/grpc_server.rs
```diff
@@ -0,0 +1,182 @@
...
+                    ids,
+                    sender: tx_get_blocks,
+                })
+                .await
+                .unwrap();
+
+            let timer = sleep(Duration::from_millis(5_000));
+            tokio::pin!(timer);
+
+            tokio::select! {
```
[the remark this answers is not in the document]
bmwill: Let's avoid the use of `sleep` and instead if the intent is for this to be a timeout to actually use the [`timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) function instead. The code will be significantly more readable.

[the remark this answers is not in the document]
bmwill: Oh didn't read @akichidis comment who already pointed this out :)

types/tests/bootstrap.rs
```diff
@@ -0,0 +1,66 @@
...
+    let proto_files = &["proto/narwhal.proto"];
+    let dirs = &["proto"];
+
+    println!(std::env!("CARGO_MANIFEST_DIR"));
+
+    let out_dir = PathBuf::from(std::env!("CARGO_MANIFEST_DIR"))
+        .join("src")
+        .join("generated");
+
+    let config = prost_build::Config::default();
```
bmwill, line 23: We should set the `bytes` config like I do [here](https://github.com/MystenLabs/narwhal/pull/171/files#diff-21641a0d45596ea45e12276ade1b91b43e87d4883fd244e1d8769c818040acd2R20) that way we can avoid doing some extra allocations

bmwill (commented):
We'll probably have to coordinate landing of our grpc prs since there's a high likelihood of conflict :)

primary/src/grpc_server.rs
akichidis: Apologies should have made my self more clear. I mean organise in a folder structure like:
```
......
|-- grpc_server
|.    |- mod.rs
|.    |- validator.rs
|.    |- proposer.rs    
|
```
see the `block_synchronizer`

primary/src/grpc_server.rs
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
huitseeker, line 2: On a more pedestrian topic:
- the grpc server will only be run in the case Narwhal is coupled with an external consensus algorithm,
- it only serves the interface with that consensus, 
- the optionality of running it should be present in setup (the optionality in running tusk is a good place to look in order to see an example)

primary/src/grpc_server.rs
```diff
@@ -0,0 +1,182 @@
...
+            let timer = sleep(Duration::from_millis(5_000));
+            tokio::pin!(timer);
+
+            tokio::select! {
+                Ok(result) = rx_get_blocks => {
+                    get_blocks_response = match result {
+                        Ok(blocks_response) => {
+                            let mut retrieval_results = vec![];
+                            for block_result in blocks_response.blocks {
+                                match block_result {
```
huitseeker, line 117: It sounds like we want an auxiliary `Fn Result<GetBlockSReponse> -> CollectionRetrievalResult` then we can `retrieval_results.push (foo(block_result))`.

primary/src/grpc_server.rs
```diff
@@ -0,0 +1,182 @@
...
+                tx_get_block_commands,
+            }
+            .run()
+            .await
+            .map_err(|e| error!("{:?}", e));
+        });
+    }
+
+    async fn run(&mut self) -> Result<(), Box<dyn std::error::Error>> {
+        let addr = "127.0.0.1:50052".parse()?;
```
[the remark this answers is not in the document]
huitseeker: We'll need a way to set up parameters for both this system and e.g. some tuning parameters of the DAG. I think the simplest way to make progress now is to make this use a nested set of parameters in the way you introduced one for the block synchronizer.

huitseeker (changes requested):
Sadly, the (local) benchmark fails with the following error in the logs:
```
huitseeker@kelpie.localdomain➜narwhal/benchmark/logs(arun/primary-test)» grep ERROR *                                                                                                                                                               [22:13:42]
primary-1.log:2022-04-30T02:10:41.32984Z ERROR primary::grpc_server: tonic::transport::Error(Transport, hyper::Error(Listen, Os { code: 48, kind: AddrInUse, message: "Address already in use" }))
primary-2.log:2022-04-30T02:10:41.335548Z ERROR primary::grpc_server: tonic::transport::Error(Transport, hyper::Error(Listen, Os { code: 48, kind: AddrInUse, message: "Address already in use" }))
primary-3.log:2022-04-30T02:10:41.341275Z ERROR primary::grpc_server: tonic::transport::Error(Transport, hyper::Error(Listen, Os { code: 48, kind: AddrInUse, message: "Address already in use" }))
```

(and to preempt your upcoming question: https://github.com/MystenLabs/narwhal/pull/192#discussion_r861951595)

types/tests/bootstrap.rs
arun-koshy: I am more than happy to switch back, but first two questions for you

- What allocations are you talking about here? 
- I tried switching but I was having issues converting to and from `Bytes` to `[u8; 32]`. Do you know a clever way to do this?

primary/tests/integration_tests.rs
arun-koshy: @akichidis just to clarify before I make the change, I would have to add sleeps to do the waits before each retry step correct? Or is there another way to do the wait you are referring to here. 

@bmwill looks like I will need to do this in this PR to get [local benchmarking working](https://github.com/MystenLabs/narwhal/pull/140#pullrequestreview-958415847) as @huitseeker pointed out. If I bind to `127.0.0.1:0` when I spawn the gRPC server a port will dynamically assigned, which is handled by some internal tonic code. How would I be able to query that port from the client? Do I have to add some custom code here or is there an easier way to do this?

primary/tests/integration_tests.rs
arun-koshy: Yes, I will work on this and ping this comment when its done.

primary/src/grpc_server.rs
arun-koshy: Would it be fair to assume that if the [`consensus`](https://github.com/MystenLabs/narwhal/blob/main/node/src/lib.rs#L96) is set to false then we are planning to run narwhal with an external consensus algorithm? So I can reuse the `consensus` bool to see if I should start the gRPC server or not?

Or should I make it a separate subcommand as part of launching a primary?

huitseeker: > Would it be fair to assume that if the [consensus](https://github.com/MystenLabs/narwhal/blob/main/node/src/lib.rs#L96) is set to false then we are planning to run narwhal with an external consensus algorithm?

Yes, totally.

huitseeker (approved):
I think we are very close to done on this one! Thank you so much @arun-koshy . To me the main points left:
- renaming `GrpcServer` to something like `ValidatorAPIServer` or some such,
- opening an issue for [Bytes-based serialization](https://github.com/MystenLabs/narwhal/pull/140#discussion_r862996233) ([context](https://docs.serde.rs/serde_bytes/)),
- a few nits in the code, for which I'm mostly calling for doc comments.

types/src/primary.rs
```diff
@@ -485,3 +509,42 @@ pub enum PrimaryWorkerMessage<PublicKey> {
...
+    pub id: CertificateDigest,
+    pub error: BlockErrorType,
+}
+
+pub type BlockResult<T> = Result<T, BlockError>;
+
+impl<T> From<BlockError> for BlockResult<T> {
+    fn from(error: BlockError) -> Self {
+        BlockResult::Err(error)
+    }
```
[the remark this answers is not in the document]
akichidis: To be honest I see the errors produced from the `synchronizer` as separate from the `block_waiter` . Yes there seems to be an overlapping I would probably avoid mixing those.

primary/src/grpc_server/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+use self::validator::NarwhalValidator;
+use crate::BlockCommand;
+use tokio::sync::mpsc::Sender;
+use tonic::transport::Server;
+use tracing::error;
+use types::ValidatorServer;
+
+mod validator;
+
+pub struct GrpcServer {
```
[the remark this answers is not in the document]
akichidis: +1

I would probably go with something neutral (e.x `ConsensusAPIGrpc` , `ConsensusGrpcServer`) and avoid using the `validator` etc, since we already have services named like this (e.x the Validator API)

types/proto/narwhal.proto
```diff
@@ -5,6 +5,124 @@ syntax = "proto3";
...
+// TODO: Make this proto the source of truth for all relevant narwhal types.
+
+/* 
+ * Proto wrapper for Narwhal type 
+ *
+ * pub struct CertificateDigest([u8; DIGEST_LEN])
+ *
+ */
+message CertificateDigest {
+    bytes f_bytes = 1;
```
bmwill, line 17: the name `f_bytes` isn't super descriptive I also don't know what `f_` means here.

bmwill (approved):
looks good! Only had a question on naming but nothing blocking, lets get this in!

types/proto/narwhal.proto
arun-koshy: I just didn't want to reuse the message name as the bytes field name. So the `f_` was just to represent field and not use the field type `bytes` as the name. But I agree it's not clear, I changed the fields to just use a more descriptive name.

PR #144, opened 2022-04-20
Add code coverage to CI

huitseeker:


.github/workflows/codecov.yml
```diff
@@ -0,0 +1,51 @@
...
+    strategy:
+      fail-fast: true
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: beta # temporarily use beta till 'instrument-coverage' stabilizes in 1.60
+          profile: default
+          override: true
+          components: llvm-tools-preview
```
bmwill, line 24: Lets drop the requirement on the beta toolchain and instead use whats in the `rust-toolchain` file (which can be updated to 1.60 if it isn't already).
```suggestion
      - uses: actions-rs/toolchain@v1
        with:
          components: llvm-tools-preview
```

bmwill (changes requested):

.github/workflows/codecov.yml
bmwill: This is due to coverage feature being stable in 1.60

huitseeker:
Fix'd, thanks for the catch @bmwill

asonnino (approved):

bmwill (approved):

PR #148, opened 2022-04-21
[refactor] Create the types crate

huitseeker:
The types crate contains all the usual suspects: `Batch`,`Round`, `Header`, `Vote`, `Certificate`, and their digests.


bmwill:
This is great, I even noticed a lot of duplicate code wrt tests earlier today that we could try to unify at some point.

One thought: Is there much benefit to having all these things broken out into different crates (aside from some compile time benefits)? Would it maybe be easier to reason about if everything was in a single super crate?

huitseeker:
> I even noticed a lot of duplicate code wrt tests earlier today that we could try to unify at some point.

Y u no open issues? 😢 

> Is there much benefit to having all these things broken out into different crates (aside from some compile time benefits)?

Well, outside of philoophy:
- @arun-koshy is blocked on some nominal typing issue (he's trying to write a test for the primary crate and feed to a structure using the primary crate through a re-export). this is one of the ways to make this disappear.
- there's also duplication in the error types that @asonnino that could disappear because of this.

asonnino (approved):
Why not making `test_utils` a separate crate that we can import under [dev-dependencies] rather than as a standard dependency? 
Does it only contains 'types'?

arun-koshy:
Thanks for doing this @huitseeker!

arun-koshy (approved):

huitseeker:
> Why not making test_utils a separate crate that we can import under [dev-dependencies] rather than as a standard dependency?

I would see a PR doing that favorably. 
Right now test-utils is a module that's only importable under a non-default feature. It's imported in primary.

PR #155, opened 2022-04-24
[crypto] Small extensions

### LINKED ISSUE #151: Review crypto trait for `Display`
### Description
We haven't made `std::fmt::Display` part of public keys, which forces us to use `std::fmt::Debug` in print statements.
This is unwieldy, we should review crypto types and select some (defilnitely not `PrivateKey` though) to implement `Display`.

See also [`C-COMMON-TRAITS`](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)

huitseeker:
Convenience methods on crypto material. Closes #151

### DIFF
--- a/crypto/src/ed25519.rs
+++ b/crypto/src/ed25519.rs
@@ -1,3 +1,6 @@
+use std::fmt::{self, Display};
+
+use base64ct::{Base64, Encoding};
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 use serde::{de, Deserialize, Serialize};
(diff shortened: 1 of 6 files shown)


crypto/src/ed25519.rs
```diff
@@ -185,6 +200,15 @@ impl KeyPair for Ed25519KeyPair {
     }
 }
 
+impl From<ed25519_dalek::Keypair> for Ed25519KeyPair {
+    fn from(dalek_kp: ed25519_dalek::Keypair) -> Self {
```
asonnino, line 204: Thanks!

asonnino (approved):

PR #158, opened 2022-04-25
Do not sync empty parents

asonnino:
Fix the following issue: https://github.com/MystenLabs/sui/issues/1529


huitseeker (approved):
Please add a TODO & issue to test this behavior (i.e. if we fire off parent requests despite having no outstanding parent request).

PR #160, opened 2022-04-26
[test, do not merge] chore: set rust-lld as the default linker on linux

huitseeker:


huitseeker:
OK,so rust-lld still requires nightly, and lld requires installation.

PR #161, opened 2022-04-26
refactor: Replace the manual retry of ConnectionWaiter by backoff

huitseeker:


asonnino (approved):
Nice one

akichidis (commented):
Changes look good, but it stress me out that we have 0 tests around this component - and especially for things that we consider important (e.x the retry policy).

@huitseeker is this something that we need to merge asap? If not, how easy would it be to add at least a basic scenario test?

huitseeker:
THis will almost certainly be made obsolete by #171, which also reuses backoff.

PR #171, opened 2022-04-26
convert internal interfaces to grpc

bmwill:
Here's a PR that converts everything to grpc. There's still some potential cleanup that could be done but i figured this would be good enough to run some benchmarks


network/src/primary.rs
```diff
@@ -0,0 +1,217 @@
...
+        let client = self.client(address);
+        let handle = tokio::spawn(
+            self.retry_config
+                .retry(move || {
+                    let mut client = client.clone();
+                    let message = message.clone();
+                    async move { client.send_message(message).await.map_err(Into::into) }
+                })
+                .map(|response| {
+                    response.expect("we retry forever so this shouldn't fail");
```
huitseeker, line 77: In this context, the endpoint is not trusted with all that this implies.
- for replacing instances of the `reliable_sender` it's Ok to use this send-with-retry, but not for instances of the `simple_sender`, which is latency-sensitive,
- the endpoint you create here in `create_client` should really be under a connect timeout, https://github.com/hyperium/tonic/commit/2b60a00614c5c4260ce0acaaa599da89bebfd267

network/src/worker.rs
```diff
@@ -0,0 +1,143 @@
...
+                })
+                .map(|response| {
+                    response.expect("we retry forever so this shouldn't fail");
+                }),
+        );
+
+        CancelHandler(handle)
+    }
+
+    pub async fn broadcast<T: VerifyingKey>(
```
huitseeker, line 84: In principle, the workers should not need any broadcast operations (lucky or otherwise).

huitseeker:
This is in really good shape! Thank you so much!
I have opened https://github.com/bmwill/narwhal/pull/1 to fix a few warts. There's one core test not passing.
I think the missing pieces are :
- plugging in the TLS crypto,
- cleaning up the primary/worker split, reviewing the endpoint establishment for safety,
- documenting the stack and outlining next steps to make the networking code have more "rpc affinity" 
WDYT?

bmwill:
> This is in really good shape! Thank you so much! I have opened [bmwill#1](https://github.com/bmwill/narwhal/pull/1) to fix a few warts. There's one core test not passing. I think the missing pieces are :
> 
> * plugging in the TLS crypto,
> * cleaning up the primary/worker split, reviewing the endpoint establishment for safety,
> * documenting the stack and outlining next steps to make the networking code have more "rpc affinity"
>   WDYT?

Yep I agree with all that as next steps. This PR was mostly about getting things converted to some common base (that and so i don't fight with conflicts any more than i have already :) ) so that we can then make those changes easier later

huitseeker (approved):
Thanks a TON, this is ✨ very cool ✨  :)

While I do think the "timeout-for-connects" thing is material, we can open an issue to fix it a bit later.

laura-makdah (commented):
Are there terminal commands to auto-generate the protobuf definitions like there are in Golang?

bmwill:
> Are there terminal commands to auto-generate the protobuf definitions like there are in Golang?

this is all taken care of by the bootstrap.rs test. Just run `cargo test --test bootstrap` to regenerate the bindings. If that gets to be too difficult we can always move it into a build.rs file (at the cost of ide breakage)

PR #173, opened 2022-04-26
support duration for Properties

akichidis:
Given the recent work on the `block_synchronizer` I got the need to use `Duration` as property coming from the `Properties` struct. Unfortunately, serializing a `Duration` looks like:
```
{
    "secs": xxxxxx, "nanos": xxxxxx
}
```

which is quite counter intuitive on a configuration file. I've wrote small deserialize/serialize methods to support defining durations in miliseconds and seconds in a simple manner, like:
```
{
   "property_1": "10ms",
   "property_2": "50s"
}
```

tried to keep it simple so I've avoided brining in any Regex library etc.

I've also refactored the code so we use only `Duration` on the properties whenever we need to (and from now on we should probably stick to that), so we avoid doing conversions and have the need to define time units as part of the variable names (e.x `_ms` ) or put the units in comments.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -362,6 +362,7 @@ mod tests {
     use crate::{duration_format, Deserialize, Import, Parameters};
     use std::{fs::File, io::Write, time::Duration};
     use tempfile::tempdir;
+    use tracing_test::traced_test;
 
     #[derive(Deserialize)]
     struct MockProperties {
@@ -372,6 +373,7 @@ mod tests {
     }
 
     #[test]
+    #[traced_test]
     fn parse_properties() {
         // GIVEN
         let input = r#"{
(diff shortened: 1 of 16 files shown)


huitseeker (changes requested):
This will need to adapt the new config parameters involved in the fabfile, see broken benchmark issue in #178

config/src/lib.rs
```diff
@@ -300,3 +324,167 @@ impl<PublicKey: VerifyingKey> Committee<PublicKey> {
             .collect()
     }
 }
+
+/// Allow us to serialize and deserialize Duration values in a more
+/// human friendly format (e.x in json files). When serialized then
+/// a string of the following format is written: [number]ms , for
+/// example "20ms". When deserialized, then a Duration is created.
+mod duration_format {
```
huitseeker, line 332: We may want to export it to a different file.  In particular, I'd love to see module-level doc ( `//!`, see https://doc.rust-lang.org/reference/comments.html#examples) that makes it spectacularly clear this supports ms and s only. 


There's a chance for people to expect [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.txt) support (see also https://github.com/time-rs/time/blob/main/src/serde/rfc3339.rs), this would be out of scope, and I don't want us discovering it at deployment time.

config/src/lib.rs
```diff
@@ -300,3 +324,167 @@ impl<PublicKey: VerifyingKey> Committee<PublicKey> {
...
+            params
+                .block_synchronizer
+                .payload_availability_timeout
+                .as_millis(),
+            4_000
+        );
+    }
+
+    #[test]
+    #[traced_test]
```
huitseeker, line 432: 👍

huitseeker (approved):
Nice work!

Could we sync with @asonnino (who will eventually upgrade to the commit of Narwhal where this is merged) to make sure this does not break Sui's config management?

config/src/lib.rs
akichidis: Good idea!  I've extracted this now to a separate file

akichidis: Regarding https://www.rfc-editor.org/rfc/rfc3339.txt you are absolutely right, definitely out of scope as I wanted something very simple for now. I've tried to make this as clear as possible on the comments.

PR #177, opened 2022-04-27
[dag] DAG Extensions 

huitseeker:
This 
- adds many convenience methods and doc tests
- adds flipping the compressible bit to a DAG vertex, enacting the semantics we want for removal
- extends tests to check compression and removal semantics

### DIFF
--- a/dag/src/lib.rs
+++ b/dag/src/lib.rs
@@ -57,6 +57,23 @@ impl<T> From<Arc<Node<T>>> for NodeRef<T> {
     }
 }
 
+impl<T> NodeRef<T> {
+    /// Returns a NodeRef pointing at the Node passed as argument
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use dag::{ Node, NodeRef };
+    ///
+    /// let node = Node::new_leaf(1, false);
+    /// // Note the 2 derefs: one for the newtype, one for the Arc
+    /// assert_eq!(Node::new_leaf(1, false), **NodeRef::from_pointee(node));
+    /// ```
+    pub fn from_pointee(val: Node<T>) -> Self {
+        Arc::new(val).into()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


dag/src/lib.rs
```diff
@@ -57,6 +57,23 @@ impl<T> From<Arc<Node<T>>> for NodeRef<T> {
...
+    /// # Example
+    ///
+    /// ```
+    /// use dag::{ Node, NodeRef };
+    ///
+    /// let node = Node::new_leaf(1, false);
+    /// // Note the 2 derefs: one for the newtype, one for the Arc
+    /// assert_eq!(Node::new_leaf(1, false), **NodeRef::from_pointee(node));
+    /// ```
+    pub fn from_pointee(val: Node<T>) -> Self {
```
akichidis, line 72: Where do we intend to use this?

akichidis (approved):
Nice work! Thanks for the extra convenience methods 💯

dag/src/lib.rs
```diff
@@ -78,8 +95,26 @@ pub struct Node<T> {
...
-impl<T: Sync + Send + std::fmt::Debug> Node<T> {
+impl<T: PartialEq> PartialEq for Node<T> {
+    fn eq(&self, other: &Self) -> bool {
+        *self.parents.load() == *other.parents.load()
+            && self.is_compressible() == other.is_compressible()
+            && self.value.eq(&other.value)
+    }
+}
+
+impl<T: Eq> Eq for Node<T> {}
```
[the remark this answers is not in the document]
huitseeker: Just indicating the [equality is total](https://doc.rust-lang.org/std/cmp/trait.Eq.html#how-can-i-implement-eq).

dag/src/lib.rs
huitseeker: As a constructor.

Issue #175, opened 2022-04-26
[benchmark] Benchmark broken since #125


huitseeker:
See https://github.com/MystenLabs/narwhal/commit/767a364e9937d38e52db1df406aa9bfc3bb4d1eb for the benchmark failure, and CI benchmark success on the prior commit.

This is actually a bootstrap failure on the config file, which does not include the new parameters:
```
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
Error: Failed to load the node's parameters

Caused by:
    Failed to read config file '.parameters.json': missing field `block_synchronizer` at line 9 column 1
```
This is due to the deserialization of the new block synchronizer parameters being absent of the `parameters.json` created in benchmarks from [this entry point](https://github.com/MystenLabs/narwhal/blob/76708dcf786a8c40d8284d99d0b9eba87b35a084/benchmark/fabfile.py#L23-L31)

Once this is fixed (by unplugging the parameters):
```
huitseeker@dapple.local➜narwhal/benchmark/logs(repair_benchmarks)» grep panick *                                                                                                                                                                                                                                                                                                                                                                                                   [15:39:56]
primary-0.log:thread 'tokio-runtime-worker' panicked at 'all branches are disabled and there is no else branch', /Users/huitseeker/tmp/narwhal/primary/src/block_synchronizer/mod.rs:221:13
[...]

PR #178, opened 2022-04-27
fix: missing block_synchronizer config from fabfile

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/175

This PR fixes the issue of the missing block_synchronizer configuration from the fabfile which was causing the benchmarks fail - basically the primaries were crashing and never be able to bootstrap.

Re-pluggin the `block_synchronizer` will be as part of the https://github.com/MystenLabs/narwhal/pull/127

### DIFF
--- a/benchmark/fabfile.py
+++ b/benchmark/fabfile.py
@@ -27,7 +27,12 @@ def local(ctx, debug=True):
         'sync_retry_delay': 10_000,  # ms
         'sync_retry_nodes': 3,  # number of nodes
         'batch_size': 500_000,  # bytes
-        'max_batch_delay': 200  # ms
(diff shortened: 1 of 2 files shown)


huitseeker (approved):
🙇 Thanks!

Just in case you find a place to use it: https://docs.rs/tracing-test/latest/tracing_test/

akichidis:
> 🙇 Thanks!
> 
> Just in case you find a place to use it: https://docs.rs/tracing-test/latest/tracing_test/

Ah nice, thanks for sharing, wasn't aware of it! I'll use it as part of https://github.com/MystenLabs/narwhal/pull/173 where I've introduced some tests

PR #182, opened 2022-04-27
Initial digests re-proposal

asonnino:
Repropose digests from garbage collected certificates that did not make it through consensus.

### DIFF
--- a/primary/src/garbage_collector.rs
+++ b/primary/src/garbage_collector.rs
@@ -22,6 +22,8 @@ pub struct GarbageCollector<PublicKey: VerifyingKey> {
     consensus_round: Arc<AtomicU64>,
     /// Receives the ordered certificates from consensus.
     rx_consensus: Receiver<Certificate<PublicKey>>,
+    /// Re-sent our own GC'ed digests to the proposed.
+    tx_our_digests: Sender<(BatchDigest, WorkerId)>,
     /// The network addresses of our workers.
     addresses: Vec<SocketAddr>,
     /// A network sender to notify our workers of cleanup events.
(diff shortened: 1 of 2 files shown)


primary/src/garbage_collector.rs
```diff
@@ -57,9 +61,19 @@ impl<PublicKey: VerifyingKey> GarbageCollector<PublicKey> {
     async fn run(&mut self) {
         let mut last_committed_round = 0;
         while let Some(certificate) = self.rx_consensus.recv().await {
-            // TODO [issue #9]: Re-include batch digests that have not been sequenced into our next block.
-
             let round = certificate.round();
+
+            // Re-include batch digests that have not been sequenced into our next block.
+            if certificate.author() == name {
```
akichidis, line 67: I thought that the certificates coming form the `rx_consensus` are the ones that have been sequenced, aren't they?

primary/src/garbage_collector.rs
```diff
@@ -57,9 +61,19 @@ impl<PublicKey: VerifyingKey> GarbageCollector<PublicKey> {
...
+                for payload in certificate.header.payload {
+                    self.tx_our_digests
```
akichidis, line 69: Probably this might be too much to change, but do we want to introduce some concept of "maximum" retries to avoid some "infinite" re-inclusions? I guess the probability might be small for this to happen (I have no data to support the argument, it's pure finger in the air estimation) but we might want to have maximum number of retries for the digests which could be piggy backed to the digests circulated within the app. When max retries to be included reached then we completely discard with related log and metrics.

asonnino:
Please ignore this PR for now. Digests re-inclusion is more complex that what it seems, nothing in this PR is complete/working atm

asonnino:
Too old, the code changed to much

PR #185, opened 2022-04-28
Output the entire tx data

asonnino:
Sui needs the full transaction output to fix bug https://github.com/MystenLabs/sui/issues/1624


akichidis:
@asonnino is this change needed in order to be able to find/confirm the issue in SUI?

asonnino:
> @asonnino is this change needed in order to be able to find/confirm the issue in SUI?

Yes, and to fix it :)

Issue #164, opened 2022-04-26
[block_synchronizer] implement the PayloadAvailabilityResponse message


akichidis:
### Story

Parent story: https://github.com/MystenLabs/narwhal/issues/67

### Description

When a `SynchronizeBlockPayload` command is received, the first step on the protocol it would be to ask the other primary peers about whether they do have the requests payloads available to sync. For this a `PrimaryMessage::PayloadAvailabilityRequest` message is being sent to all the peers. As part of this task we should implement the handling of that message from the other primary peers. More specifically we should enhance the primary in order to:
* be able to receive and deserialize the message
* check in its local `certificates_store` whether it has the requested certificates
* check in its local `payload_store` whether it has the requested payload
* respond back to the primary with the availability by sending a [PrimaryMessage::PayloadAvailabilityResponse](https://github.com/MystenLabs/narwhal/blob/767a364e9937d38e52db1df406aa9bfc3bb4d1eb/primary/src/primary.rs#L66) message . For each `CertificateDigest` that has been requested, a boolean value should be sent with `true` designating that the peer is able to sync the payload, and `false` in the opposite case.

PR #186, opened 2022-04-28
feat: handle the payload availability message

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/164

This PR is implementing the handling of the `PrimaryMessage::PayloadAvailabilityRequest` message on the primary nodes. I've enhanced the `helper` component as this is probably the closest could re-use. I've avoided introducing something new at this point, until we need to expand further the protocol with more messages etc. I've also performed a small refactoring on the helper so I can de-dupe some code around sending the message back to the requestor.

### DIFF
--- a/primary/src/helper.rs
+++ b/primary/src/helper.rs
@@ -72,8 +72,8 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
                     self.process_certificates(certificate_ids, requestor, true)
                         .await;
                 }
-                // A request that another primary sends us to check whether we have
-                // available to serve the payloads for the provided certificate_ids.
+                // A request that another primary sends us to ask whether we
+                // can serve batch data for the provided certificate_ids.
                 PrimaryMessage::PayloadAvailabilityRequest {
                     certificate_ids,
                     requestor,
@@ -88,10 +88,10 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
         }
     }
 
-    /// Checks for each provided digest of the list of digests they node has
-    /// the corresponding certificate within its storage and the payload as
-    /// well. If yes, then the requestor can follow up and request the batches
-    /// payload from the primary's workers via the standard syncing approach.
+    /// Processes a payload availability request by checking we have the
+    /// certificate & batch data for each certificate digest in digests,
+    /// and reports on each fully available item in the request in a
+    /// PayloadAvailabilityResponse.
     async fn process_payload_availability(
         &mut self,
         digests: Vec<CertificateDigest>,
@@ -130,19 +130,15 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
             }
         };
 
-        let certificates_and_ids: Vec<(CertificateDigest, Option<Certificate<PublicKey>>)> =
-            digests.into_iter().zip(certificates).collect();
-
-        for (id, certificate_option) in certificates_and_ids {
+        for (id, certificate_option) in digests.into_iter().zip(certificates) {
             // Find the batches only for the certificates that exist
             if let Some(certificate) = certificate_option {
-                let payload: Vec<(BatchDigest, WorkerId)> =
-                    certificate.header.payload.into_iter().collect();
-
-                let payload_available = match self.payload_store.read_all(payload).await {
-                    Ok(payload_result) => {
-                        payload_result.into_iter().all(|x| x.is_some()).to_owned()
-                    }
+                let payload_available = match self
+                    .payload_store
+                    .read_all(certificate.header.payload)
+                    .await
+                {
+                    Ok(payload_result) => payload_result.into_iter().all(|x| x.is_some()),
                     Err(err) => {
                         // we'll assume that we don't have available the payloads,
                         // otherwise and error response should be sent back.
--- a/primary/src/helper.rs
+++ b/primary/src/helper.rs
@@ -1,26 +1,31 @@
 // Copyright (c) 2021, Facebook, Inc. and its affiliates
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
-use crate::primary::PrimaryMessage;
-use config::Committee;
+use crate::{primary::PrimaryMessage, PayloadToken};
+use config::{Committee, WorkerId};
 use crypto::traits::VerifyingKey;
 use network::PrimaryNetwork;
 use store::Store;
 use tokio::sync::mpsc::Receiver;
 use tracing::{error, warn};
-use types::{Certificate, CertificateDigest};
+use types::{BatchDigest, Certificate, CertificateDigest};
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


primary/src/tests/helper_tests.rs
```diff
@@ -180,6 +199,218 @@ async fn test_process_certificates_batch_mode() {
...
+
+        rocksdb
+            .put_cf(
+                &rocksdb
+                    .cf_handle(CERTIFICATES_CF)
+                    .expect("Couldn't find column family"),
+                serialised_key,
+                dummy_value,
+            )
+            .expect("Couldn't insert value");
```
akichidis, line 362: Didn't find a better to inject/trigger an error for our storage to test the error path on the code. 

I am going to explore offline the https://docs.rs/mockall/latest/mockall/ so can see how we can handle such stuff in a more pretty way (and mocking in tests in general), but not for now.

primary/src/primary.rs
```diff
@@ -65,6 +66,7 @@ pub enum PrimaryMessage<PublicKey: VerifyingKey> {
 
     PayloadAvailabilityResponse {
         payload_availability: Vec<(CertificateDigest, bool)>,
+        from: PublicKey,
```
akichidis, line 69: Unfortunately until we have authentication between the nodes we'll have to include the peer's name/key between the requests/responses for messages that need to identify per peer - which is of course unsafe.

primary/src/helper.rs
```diff
@@ -1,41 +1,50 @@
...
+    name: PublicKey,
     /// The committee information.
     committee: Committee<PublicKey>,
-    /// The persistent storage.
-    store: Store<CertificateDigest, Certificate<PublicKey>>,
-    /// Input channel to receive certificates requests.
+    /// The certificate persistent storage.
+    certificate_store: Store<CertificateDigest, Certificate<PublicKey>>,
+    /// The payloads (batches) persistent storage.
+    payload_store: Store<(BatchDigest, WorkerId), PayloadToken>,
```
huitseeker, line 27: Nit : I wonder if we're not creating confusion by introducing the keyword `Payload` in many names, where "BatchData" would have sufficed using old concepts.

primary/src/helper.rs
```diff
@@ -63,35 +72,128 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
...
                     panic!("Received unexpected message!");
                 }
             }
         }
     }
 
+    /// Checks for each provided digest of the list of digests they node has
+    /// the corresponding certificate within its storage and the payload as
+    /// well. If yes, then the requestor can follow up and request the batches
+    /// payload from the primary's workers via the standard syncing approach.
```
huitseeker, line 94: Nit: consider something like:
 "Processes a payload availability request by checking we have the certificate & batch data for each certificate digest in `digests`, and reports on each fully available item in the request in a `PayloadAvailabilityResponse`."

primary/src/helper.rs
```diff
@@ -63,35 +72,128 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
...
+                return;
+            }
+        };
+
+        let mut result: Vec<(CertificateDigest, bool)> = Vec::new();
+
+        let certificates = match self.certificate_store.read_all(digests.to_owned()).await {
+            Ok(certificates) => certificates,
+            Err(err) => {
+                error!("Error while retrieving certificates: {err}");
```
huitseeker, line 114: This is a failure on a simple DB read: we're going to have many such ones in logs. 
We can enter a [span](https://docs.rs/tracing/latest/tracing/#spans), equipped with context from the digests, to help us log what happened here in a precise way without encumbering this function with logging code. 

Here's an example of it in action in Sui:
https://github.com/MystenLabs/sui/blob/1a25e68fbc849684344680a4da134d6a62b136c3/sui_core/src/authority_server.rs#L169-L174

primary/src/helper.rs
```diff
@@ -63,35 +72,128 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
...
+                    .await;
+
+                return;
+            }
+        };
+
+        let certificates_and_ids: Vec<(CertificateDigest, Option<Certificate<PublicKey>>)> =
+            digests.into_iter().zip(certificates).collect();
+
+        for (id, certificate_option) in certificates_and_ids {
```
huitseeker, line 136: You're materializing the iterator with `collect()` and immediately converting this Vec into another iterator. Consider iterating on the zip instead.

primary/src/helper.rs
```diff
@@ -63,35 +72,128 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
...
+            // Find the batches only for the certificates that exist
+            if let Some(certificate) = certificate_option {
+                let payload: Vec<(BatchDigest, WorkerId)> =
+                    certificate.header.payload.into_iter().collect();
+
+                let payload_available = match self.payload_store.read_all(payload).await {
+                    Ok(payload_result) => {
+                        payload_result.into_iter().all(|x| x.is_some()).to_owned()
```
huitseeker, line 144: Isn't the final result a boolean? Why the `to_owned`?

primary/src/helper.rs
```diff
@@ -63,35 +72,128 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
...
+                    }
+                    Err(err) => {
+                        // we'll assume that we don't have available the payloads,
+                        // otherwise and error response should be sent back.
+                        error!("Error while retrieving payloads: {err}");
```
huitseeker, line 149: Same remark about an initial span helping you with context.

huitseeker (approved):
The logic LGTM, I did leave a few comments that I hope can help the error reporting.

primary/src/helper.rs
akichidis: Ah, that's a very good point @huitseeker and I think we might need to move into some renaming in general. I did stick to the `payload_` naming convention that has been used in several places (especially around the store semantics). I would be happy to align in one name (e.x `batch`) and stick with that everywhere

huitseeker: Here's my plug to ruplacer: https://github.com/dmerejkowsky/ruplacer

primary/src/helper.rs
akichidis: That a really good idea. Ok let me address this to a follow up PR so I can deal with tracing a bit more properly. I'll actually going to look into how I could do this in a more centralised way - meaning starting the tracing from the request level - will take the `helper` as an example and provide something.

primary/src/helper.rs
akichidis: An oversight from some refactoring, thanks!

primary/src/helper.rs
akichidis: Issue here: https://github.com/MystenLabs/narwhal/issues/201

primary/src/helper.rs
akichidis: oh nice, thanks!

PR #187, opened 2022-04-28
[refactor] Separate test utilities in their own crate

### LINKED ISSUE #117: Unify consensus testing facilities
### Description
The consensus crate is growing and there are now a number of testing facilities spread across the unit test files (e.g., `consensus_tests.rs`, `subscriber_tests.rs`, etc). These testing facilities are used multiple times in various test files (not only in the one where they are defined). It would great to group all facilities into a single file `common.rs`.

huitseeker:
Fixes #117


test_utils/src/lib.rs
```diff
@@ -313,3 +328,68 @@ pub fn resolve_name_and_committee(
...
+    for _ in 0..num_of_transactions {
+        transactions.push(transaction());
+    }
+
+    Batch(transactions)
+}
+
+const BATCHES_CF: &str = "batches";
+
+pub fn open_batch_store() -> Store<BatchDigest, worker::SerializedBatchMessage> {
```
akichidis, line 387: Since we are moving this , then we could also probably move the [create_db_stores](https://github.com/MystenLabs/narwhal/blob/767a364e9937d38e52db1df406aa9bfc3bb4d1eb/primary/src/tests/common.rs#L26) ? I know it's very `primary` specific, but since we are doing for the batch_store it would be nice to have all in one place.

akichidis (approved):
Thanks for doing this!! 💯

test_utils/src/lib.rs
huitseeker: There's unfortunately many such functions all over the code base still, including here but also in consensus ...
The place where those leftovers are still useful is in limiting cyclic dependencies from the test utils to every other crate: we wouldn't want the testing of e.g. primary to be blocked by worker (the way it is now due to such a dependency, though it will go away when I fix #188 ).

Issue #188, opened 2022-04-28
[bug] Batch Digest mismatch 


huitseeker:
Found while testing #187:

We define two ways of hashing a batch. 
1. https://github.com/MystenLabs/narwhal/blob/2fabfe47a9adcbc2f3d75144f57921dd325ab6c0/types/src/lib.rs#L58-L66
2. https://github.com/MystenLabs/narwhal/blob/2fabfe47a9adcbc2f3d75144f57921dd325ab6c0/worker/src/batch_maker.rs#L118-L122

Also corresponding to 2.:
- https://github.com/MystenLabs/narwhal/blob/2fabfe47a9adcbc2f3d75144f57921dd325ab6c0/worker/src/processor.rs#L33-L38
- https://github.com/MystenLabs/narwhal/blob/2fabfe47a9adcbc2f3d75144f57921dd325ab6c0/executor/src/batch_loader.rs#L181-L184

We should align these.

PR #189, opened 2022-04-28
Fix seq number bug

asonnino:
* Fix a nasty bug in the consensus
* Print rare debug log for consensus output


huitseeker (commented):
Ouch! IIUC, it's an off-by-one error for every transaction on the receiving end of `ConsensusOutput` correct?

Would it be possible to add a test for this?

consensus/src/tusk.rs
```diff
@@ -263,6 +262,9 @@ impl<PublicKey: VerifyingKey> Consensus<PublicKey> {
                     certificate: x,
                     consensus_index,
                 });
+
+                // Increase the global consensus index.
+                consensus_index += 1;
```
akichidis, line 267: Ok, so if I understand correctly this was starting the index for the first certificate from 1 and not from 0, right? What about when the `store.read_last_consensus_index()` is called?

https://github.com/MystenLabs/narwhal/blob/976355d2ff75aefc6b7ec7fd32742d3d7188c4e0/consensus/src/tusk.rs#L122

I assume we want to continue then from the "next" value of the consensus index. So for example, if we last stored the index `5` , then we want to continue from `6`. Would that be guaranteed as well if you change the index to that point here?

akichidis: Also, the fact that no test broke by making this change tells me that probably we don't have coverage, could you please add at least the simplest case that could verify/catch this?

asonnino:
> Ouch! IIUC, it's an off-by-one error for every transaction on the receiving end of `ConsensusOutput` correct?
> 
> Would it be possible to add a test for this?

Very hard to test because it depends on the client (Sui in this case, who could have been using any offset)

huitseeker:
Would a mock subscriber work? Just to make sure we run through the code? e.g. the `handle_client_transactions` test in the worker has help me ferret out a bug in #188 and I am very grateful it was here.

asonnino:
> Would a mock subscriber work? Just to make sure we run through the code? e.g. the `handle_client_transactions` test in the worker has help me ferret out a bug in #188 and I am very grateful it was here.

yes we can do that. and you're right we should have tests for this kind of things

consensus/src/tusk.rs
asonnino: No tests broke because it is not a bug per se. The problem is that the client (not in the Narwhal codebase) expected sequence numbers starting at '1' while narwhal started its sequence numbers at '0'. A different client (other than Sui) may very well start its sequence numbers as '0', or 'x'.

asonnino: More intriguing however is why this bug wasn't triggered in a narwhal test (it was triggered in a test in Sui): #191

asonnino:
> Would a mock subscriber work? Just to make sure we run through the code? e.g. the `handle_client_transactions` test in the worker has help me ferret out a bug in #188 and I am very grateful it was here.

There were tests... in `consensus/srs/tests/subscriber_tests.rs`... the problem is that I forgot to add the following lines to `subscriber.rs`
```
#[cfg(any(test, feature = "benchmark"))]
#[path = "tests/subscriber_tests.rs"]
pub mod subscriber_tests;
```
So all the tests were gracefully ignored :)

consensus/src/tests/subscriber_tests.rs
```diff
@@ -43,10 +40,10 @@ pub async fn spawn_node(
     let committee = mock_committee(&keys[..]);
 
     // Create the storages.
-    let consensus_store_path = temp_testdir::TempDir::default();
-    let consensus_store = make_consensus_store(&consensus_store_path);
-    let certificate_store_path = temp_testdir::TempDir::default();
-    let certificate_store = make_certificate_store(&certificate_store_path);
+    let consensus_store_path = tempfile::tempdir().unwrap();
+    let consensus_store = make_consensus_store(consensus_store_path.path());
+    let certificate_store_path = tempfile::tempdir().unwrap();
```
huitseeker, line 45: Hopefully you should be able to call `test_utils::temp_dir()`

consensus/Cargo.toml
```diff
@@ -28,6 +28,7 @@ criterion = "0.3.5"
 ed25519-dalek = { version = "1.0.1", features = ["batch", "serde"] }
 pprof = { version = "0.8.0", features = ["criterion", "flamegraph"] }
 test_utils = { path = "../test_utils"}
+tempfile = "3.3.0"
```
huitseeker, line 31: this shouldn't be needed

consensus/src/tusk.rs
```diff
@@ -244,25 +246,23 @@ impl<PublicKey: VerifyingKey> Consensus<PublicKey> {
...
                 sequence.push(ConsensusOutput {
                     certificate: x,
                     consensus_index,
                 });
+
+                // Increase the global consensus index.
+                consensus_index += 1;
+
+                // Persist the update.
+                // TODO [issue #116]: Ensure this is not a performance bottleneck.
```
huitseeker, line 264: Nit: Did you run the consensus micro-benchmark?
(not that this matters as you just moved the stuff)

consensus/src/subscriber.rs
```diff
@@ -13,6 +13,10 @@ use tokio::{
 use tracing::{debug, error};
 use types::{Certificate, CertificateDigest};
 
+#[cfg(any(test, feature = "benchmark"))]
```
huitseeker, line 16: The `feature = "benchmark"` is not necessary and will probably break stuff

huitseeker (approved):
Thanks!!

consensus/src/tusk.rs
huitseeker: (oh: I did and it's fine)

consensus/src/tests/subscriber_tests.rs
asonnino: fixed

consensus/Cargo.toml
asonnino: fixed

consensus/src/subscriber.rs
asonnino: good catch

PR #192, opened 2022-04-29
Create Central Accounting layer

lanvidr:
Addresses issue https://github.com/MystenLabs/sui/issues/5216

This creates a Central Accounting Layer along with a new type of network receiver that interfaces with the Central Accounting. The Central Accounting Layer takes over the responsibility for calling `dispatch` on the respective MessageHandler that is associated with the receiver, so that the receiver no longer calls `dispatch`. The accounting receiver  sends external messages it receives, along with some metadata, to the Central Accounting Layer, which will later contain rate-limiting and possibly prioritization logic. For now, it simply calls `dispatch`. 

As a next step, we need to adjust the MessageHandler interface to allow for being able to track and record when each message's processing has completed. Although we forward back responses through the Central Accounting Layer, because "Acks" are either returned immediately or never, the time at which the response is forwarded cannot be used to signal that the message processing is complete.  Another approach might be to implement the Drop for a custom struct witch contains the message Bytes, however, the majority of the components deserialize the message immediately upon reception, which would cause the drop to occur before message processing has completed. We can consider to deserialize the message before forwarding, and using the Drop impl strategy, or, refactor each component to explicitly send notice of message completion.

### DIFF
--- a/network/src/accounting_receiver.rs
+++ b/network/src/accounting_receiver.rs
@@ -1,52 +1,41 @@
 // Copyright (c) 2021, Facebook, Inc. and its affiliates
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
+use crate::central_accounting::{AccountingMessage, ReceiverId};
 use crate::error::NetworkError;
-use async_trait::async_trait;
+use crate::receiver::TcpWriter;
+use crate::Writer;
 use bytes::Bytes;
-use futures::stream::{SplitSink, StreamExt as _};
+use futures::stream::StreamExt as _;
 use futures::SinkExt;
-use std::borrow::BorrowMut;
-use std::{error::Error, net::SocketAddr};
+use std::net::SocketAddr;
 use tokio::net::{TcpListener, TcpStream};
-use tokio::sync::mpsc::{channel, Receiver as mspcReceiver, Sender};
+use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio_util::codec::{Framed, LengthDelimitedCodec};
 use tracing::{debug, info, warn};
 
-/// Convenient alias for the writer end of the TCP channel.
-pub type TcpWriter = SplitSink<Framed<TcpStream, LengthDelimitedCodec>, Bytes>;
-
-/// Convenient alias for the output of the dispatch function.
-pub type Writer = Sender<Bytes>;
-
-#[async_trait]
-pub trait MessageHandler: Clone + Send + Sync + 'static {
-    /// Defines how to handle an incoming message. A typical usage is to define a `MessageHandler` with a
-    /// number of `Sender<T>` channels. Then implement `dispatch` to deserialize incoming messages and
-    /// forward them through the appropriate delivery channel. Then `writer` can be used to send back
-    /// responses or acknowledgments to the sender machine (see unit tests for examples).
-    async fn dispatch(&self, writer: &mut Writer, message: Bytes) -> Result<(), Box<dyn Error>>;
-}
-
 /// For each incoming request, we spawn a new runner responsible to receive messages and forward them
-/// through the provided deliver channel.
-pub struct Receiver<Handler: MessageHandler> {
+/// to the CentralAccounting layer through the provided channel.
+pub struct AccountingReceiver {
     /// Address to listen to.
     address: SocketAddr,
-    /// Struct responsible to define how to handle received messages.
-    handler: Handler,
 }
 
-impl<Handler: MessageHandler> Receiver<Handler> {
+impl AccountingReceiver {
+    /// Create a new AccountingReceiver.
+    pub fn new(address: SocketAddr) -> Self {
+        AccountingReceiver { address }
+    }
+
     /// Spawn a new network receiver handling connections from any incoming peer.
-    pub fn spawn(address: SocketAddr, handler: Handler) {
+    pub fn spawn(self, id: ReceiverId, tx_message: Sender<AccountingMessage>) {
         tokio::spawn(async move {
-            Self { address, handler }.run().await;
+            self.run(id, tx_message).await;
         });
     }
 
     /// Main loop responsible to accept incoming connections and spawn a new runner to handle it.
-    async fn run(&self) {
+    async fn run(self, id: ReceiverId, tx_forwarding: Sender<AccountingMessage>) {
         let listener = TcpListener::bind(&self.address)
             .await
             .expect("Failed to bind TCP port");
@@ -62,30 +51,34 @@ impl<Handler: MessageHandler> Receiver<Handler> {
                 }
             };
             info!("Incoming connection established with {peer}");
-            Self::spawn_runner(socket, peer, self.handler.clone()).await;
+
+            Self::spawn_runner(socket, peer, id, tx_forwarding.clone()).await;
         }
     }
 
-    /// Spawn a new runner to handle a specific TCP connection. It receives messages and process them
-    /// using the provided handler.
-    async fn spawn_runner(socket: TcpStream, peer: SocketAddr, handler: Handler) {
+    /// Spawn a new runner to handle a specific TCP connection. It receives messages and forwards them
+    /// using to the CentralAccounting layer.
+    async fn spawn_runner(
+        socket: TcpStream,
+        peer: SocketAddr,
+        receiver_id: ReceiverId,
+        tx_forwarding: Sender<AccountingMessage>,
+    ) {
         tokio::spawn(async move {
             let transport = Framed::new(socket, LengthDelimitedCodec::new());
-            let (writer, mut reader) = transport.split();
-
-            let (tx_response, tr_response): (Writer, mspcReceiver<Bytes>) = channel(100);
-            Self::spawn_response_writer(tr_response, writer).await;
+            let (tcp_writer, mut reader) = transport.split();
+            let (tx_response, tr_response): (Writer, Receiver<Bytes>) = channel(100);
+            Self::spawn_response_writer(tr_response, tcp_writer).await;
 
             while let Some(frame) = reader.next().await {
                 match frame.map_err(|e| NetworkError::FailedToReceiveMessage(peer, e)) {
                     Ok(message) => {
-                        if let Err(e) = handler
-                            .dispatch(tx_response.clone().borrow_mut(), message.freeze())
-                            .await
-                        {
-                            warn!("{e}");
-                            return;
-                        }
+                        let accounting_message = AccountingMessage {
+                            receiver_id,
+                            message: message.freeze(),
+                            tx_response: tx_response.clone(),
+                        };
+                        _ = tx_forwarding.send(accounting_message).await;
                     }
                     Err(e) => {
                         warn!("{e}");
@@ -99,8 +92,8 @@ impl<Handler: MessageHandler> Receiver<Handler> {
 
     /// Spawns a new task that will listen for any responses from the accounting layer and
     /// forward them to the tcp writer.
-    async fn spawn_response_writer(
-        mut tr_response: mspcReceiver<Bytes>,
+    pub async fn spawn_response_writer(
+        mut tr_response: Receiver<Bytes>,
         mut tcp_writer: TcpWriter,
     ) {
         tokio::spawn(async move {
--- a/network/src/central_accounting.rs
+++ b/network/src/central_accounting.rs
@@ -1,0 +1,106 @@
+// Copyright (c) 2021, Facebook, Inc. and its affiliates
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use crate::{MessageHandler, Writer};
+use bytes::Bytes;
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use tokio::select;
+use tokio::sync::mpsc::{channel, Receiver as mspcReceiver, Sender};
+use tracing::warn;
+
+#[cfg(test)]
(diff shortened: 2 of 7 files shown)


network/src/central_accounting.rs
```diff
@@ -0,0 +1,106 @@
...
+    async fn run(&mut self) {
+        loop {
+            select! {
+                Some(mut message) = self.tr_message.recv() => {
+                    // use receiver_id to look up the handler for this receiver
+                    let handler = self.receiver_info.get(&message.receiver_id);
+                    let handler = match handler {
+                        Some(h) => h,
+                        None => {
+                            continue;
```
akichidis, line 91: We might want to put a log here. Under normal circumstances we shouldn't receive here messages of unknown receiver ids (not having a handler), right?

network/src/central_accounting.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// necessary to start an AccountingReceiver.
+    pub async fn register(&mut self, handler: Handler) -> (ReceiverId, Sender<AccountingMessage>) {
+        let id = get_id();
+        self.receiver_info.insert(id, handler);
+        (id, self.tx_message.clone())
+    }
+
+    /// Spawn runs the CentralAccounting. This should be called after registration of every
+    /// accounting_receiver.
+    pub async fn spawn(mut self) {
```
akichidis, line 74: Not necessary, but to help the developer - I guess - maybe we can do a simple check here in case there is no registered handler and panic in this case (or something like a `debug_assert!`)

network/src/central_accounting.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// The message Bytes read from TCP connection.
+    pub(crate) message: Bytes,
+
+    /// The channel to forward the response.
+    pub(crate) tx_response: Writer,
+}
+
+impl<Handler: MessageHandler> CentralAccounting<Handler> {
+    /// Create a new CentralAccounting.
+    pub fn new(channel_capacity: usize) -> Self {
```
akichidis, line 54: Could you please provide in comments what the `channel_capacity` is being used for?

network/src/tests/receiver_tests.rs
```diff
@@ -49,3 +74,81 @@ async fn receive() {
...
+
+    // Ensure the message gets passed to the channel.
+    let message = rx.recv().await;
+    assert!(message.is_some());
+    let received = message.unwrap();
+    assert_eq!(received, sent);
+}
+
+#[tokio::test]
+async fn accounting_receive() {
```
akichidis, line 101: I would suggest to extract this to its own dedicated file `accounting_receiver_tests` so its clearer to expect where to find the component's tests

network/src/central_accounting.rs
```diff
@@ -0,0 +1,106 @@
...
+            tx_message,
+            tr_message,
+            receiver_info: HashMap::new(),
+        }
+    }
+
+    /// Register will register an AccountingReceiver with the CentralAccounting. This function is
+    /// to be called once for each AccountingReceiver, and it will provide the parameters
+    /// necessary to start an AccountingReceiver.
+    pub async fn register(&mut self, handler: Handler) -> (ReceiverId, Sender<AccountingMessage>) {
```
akichidis, line 66: I believe this doesn't need to be `async` unless there is some specific reason?

network/src/central_accounting.rs
akichidis: Also, probably this doesn't need to be `async`?

network/src/tests/receiver_tests.rs
```diff
@@ -49,3 +74,81 @@ async fn receive() {
...
+    let (receiver_id_1, tx_message_1) = central_accounting.register(test_handler_1).await;
+    let (receiver_id_2, tx_message_2) = central_accounting.register(test_handler_2).await;
+
+    // Start the central accounting.
+    central_accounting.spawn().await;
+
+    // Start the accounting receivers with the provided receiverIDs and tx_forwarding channels.
+    accounting_receiver_1.spawn(receiver_id_1, tx_message_1);
+    accounting_receiver_2.spawn(receiver_id_2, tx_message_2);
+    sleep(Duration::from_millis(50)).await;
```
akichidis, line 127: Ah, I think eventually this can make the test flaky when we end up running in different environments. I know that things so far have been ok, but those assumptions could change (for example, why `50ms` and not `100ms`?) . 

I believe the way to go would be to make the code bellow - which tries to connect to the tcp - to retry a few times (with some intermediate sleep) before failing permanently. 

In any case, leaving like this for now is fine, but we might need to re-think the approach we take in general in such cases

akichidis (commented):
Great work @laura-makdah ! I going to have a look again, but so far looks great!

network/src/accounting_receiver.rs
```diff
@@ -0,0 +1,111 @@
...
+    async fn spawn_runner(
+        socket: TcpStream,
+        peer: SocketAddr,
+        receiver_id: ReceiverId,
+        tx_forwarding: Sender<AccountingMessage>,
+    ) {
+        tokio::spawn(async move {
+            let transport = Framed::new(socket, LengthDelimitedCodec::new());
+            let (tcp_writer, mut reader) = transport.split();
+            let (tx_response, tr_response): (Writer, Receiver<Bytes>) = channel(100);
```
asonnino, line 70: Why a channel capacity of 100?

asonnino, line 70: Do you need to specify the type?

network/src/accounting_receiver.rs
```diff
@@ -0,0 +1,111 @@
...
+
+            while let Some(frame) = reader.next().await {
+                match frame.map_err(|e| NetworkError::FailedToReceiveMessage(peer, e)) {
+                    Ok(message) => {
+                        let accounting_message = AccountingMessage {
+                            receiver_id,
+                            message: message.freeze(),
+                            tx_response: tx_response.clone(),
+                        };
+                        _ = tx_forwarding.send(accounting_message).await;
```
asonnino, line 81: Do we want to unwrap/expect this result?

network/src/central_accounting.rs
asonnino: Should we ensure it is not set to 0?

network/src/receiver.rs
```diff
@@ -69,11 +71,18 @@ impl<Handler: MessageHandler> Receiver<Handler> {
     async fn spawn_runner(socket: TcpStream, peer: SocketAddr, handler: Handler) {
         tokio::spawn(async move {
             let transport = Framed::new(socket, LengthDelimitedCodec::new());
-            let (mut writer, mut reader) = transport.split();
+            let (writer, mut reader) = transport.split();
+
+            let (tx_response, tr_response): (Writer, mspcReceiver<Bytes>) = channel(100);
```
huitseeker, line 76: Nit: the codebase prefix for the receiver of a channel is `rx_`

network/src/receiver.rs
```diff
@@ -5,17 +5,19 @@ use crate::error::NetworkError;
 use async_trait::async_trait;
 use bytes::Bytes;
 use futures::stream::{SplitSink, StreamExt as _};
+use futures::SinkExt;
+use std::borrow::BorrowMut;
 use std::{error::Error, net::SocketAddr};
 use tokio::net::{TcpListener, TcpStream};
+use tokio::sync::mpsc::{channel, Receiver as mspcReceiver, Sender};
```
huitseeker, line 12: Nit: Camel case for the type alias.

network/src/receiver.rs
```diff
@@ -87,4 +96,23 @@ impl<Handler: MessageHandler> Receiver<Handler> {
...
+    /// Spawns a new task that will listen for any responses from the accounting layer and
+    /// forward them to the tcp writer.
+    async fn spawn_response_writer(
+        mut tr_response: mspcReceiver<Bytes>,
+        mut tcp_writer: TcpWriter,
+    ) {
+        tokio::spawn(async move {
+            loop {
+                tokio::select! {
+                   Some(response) = tr_response.recv() => {
```
huitseeker, line 109: This branch has a plausible story of being disabled (because the future is not ready, with the handler not having fired a response) at many passes through the loop. Because this `select!` call has no `else`, this would trigger a panic. Let's avoid it.

network/src/accounting_receiver.rs
huitseeker: we have a `node::CHANNEL_CAPACITY`, but here I'm not against having a specific `networking::CHANNEL_CAPACITY`

network/src/receiver.rs
```diff
@@ -5,17 +5,19 @@ use crate::error::NetworkError;
...
-#[cfg(test)]
-#[path = "tests/receiver_tests.rs"]
-pub mod receiver_tests;
-
 /// Convenient alias for the writer end of the TCP channel.
-pub type Writer = SplitSink<Framed<TcpStream, LengthDelimitedCodec>, Bytes>;
+pub type TcpWriter = SplitSink<Framed<TcpStream, LengthDelimitedCodec>, Bytes>;
+
+/// Convenient alias for the output of the dispatch function.
+pub type Writer = Sender<Bytes>;
```
huitseeker, line 20: The channel for sending back the handler's response is a full-fleged channel. As long as we do not have the ability to stream to a TCP connection (specifically, we do not have multiplexing capabilities), does this make sense? I.o.w. if we were to fire off two handlers and they each could send a stream of repsonse messages of size > 1, do we expect the other end to make sense of this?

Concretely: should we make this a `oneshot::channel` until we have some multiplexing capability?

network/src/receiver.rs
```diff
@@ -69,11 +71,18 @@ impl<Handler: MessageHandler> Receiver<Handler> {
...
+            let (writer, mut reader) = transport.split();
+
+            let (tx_response, tr_response): (Writer, mspcReceiver<Bytes>) = channel(100);
+            Self::spawn_response_writer(tr_response, writer).await;
+
             while let Some(frame) = reader.next().await {
                 match frame.map_err(|e| NetworkError::FailedToReceiveMessage(peer, e)) {
                     Ok(message) => {
-                        if let Err(e) = handler.dispatch(&mut writer, message.freeze()).await {
+                        if let Err(e) = handler
```
huitseeker, line 82: There is an alternative to channels if we are OK with the handler returning a single value (look to my comment above mentioning a `oneshot::channel`). 

```rust
pub struct Outbound<F> {
    future: F,
    writer: TcpWriter,
}

impl<F> Outbound<F> {
    pub fn new(future: F, writer: TcpWriter) -> Self {
        Self { future, writer }
    }
}

impl<F> Future for Outbound<F>
where F: Future<Output = io::Result<Bytes>> + Unpin
{
    type Output = io::Result<())>;

    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        let response = ready!(Pin::new(&mut self.future).poll(cx))?;
        self.writer.send(response).await
        Poll::Ready(Ok(()))
    }
}
```

network/src/accounting_receiver.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+impl AccountingReceiver {
+    /// Create a new AccountingReceiver.
+    pub fn new(address: SocketAddr) -> Self {
+        AccountingReceiver { address }
+    }
+
+    /// Spawn a new network receiver handling connections from any incoming peer.
+    pub fn spawn(self, id: ReceiverId, tx_message: Sender<AccountingMessage>) {
```
huitseeker, line 31: Nit: I would rename to `tx_forwarding` for consistency.

network/src/central_accounting.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fn new(channel_capacity: usize) -> Self {
+        let (tx_message, tr_message) = channel(channel_capacity);
+        CentralAccounting {
+            tx_message,
+            tr_message,
+            receiver_info: HashMap::new(),
+        }
+    }
+
+    /// Register will register an AccountingReceiver with the CentralAccounting. This function is
```
huitseeker, line 63: Does it make sense to register without wanting to start a receiver already? i.o.w, right now when starting a receiver we immediately match a `SocketAddr` and a `handler`, does the ID have to be known by the caller, or could we return the `AccountingReceiver` directly?

One reason I could see would be to reuse the same handler for several receivers, but this doesn't seem to be possible with the current API (nor do I see a use for this). Is there an advantage for not returning the `AccountingReceiver` I haven't seen?

One advantage I see for the direct return is it becomes impossible to create such a receiver without registration (because it hides the `Sender<AccountingMessage>` that would allow one to do so).

huitseeker (changes requested):
Thanks a bunch for this! The central accounting logic makes sense.

Did you run the (local) benchmark? I'm afraid it fails for me, because of the behaviour of `spawn_response_writer` under sender starvation.

Some of my comments on `Receiver` also apply to the similar `AccountingReceiver`, but I have not replicated them there.

network/src/receiver.rs
laura-makdah: Thanks for this comment, what do you mean by unwind our response task?
I assumed that mspc channels behave as oneshot channels and this loop would stop when the channel was closed or dropped.

network/src/receiver.rs
laura-makdah: @huitseeker Are there any other steps a contributor should go through, such as running the benchmarks, when working on Narwhal?

huitseeker: > Are there any other steps a contributor should go through, such as running the benchmarks, when working on Narwhal?

That's it. Of course, we should have integration tests (#75), but the (local) benchmark serves that purpose for us until we do.

laura-makdah:
Closing this because it this approach is now obsolete.

PR #193, opened 2022-04-29
wire the CertificatesBatchResponse to the primary

akichidis:
Following the https://github.com/MystenLabs/narwhal/pull/186 , this PR is wiring in the `CertificatesBatchResponse` responses to the primary in order to be sent to the `block_synchronizer` . Also some small refactoring was done to the `helper` component in order to:
* not process requests with empty block_ids
* respond back for the `CertificatesBatchResponse` with a negative response (certificates not found) when an error in storage is raised (once we have proper erroring response mechanism can convert to that)

### DIFF
--- a/primary/src/helper.rs
+++ b/primary/src/helper.rs
@@ -171,6 +171,11 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
         origin: PublicKey,
         batch_mode: bool,
     ) {
+        if digests.is_empty() {
+            warn!("Request with empty digests received - ignore request");
+            return;
+        }
+
         // get the requestor's address.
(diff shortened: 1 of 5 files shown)


primary/src/helper.rs
```diff
@@ -190,10 +195,7 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
             Ok(certificates) => certificates,
             Err(err) => {
                 error!("Error while retrieving certificates: {err}");
-
-                // just return at this point since we have no way
-                // to communicate to the requestor an error.
```
huitseeker, line 195: Nit; I liked this comment :)

huitseeker (approved):
Thanks!

PR #197, opened 2022-05-03
adds max concurrent requests with towers

lanvidr:
Addresses https://github.com/MystenLabs/sui/issues/5216 with our new GRPC network.

This creates a new config parameter for setting maximum concurrent requests, and applies a limit to the worker to worker and primary to primary servers, which are the untrusted communications. Based on my understanding from reading the code, it seems like one instance of narwhal is either a primary *or* a worker, so whichever it is, it will instantiate only one of these servers, and so the limits don't need to be coordinated. Let me know if this is not accurate. 
Advice is welcome on what to set as the default value.

Upon running the benchmarks with setting the concurrency limit of 500,000 I see this result on my laptop:
```
-----------------------------------------
 SUMMARY:
-----------------------------------------
 + CONFIG:
 Faults: 0 node(s)
 Committee size: 4 node(s)
 Worker(s) per node: 1 worker(s)
 Collocate primary and workers: True
 Input rate: 50,000 tx/s
 Transaction size: 512 B
 Execution time: 7 s

 Header size: 1,000 B
 Max header delay: 200 ms
 GC depth: 50 round(s)
 Sync retry delay: 10,000 ms
 Sync retry nodes: 3 node(s)
 batch size: 500,000 B
 Max batch delay: 200 ms
 Max concurrent requests: 500,000 <---- *new!*

 + RESULTS:
 Consensus TPS: 40,447 tx/s
 Consensus BPS: 20,708,609 B/s
 Consensus latency: 893 ms

 End-to-end TPS: 39,706 tx/s
 End-to-end BPS: 20,329,690 B/s
 End-to-end latency: 1,030 ms
```

then when setting the concurrency limit to 2:
```-----------------------------------------
 SUMMARY:
-----------------------------------------
 + CONFIG:
 Faults: 0 node(s)
 Committee size: 4 node(s)
 Worker(s) per node: 1 worker(s)
 Collocate primary and workers: True
 Input rate: 50,000 tx/s
 Transaction size: 512 B
 Execution time: 1 s

 Header size: 1,000 B
 Max header delay: 200 ms
 GC depth: 50 round(s)
 Sync retry delay: 10,000 ms
 Sync retry nodes: 3 node(s)
 batch size: 500,000 B
 Max batch delay: 200 ms
 Max concurrent requests: 2

 + RESULTS:
 Consensus TPS: 9,267 tx/s
 Consensus BPS: 4,744,495 B/s
 Consensus latency: 525 ms

 End-to-end TPS: 7,636 tx/s
 End-to-end BPS: 3,909,456 B/s
 End-to-end latency: 661 ms
 ```
The performance noticeably decreases when the concurrency limit is very low.

### DIFF
--- a/worker/src/worker.rs
+++ b/worker/src/worker.rs
@@ -209,7 +209,7 @@ impl<PublicKey: VerifyingKey> Worker<PublicKey> {
             tx_client_helper,
             tx_processor,
         }
-        .spawn(address);
+        .spawn(address, self.parameters.max_concurrent_requests);
 
         // The `Helper` is dedicated to reply to batch requests from other workers.
         Helper::spawn(
@@ -294,8 +294,10 @@ struct WorkerReceiverHandler<PublicKey: VerifyingKey> {
 }
 
 impl<PublicKey: VerifyingKey> WorkerReceiverHandler<PublicKey> {
-    fn spawn(self, address: SocketAddr) {
+    fn spawn(self, address: SocketAddr, max_concurrent_requests: usize) {
+        let concurrent_limit = tower::limit::ConcurrencyLimitLayer::new(max_concurrent_requests);
         let service = tonic::transport::Server::builder()
+            .layer(concurrent_limit)
             .add_service(WorkerToWorkerServer::new(self))
             .serve(address);
         tokio::spawn(service);
--- a/worker/src/worker.rs
+++ b/worker/src/worker.rs
@@ -295,9 +295,8 @@ struct WorkerReceiverHandler<PublicKey: VerifyingKey> {
 
 impl<PublicKey: VerifyingKey> WorkerReceiverHandler<PublicKey> {
     fn spawn(self, address: SocketAddr, max_concurrent_requests: usize) {
-        let concurrent_limit = tower::limit::ConcurrencyLimitLayer::new(max_concurrent_requests);
         let service = tonic::transport::Server::builder()
-            .layer(concurrent_limit)
+            .concurrency_limit_per_connection(max_concurrent_requests)
             .add_service(WorkerToWorkerServer::new(self))
             .serve(address);
         tokio::spawn(service);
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: worker/src/worker.rs [hunk 1]
```diff
@@ -294,8 +294,10 @@ struct WorkerReceiverHandler<PublicKey: VerifyingKey> {
 }
 
 impl<PublicKey: VerifyingKey> WorkerReceiverHandler<PublicKey> {
-    fn spawn(self, address: SocketAddr) {
+    fn spawn(self, address: SocketAddr, max_concurrent_requests: usize) {
+        let concurrent_limit = tower::limit::ConcurrencyLimitLayer::new(max_concurrent_requests);
         let service = tonic::transport::Server::builder()
+            .layer(concurrent_limit)
```
bmwill, line 300: Instead of building a `ConcurrencyLimitLayer` we can actually use the [concurrency_limit_per_connection](https://docs.rs/tonic/latest/tonic/transport/server/struct.Server.html#method.concurrency_limit_per_connection) function on the Server builder directly, assuming i'm understanding the semantics you're after

akichidis: That's interesting. I thought that the `ConcurrencyLimitLayer` will be applied across all connections - not per connection (but per `service` though). @laura-makdah I think it would help clarifying the intention so we make sure we are aligned here.

akichidis:
@laura-makdah I've read the issue here https://github.com/MystenLabs/sui/issues/5216 , are we also looking to do prioritisation as well on follow-up PRs?

laura-makdah:
@akichidis There is no immediate need for prioritization, and now that our network has changed to GRPC, the context of the original ticket is a bit different, since it is easer to layer functionality using these libraries, we don't need to build another layer from scratch anymore. Thanks to Brandon!

akichidis (approved):
LGTM 👍  Nice work confirming via the benchmarks as well!

huitseeker:
Awesome work @laura-makdah !

Issue #163, opened 2022-04-26
[block_synchronizer] do not synchronise existing in store headers & payload


akichidis:
### Story

Parent story: https://github.com/MystenLabs/narwhal/issues/67

### Description

`block_synchronizer` doesn't currently check at all about the state of the `headers` or the `payloads` that need to be synchronized. More specifically, if the required entities already exist in the local storage, `block_synchronizer` will not know about and will always initiate the protocol to fetch from peers.

To disambiguate things we should always check within the `block_synchronizer` whether the requested quantities are already available in storage and just proceed fetching only any ones that are missing.

For the `SynchronizeBlockHeaders` command, the response should be enhanced with some extra value to denote whether the certificate's payload is already synchronized or needs further synchronization. For example, at the moment we respond back to the consumer for each certificate result with a `Result<Certificate, Error>` response. We could probably expand this with a structure:
```
struct BlockHeader {
    certificate: Certificate,
    fetched_from_storage: bool // true if batch digests exist in the payload_store, false otherwise
}

Result<BlockHeader, Error>
```

PR #198, opened 2022-05-03
[block_synchronizer] do not synchronise existing in store headers & payload

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/163

This PR is introducing some extra logic on the `block_synchronizer` to first query in the storage for availability before synchronizing either a `certificate` or a `payload`.

**Command::SynchronizeBlockHeaders**
If a certificate is found in local storage, then the response is immediately sent to the consumer and is excluded from synching. If there is any certificate left to synchronize, then the rest of the process will be followed. If all can be served via the local storage, then no synching process will be triggered.

**Command::SynchronizeBlockPayload**
Similar to the `Command::SynchronizeBlockHeaders` case, the `payload_store` will be queried in order to figure out whether the payload is available. The assumption is that if a certificate's payload batches can all be found to the `payload_store` , then the batch should be available to the workers - that's how the rest of the codebase treats the `payload_store`

### DIFF
--- a/primary/src/block_synchronizer/mod.rs
+++ b/primary/src/block_synchronizer/mod.rs
@@ -92,16 +92,16 @@ pub enum Command<PublicKey: VerifyingKey> {
 enum State<PublicKey: VerifyingKey> {
     HeadersSynchronized {
         request_id: RequestID,
-        certificates: HashMap<CertificateDigest, Result<BlockHeader<PublicKey>, SyncError>>,
+        certificates: HashMap<CertificateDigest, BlockSynchronizeResult<BlockHeader<PublicKey>>>,
     },
     PayloadAvailabilityReceived {
         request_id: RequestID,
-        certificates: HashMap<CertificateDigest, Result<BlockHeader<PublicKey>, SyncError>>,
+        certificates: HashMap<CertificateDigest, BlockSynchronizeResult<BlockHeader<PublicKey>>>,
         peers: Peers<PublicKey, Certificate<PublicKey>>,
     },
     PayloadSynchronized {
         request_id: RequestID,
-        result: Result<BlockHeader<PublicKey>, SyncError>,
+        result: BlockSynchronizeResult<BlockHeader<PublicKey>>,
     },
 }
 
@@ -478,10 +478,10 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
                 // Reply back directly with the found from storage certificates
                 let futures: Vec<_> = found
                     .into_iter()
-                    .filter(|e| e.1.is_some())
-                    .map(|(_, c)| {
+                    .flat_map(|(_, c)| c)
+                    .map(|c| {
                         respond_to.send(Ok(BlockHeader {
-                            certificate: c.unwrap(),
+                            certificate: c,
                             fetched_from_storage: true,
                         }))
                     })
@@ -543,7 +543,7 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
 
         for r in join_all(futures).await {
             if r.is_err() {
-                error!("Couldn't send message to channel [{:?}]", r.err().unwrap());
+                error!("Couldn't send message to channel {:?}", r.err());
             }
         }
 
@@ -882,14 +882,14 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
         peers: &Peers<PublicKey, Certificate<PublicKey>>,
         block_ids: Vec<CertificateDigest>,
         timeout: bool,
-    ) -> HashMap<CertificateDigest, Result<BlockHeader<PublicKey>, SyncError>> {
+    ) -> HashMap<CertificateDigest, BlockSynchronizeResult<BlockHeader<PublicKey>>> {
         let mut certificates_by_id: HashMap<CertificateDigest, Certificate<PublicKey>> = peers
             .unique_values()
             .into_iter()
             .map(|c| (c.digest(), c))
             .collect();
 
-        let mut result: HashMap<CertificateDigest, Result<BlockHeader<PublicKey>, SyncError>> =
+        let mut result: HashMap<CertificateDigest, BlockSynchronizeResult<BlockHeader<PublicKey>>> =
             HashMap::new();
 
         for block_id in block_ids {
--- a/primary/src/block_synchronizer/mod.rs
+++ b/primary/src/block_synchronizer/mod.rs
@@ -488,8 +488,8 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
                     .collect();
 
                 for r in join_all(futures).await {
-                    if r.is_err() {
-                        error!("Couldn't send message to channel [{:?}]", r.err().unwrap());
+                    if let Err(err) = r {
+                        error!("Couldn't send message to channel [{:?}]", err);
                     }
                 }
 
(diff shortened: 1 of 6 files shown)


primary/src/block_synchronizer/mod.rs
```diff
@@ -40,7 +43,17 @@ pub mod responses;
...
-type ResultSender<T> = Sender<BlockSynchronizeResult<Certificate<T>>>;
+#[derive(Debug, Clone)]
+pub struct BlockHeader<PublicKey: VerifyingKey> {
+    certificate: Certificate<PublicKey>,
+    /// It designates whether the requested quantity (either the certificate
+    /// or the payload) has been retrieved via the local storage. If true,
+    /// the it used the storage. If false, then it has been fetched via
+    /// the peers.
+    fetched_from_storage: bool,
+}
```
bmwill, line 54: I'm still building context on Narwhal but when I see something called a `BlockHeader` it makes me things this is something that would be shared to other peers. Given the actual makeup of this type (ie the presence of the `fetched_from_storage` bit) I take it this is not the case and is intended to remain a local type and never transmitted on the wire, correct?

akichidis: That is correct. I totally see what you mean and when I named it like this I kind of had a similar feeling. I originally wanted to name something like `BlockHeaderResult` but then there would probably be confusion around semantics (e.x this is not a Rust `Result`). I am really open to suggestions.

primary/src/block_synchronizer/mod.rs
```diff
@@ -425,6 +455,101 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
...
+                    Vec<(CertificateDigest, Option<Certificate<PublicKey>>)>,
+                ) = block_ids
+                    .into_iter()
+                    .zip(certificates)
+                    .partition(|f| f.1.is_some());
+
+                // Reply back directly with the found from storage certificates
+                let futures: Vec<_> = found
+                    .into_iter()
+                    .filter(|e| e.1.is_some())
```
huitseeker, line 481: You just did that `filter` in `partition`. Consider e.g. this instead:
```rust
                let futures: Vec<_> = found
                    .into_iter()
                    .flat_map(|(_, c)| c)
                    .map(|c| {
                        respond_to.send(Ok(BlockHeader {
                            certificate: c,
                            fetched_from_storage: true,
                        }))
                    })
                    .collect();
```

primary/src/block_synchronizer/mod.rs
```diff
@@ -425,6 +455,101 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
...
+                            certificate: c.unwrap(),
+                            fetched_from_storage: true,
+                        }))
+                    })
+                    .collect();
+
+                for r in join_all(futures).await {
+                    if r.is_err() {
+                        error!("Couldn't send message to channel [{:?}]", r.err().unwrap());
+                    }
```
huitseeker, line 493: Consider this use of `Display` that does not require `unwrap`:

```rust
            if let Err(err) = r {
                error!("Couldn't send message to channel [{}]", err);
            }
```

primary/src/block_synchronizer/mod.rs
```diff
@@ -425,6 +455,101 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
...
+                    certificate,
+                    fetched_from_storage: true,
+                })));
+            }
+        }
+
+        for r in join_all(futures).await {
+            if r.is_err() {
+                error!("Couldn't send message to channel [{:?}]", r.err().unwrap());
+            }
```
huitseeker, line 547: Consider this use of `Display` that does not require `unwrap`:

```rust
            if let Err(err) = r {
                error!("Couldn't send message to channel [{}]", err);
            }
```

primary/src/block_synchronizer/mod.rs
```diff
@@ -754,21 +882,27 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
         peers: &Peers<PublicKey, Certificate<PublicKey>>,
         block_ids: Vec<CertificateDigest>,
         timeout: bool,
-    ) -> HashMap<CertificateDigest, Result<Certificate<PublicKey>, SyncError>> {
+    ) -> HashMap<CertificateDigest, Result<BlockHeader<PublicKey>, SyncError>> {
```
huitseeker, line 885: You could use `BlockSynchronizeResult` here.

primary/src/block_synchronizer/mod.rs
```diff
@@ -754,21 +882,27 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
...
         let mut certificates_by_id: HashMap<CertificateDigest, Certificate<PublicKey>> = peers
             .unique_values()
             .into_iter()
             .map(|c| (c.digest(), c))
             .collect();
 
-        let mut result: HashMap<CertificateDigest, Result<Certificate<PublicKey>, SyncError>> =
+        let mut result: HashMap<CertificateDigest, Result<BlockHeader<PublicKey>, SyncError>> =
```
huitseeker, line 892: You could use `BlockSynchronizeResult` here.

primary/src/block_synchronizer/mod.rs
```diff
@@ -79,16 +92,16 @@ pub enum Command<PublicKey: VerifyingKey> {
...
     PayloadAvailabilityReceived {
         request_id: RequestID,
-        certificates: HashMap<CertificateDigest, Result<Certificate<PublicKey>, SyncError>>,
+        certificates: HashMap<CertificateDigest, Result<BlockHeader<PublicKey>, SyncError>>,
         peers: Peers<PublicKey, Certificate<PublicKey>>,
     },
     PayloadSynchronized {
         request_id: RequestID,
-        result: Result<Certificate<PublicKey>, SyncError>,
+        result: Result<BlockHeader<PublicKey>, SyncError>,
```
huitseeker, line 104: You could use `BlockSynchronizeResult` here.

huitseeker (commented):
Thanks Tasos, the lookups look great!

 The `BlockSynchronizeResult` type and maintaining its provenance field seem unnecessary: could we do without?

primary/src/block_synchronizer/mod.rs
```diff
@@ -40,7 +43,17 @@ pub mod responses;
 /// proceed to next state.
 const CERTIFICATE_RESPONSES_RATIO_THRESHOLD: f32 = 0.5;
 
-type ResultSender<T> = Sender<BlockSynchronizeResult<Certificate<T>>>;
+#[derive(Debug, Clone)]
+pub struct BlockHeader<PublicKey: VerifyingKey> {
```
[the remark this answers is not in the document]
akichidis: I've addressed the rest of comments so If you are ok to keep the `BlockHeader` as is (given the above) do you mind approving as this will allow me to progress with follow up work. Otherwise let's discuss further about the `BlockHeader`.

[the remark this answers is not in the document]
huitseeker: I'm ok with temporizing on this. Please open an issue: essentially, we want `BlockHeader` to take different branches depending on its flag, and ideally not just in tests.

There are ultimately several prettier ways of doing this, but we'll discuss the stylistic details once we have a clearer picture of the control flow.

primary/src/block_synchronizer/mod.rs
huitseeker: This one is still live.

primary/src/block_synchronizer/mod.rs
```diff
@@ -425,6 +455,101 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
...
+                futures.push(respond_to.send(Ok(BlockHeader {
+                    certificate,
+                    fetched_from_storage: true,
+                })));
+            }
+        }
+
+        for r in join_all(futures).await {
+            if r.is_err() {
+                error!("Couldn't send message to channel {:?}", r.err());
```
huitseeker, line 546:
```suggestion
            if let Err(err) = r {
                error!("Couldn't send message to channel {:?}", err);
```

huitseeker (approved):
Thanks a lot Tasos!

Issue #201, opened 2022-05-03
tracing: introduce spans on the helper component


akichidis:
### Description

Introduce tracing for the `helper` component in `primary` (as a "test" case) to trace:
* incoming Certificate requests
* errors raised

I will not try to solve the across codebase tracing, but take this as a showcase / first adopter and see how this could fit our case. Ideally we want to reduce error logging or at least be able to track via spans as well.

PR #204, opened 2022-05-04
introduce span tracing for helper

akichidis:
Resolves: #201 

This PR constitutes a first attempt to introduce some tracing via spans on the helper module. An error type has been introduced in order to make the capturing of errors more natural via the tracing framework instead of always having to print custom messages. Ideally, we would like to follow semantic conventions as those mentioned on the [opentelemetry](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions) to get our application closer to the distributed tracing patterns - for now followed minimal configuration.

### DIFF
--- a/primary/src/helper.rs
+++ b/primary/src/helper.rs
@@ -3,16 +3,29 @@
 // SPDX-License-Identifier: Apache-2.0
 use crate::{primary::PrimaryMessage, PayloadToken};
 use config::{Committee, WorkerId};
-use crypto::traits::VerifyingKey;
+use crypto::traits::{EncodeDecodeBase64, VerifyingKey};
 use network::PrimaryNetwork;
-use store::Store;
+use store::{Store, StoreError};
+use thiserror::Error;
 use tokio::sync::mpsc::Receiver;
-use tracing::{error, instrument, warn};
+use tracing::{error, instrument};
 use types::{BatchDigest, Certificate, CertificateDigest};
 
 #[cfg(test)]
 #[path = "tests/helper_tests.rs"]
 mod helper_tests;
+
+#[derive(Debug, Error)]
+enum HelperError {
+    #[error("Received message from unknown authority {0}")]
+    UnknownAuthority(String),
+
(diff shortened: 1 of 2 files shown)


primary/src/helper.rs
```diff
@@ -60,7 +73,7 @@ impl<PublicKey: VerifyingKey> Helper<PublicKey> {
                 // the data source (dictated by the digests parameter). The results
                 // will be emitted one by one to the consumer.
                 PrimaryMessage::CertificatesRequest(digests, origin) => {
-                    self.process_certificates(digests, origin, false).await;
+                    let _ = self.process_certificates(digests, origin, false).await;
```
bmwill, line 76: do we want to log when these errors occur instead of ignoring them?

bmwill (approved):
small question but looks good otherwise

primary/src/helper.rs
akichidis: Actually the reason I introduced those errors is specifically to have a "transparent" way of capturing them instead of me having to introduce `error!` messages. For the `process_certificates` the macro I've used is the following:
```
#[instrument(level="debug", skip_all, fields(certificate_ids = ?digests, mode = batch_mode), err)]
```

in the end you'll see that the `err` has been included. For methods that return a `Result`, by including the attribute `err` on the `instrument` macro we are automatically capturing the returned error (see docs [here](https://docs.rs/tracing/latest/tracing/attr.instrument.html))

So based on the setup of the methods those errors will be captured and printed to the spans / logs as well.

To be honest, I believe that we want to use this pattern in various places so we do have a common nice way of reporting errors without having to introduce custom `err` messages and also ensure that errors will be captured on the span (later on distributed tracing will come up as failures which is something we want)

bmwill: Thanks for a more detailed explanation! I think that makes quite a bit of sense.

huitseeker (approved):
This looks fantastic in all respects, thanks!

Issue #183, opened 2022-04-28
[get_collections] wire block_synchronizer in block_waiter component


akichidis:
### Story

Parent story: https://github.com/MystenLabs/narwhal/issues/67

### Description

At the moment when `block_waiter` is not able to find a certificate in the local storage it returns an error about non-found block. We want to ensure that this won't happen and we'll always perform a best effort to have the certificate available before permanently fail. For that reason we would like to use the `block_synchronizer` to:
* fetch missing certificates
* synchronize the payload for the missing certificates

**Technical Approach**
We should inject the `block_synchronizer's` channel sender to be able to send the commands to fetch the missing certificates & sync payloads.

Currently the `block_waiter` is checking to find about whether certificates are available via the `certificate_store`. As the `block_synchronizer` is already doing this check for us - and for the availability of the payload - we can just defer the check to them. So whenever a `GetBlock` or a `GetBlocks` request is received, we can always send a `SynchronizeBlockHeaders` command to the `block_synchronizer` . If the certificates are available, then we'll immediately get the response back. If not, then the process to fetch the certificates will be initiated. 

For the certificates that have been found with the payload available, we don't have to do anything further apart from adding them to a list to fetch the payloads from the workers. For the once that haven't been found we fail. For the ones that have been found, but payload is not synced, then the payload syncing process should be triggered by sending a `SynchronizeBlockPayload` to sync them. Once we have received then responses, then we can proceed to fetch the payloads from the workers.

PR #214, opened 2022-05-06
wire block synchronizer to block waiter (part 1)

akichidis:
Ref: https://github.com/MystenLabs/narwhal/issues/183

This PR is wiring in the `block_synchronizer` , via tokio channels, to the `block_waiter` component to ensure that we are able to:
1. synchronize (fetch) missing certificates via peers
2. synchronize (fetch) missing payload (batches) via peers

The case (1) we don't _really_ expect to happen, as via the protocol we'll expect to have the certificates available via the `read_causal` call.

The case (2) is the most probable one which we mostly care.

However, this PR is tackling case (1) and there will be a follow up PR (part 2) to tackle case (2).

So from now on we'll rely 100% to the `block_synchronizer` to fetch the certificates and we don't have the need to query the `certificate_store` ourselfs - `block_synchronizer` is doing that anyways and will serve us fast cached (via rocksdb) results.

Also the `MockBlockSynchronizer` is introduced as a way to mock the synchronizer's responses without having the need to use the actual component and going through the fuzz of setting up etc. The mock will be extended for the `payload` cases as well

### DIFF
--- a/primary/src/block_synchronizer/mock.rs
+++ b/primary/src/block_synchronizer/mock.rs
@@ -1,0 +1,170 @@
+use crate::block_synchronizer::{BlockHeader, BlockSynchronizeResult, Command};
+use crypto::traits::VerifyingKey;
+use std::collections::HashMap;
+use tokio::sync::{
+    mpsc::{channel, Receiver, Sender},
+    oneshot,
+};
+use types::CertificateDigest;
+
+#[derive(Debug)]
+enum Core<PublicKey: VerifyingKey> {
+    SynchronizeBlockHeaders {
+        block_ids: Vec<CertificateDigest>,
+        times: u32,
+        result: Vec<BlockSynchronizeResult<BlockHeader<PublicKey>>>,
+        ready: oneshot::Sender<()>,
+    },
+    AssertExpectations {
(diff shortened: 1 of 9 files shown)


primary/src/block_synchronizer/mock.rs
```diff
@@ -0,0 +1,170 @@
...
+                result
+            );
+        }
+    }
+}
+
+/// A mock helper for the BlockSynchronizer to help us mock the responses
+/// eliminating the need to wire in the actual BlockSynchronizer when needed
+/// for other components.
+pub struct MockBlockSynchronizer<PublicKey: VerifyingKey> {
```
akichidis, line 100: Unfortunately I had to do the separation between the `MockBlockSynchronizer` and the "inner" struct `MockBlockSynchronizerCore` to allow me have a way to interface with the engine after we have spin up the task. Otherwise I wouldn't be able to perform follow up calls, like the `assert_expectations` since my object would be moved. That complicated things a bit (thanks Rust) but I consider it still straightforward

akichidis:
Following the conversations with @huitseeker about causal completion, I am closing this issue and will open a follow up with a synchronous approach to wait for causally completing missing certificates.

PR #217, opened 2022-05-09
fix: rename header_store to certificate_store

akichidis:
While going through the code in `header_waiter` I came across the `header_store` which is in fact the `certificate_store`. The naming compared to the expected functionality is quite confusing (@huitseeker I saw that you actually fixed this as a bug where the `header_store` was erroneously passed in the past). If it wasn't intentional to leave this as `header_store` let's do the renaming as it makes the semantics clearer (waiting on the `certificate_store` doesn't require syncing the `headers`/ `batches` as well).


huitseeker (approved):
Yup, makes sense to me!

PR #220, opened 2022-05-09
network: use 'multiaddr' everywhere

bmwill:
This patch converts all uses of `SocketAddr` for identifiying the
network address of a node to `Multiaddr`.

In addition this patch leverages the `config::utils::get_available_port`
utility function to improve port allocation in tests.

### DIFF
--- a/executor/src/batch_loader.rs
+++ b/executor/src/batch_loader.rs
@@ -9,16 +9,13 @@ use config::WorkerId;
 use consensus::ConsensusOutput;
 use crypto::traits::VerifyingKey;
 use futures::stream::StreamExt;
-use std::{
-    collections::{HashMap, HashSet},
-    net::SocketAddr,
-};
+use multiaddr::Multiaddr;
+use std::collections::{HashMap, HashSet};
 use store::Store;
 use tokio::{
     sync::mpsc::{channel, Receiver, Sender},
     task::JoinHandle,
 };
-use tonic::transport::Channel;
 use tracing::warn;
 use types::{BatchDigest, BincodeEncodedPayload, ClientBatchRequest, WorkerToWorkerClient};
 use worker::SerializedBatchMessage;
@@ -30,7 +27,7 @@ pub struct BatchLoader<PublicKey: VerifyingKey> {
     /// Receive consensus outputs for which to download the associated transaction data.
     rx_input: Receiver<ConsensusOutput<PublicKey>>,
     /// The network addresses of the consensus workers.
-    addresses: HashMap<WorkerId, SocketAddr>,
+    addresses: HashMap<WorkerId, Multiaddr>,
     /// A map of connections with the consensus workers.
     connections: HashMap<WorkerId, Sender<Vec<BatchDigest>>>,
 }
@@ -40,7 +37,7 @@ impl<PublicKey: VerifyingKey> BatchLoader<PublicKey> {
     pub fn spawn(
         store: Store<BatchDigest, SerializedBatchMessage>,
         rx_input: Receiver<ConsensusOutput<PublicKey>>,
-        addresses: HashMap<WorkerId, SocketAddr>,
+        addresses: HashMap<WorkerId, Multiaddr>,
     ) -> JoinHandle<SubscriberResult<()>> {
         tokio::spawn(async move {
             Self {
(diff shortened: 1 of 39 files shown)


bmwill:
Note that we should not land this PR until https://github.com/MystenLabs/mysten-infra/pull/70 lands and we update this PR to point directly at a commit hosted in the https://github.com/MystenLabs/mysten-infra repo. This PR currently points at my local fork of the mysten-infra repo.

executor/src/batch_loader.rs
```diff
@@ -121,10 +122,9 @@ impl SyncConnection {
 
     /// Main loop keeping the connection with a worker alive and receive batches to download.
     async fn run(&mut self) {
-        let url = format!("http://{}", self.address);
-        let channel = Channel::from_shared(url)
-            .expect("URI should be valid")
-            .connect_lazy();
+        let config = mysten_network::config::Config::new();
+        //TODO don't panic on bad address
+        let channel = config.connect_lazy(&self.address).unwrap();
```
huitseeker, line 127: Yep. Do you know if we can parse earlier @asonnino ?

network/src/primary.rs
```diff
@@ -30,32 +32,33 @@ impl Default for PrimaryNetwork {
...
         self.clients
-            .entry(address)
-            .or_insert_with(|| Self::create_client(address))
+            .entry(address.clone())
+            .or_insert_with(|| Self::create_client(&self.config, address))
             .clone()
     }
 
-    fn create_client(address: SocketAddr) -> PrimaryToPrimaryClient<Channel> {
-        // TODO use TLS
```
huitseeker, line 48: I'd like to not lose this comment, because in the mTLS context, using TLS is not just about a change of protocol, but requires configuring a custom verifier.

network/src/primary.rs
```diff
@@ -164,29 +167,30 @@ impl PrimaryNetwork {
...
         self.clients
-            .entry(address)
-            .or_insert_with(|| Self::create_client(address))
+            .entry(address.clone())
+            .or_insert_with(|| Self::create_client(&self.config, address))
             .clone()
     }
 
-    fn create_client(address: SocketAddr) -> PrimaryToWorkerClient<Channel> {
-        // TODO use TLS
```
huitseeker, line 179: Ditto: I'd like to not lose this comment, because in the mTLS context, using TLS is not just about a change of protocol, but requires configuring a custom verifier.

network/src/worker.rs
```diff
@@ -27,32 +29,33 @@ impl Default for WorkerNetwork {
...
         self.clients
-            .entry(address)
-            .or_insert_with(|| Self::create_client(address))
+            .entry(address.clone())
+            .or_insert_with(|| Self::create_client(&self.config, address))
             .clone()
     }
 
-    fn create_client(address: SocketAddr) -> WorkerToWorkerClient<Channel> {
-        // TODO use TLS
```
huitseeker, line 45: And ditto: I'd like to not lose this comment, because in the mTLS context, using TLS is not just about a change of protocol, but requires configuring a custom verifier.

worker/src/primary_connector.rs
```diff
@@ -42,17 +42,15 @@ impl PrimaryConnector {
...
 pub struct PrimaryClient {
     /// The primary network address.
-    _primary_address: SocketAddr,
+    _primary_address: Multiaddr,
     client: WorkerToPrimaryClient<Channel>,
 }
 
 impl PrimaryClient {
-    pub fn new(address: SocketAddr) -> Self {
-        // TODO maybe use TLS
```
huitseeker, line 51: Ditto: I'd like to not lose this comment, because in the mTLS context, using TLS is not just about a change of protocol, but requires configuring a custom verifier.

huitseeker (approved):
I'm happy with this modulo the switching of pointers to the mysten-infra repo! I left a few comments on the order of more spectacularly :
- I like TODOs referring to #200 to mark places to revisit,
- I'd like to track that we're parsing addresses from configuration a bit late (and this generating panicks), whereas we'd be more robust in parsing all of our config info in a structured object ASAP (/cc @asonnino for executor config)

network/src/primary.rs
bmwill: Opened https://github.com/MystenLabs/mysten-infra/issues/73 for this

Issue #194, opened 2022-04-29
[remove_collections] introduce gRPC endpoint


akichidis:
### Story

Parent story: https://github.com/MystenLabs/narwhal/issues/70

### Description

As part of this task a new gRPC endpoint should be introduced to `remove_collections`. The description of the endpoint can be found on the document here: https://www.notion.so/mystenlabs/Narwhal-Reference-Mempool-API-6e6268556f1249bc958d32f4bd1d5b77#889a21a374b74a9d8ee38fc40f7489b2

The endpoint should call the [block_remover](https://github.com/MystenLabs/narwhal/blob/fecbfc9385e5599aed504d2b63de4b8763d4b6a2/primary/src/block_remover.rs#L77) to achieve that.

huitseeker:
closed in #224

PR #224, opened 2022-05-10
Add RemoveCollections endpoint to Consensus API

arun-koshy:
More info can be found in [Issue#194](https://github.com/MystenLabs/narwhal/issues/194)

Started and reverted some refactoring of `BlockRemover` but opened [Issue#222](https://github.com/MystenLabs/sui/issues/5262) for future improvements.

Still need to add an error case to the integration test which I will need some help from @akichidis as my attempts have not returned an error.

### DIFF
--- a/primary/src/grpc_server/mod.rs
+++ b/primary/src/grpc_server/mod.rs
@@ -14,8 +14,8 @@ mod validator;
 pub struct ConsensusAPIGrpc {
     socket_addr: Multiaddr,
     tx_get_block_commands: Sender<BlockCommand>,
-    get_collections_timeout: Duration,
     tx_block_removal_commands: Sender<BlockRemoverCommand>,
+    get_collections_timeout: Duration,
     remove_collections_timeout: Duration,
 }
 
@@ -23,16 +23,16 @@ impl ConsensusAPIGrpc {
     pub fn spawn(
         socket_addr: Multiaddr,
         tx_get_block_commands: Sender<BlockCommand>,
-        get_collections_timeout: Duration,
         tx_block_removal_commands: Sender<BlockRemoverCommand>,
+        get_collections_timeout: Duration,
         remove_collections_timeout: Duration,
     ) {
         tokio::spawn(async move {
             let _ = Self {
                 socket_addr,
                 tx_get_block_commands,
-                get_collections_timeout,
                 tx_block_removal_commands,
+                get_collections_timeout,
                 remove_collections_timeout,
             }
             .run()
@@ -44,8 +44,8 @@ impl ConsensusAPIGrpc {
     async fn run(&mut self) -> Result<(), Box<dyn std::error::Error>> {
         let narwhal = NarwhalValidator::new(
             self.tx_get_block_commands.to_owned(),
-            self.get_collections_timeout,
             self.tx_block_removal_commands.to_owned(),
+            self.get_collections_timeout,
             self.remove_collections_timeout,
         );
 
--- a/primary/src/grpc_server/mod.rs
+++ b/primary/src/grpc_server/mod.rs
@@ -3,7 +3,7 @@
 use std::time::Duration;
 
 use self::validator::NarwhalValidator;
-use crate::BlockCommand;
+use crate::{BlockCommand, BlockRemoverCommand};
 use tokio::sync::mpsc::Sender;
 use tonic::transport::Server;
 use tracing::error;
@@ -15,6 +15,8 @@ pub struct ConsensusAPIGrpc {
     socket_addr: String,
     tx_get_block_commands: Sender<BlockCommand>,
     get_collections_timeout: Duration,
+    tx_block_removal_commands: Sender<BlockRemoverCommand>,
+    remove_collections_timeout: Duration,
 }
 
 impl ConsensusAPIGrpc {
@@ -22,12 +24,16 @@ impl ConsensusAPIGrpc {
         socket_addr: String,
         tx_get_block_commands: Sender<BlockCommand>,
         get_collections_timeout: Duration,
+        tx_block_removal_commands: Sender<BlockRemoverCommand>,
+        remove_collections_timeout: Duration,
     ) {
         tokio::spawn(async move {
             let _ = Self {
                 socket_addr,
                 tx_get_block_commands,
                 get_collections_timeout,
+                tx_block_removal_commands,
+                remove_collections_timeout,
             }
             .run()
             .await
@@ -40,6 +46,8 @@ impl ConsensusAPIGrpc {
         let narwhal = NarwhalValidator::new(
             self.tx_get_block_commands.to_owned(),
             self.get_collections_timeout,
+            self.tx_block_removal_commands.to_owned(),
+            self.remove_collections_timeout,
         );
 
         Server::builder()
--- a/primary/src/grpc_server/validator.rs
+++ b/primary/src/grpc_server/validator.rs
@@ -9,8 +9,9 @@ use tokio::{
 };
 use tonic::{Request, Response, Status};
 use types::{
-    BatchMessageProto, BlockError, CollectionRetrievalResult, Collections, Empty,
-    GetCollectionsResponse, Validator,
+    BatchMessageProto, BlockError, CertificateDigest, CertificateDigestProto,
+    CollectionRetrievalResult, Empty, GetCollectionsRequest, GetCollectionsResponse,
+    RemoveCollectionsRequest, Validator,
 };
 
 #[derive(Debug)]
@@ -41,17 +42,12 @@ impl NarwhalValidator {
 impl Validator for NarwhalValidator {
     async fn remove_collections(
         &self,
-        request: Request<Collections>,
+        request: Request<RemoveCollectionsRequest>,
     ) -> Result<Response<Empty>, Status> {
         let collection_ids = request.into_inner().collection_ids;
         let remove_collections_response = if !collection_ids.is_empty() {
             let (tx_remove_block, mut rx_remove_block) = channel(1);
-            let mut ids = vec![];
-            for collection_id in collection_ids {
-                ids.push(collection_id.try_into().map_err(|err| {
-                    Status::invalid_argument(format!("Could not serialize: {:?}", err))
-                })?);
-            }
+            let ids = parse_certificate_digests(collection_ids)?;
             self.tx_block_removal_commands
                 .send(BlockRemoverCommand::RemoveBlocks {
                     ids,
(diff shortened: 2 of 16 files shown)


primary/src/grpc_server/validator.rs
```diff
@@ -2,40 +2,89 @@
...
+    ) -> Result<Response<Empty>, Status> {
+        let collection_ids = request.into_inner().collection_ids;
+        let remove_collections_response = if !collection_ids.is_empty() {
+            let (tx_remove_block, mut rx_remove_block) = channel(1);
+            let mut ids = vec![];
+            for collection_id in collection_ids {
+                ids.push(collection_id.try_into().map_err(|err| {
+                    Status::invalid_argument(format!("Could not serialize: {:?}", err))
+                })?);
+            }
```
akichidis, line 54: Since we are performing the exact same conversions for [get_collections](https://github.com/MystenLabs/narwhal/pull/224/files#diff-0a5c9d58935199886da37e8caaccb463c22ffc45caff1340e467f819f0616d7fR92) , maybe we could introduce a method to reuse in both places?

primary/src/grpc_server/validator.rs
```diff
@@ -2,40 +2,89 @@
...
+            self.tx_block_removal_commands
+                .send(BlockRemoverCommand::RemoveBlocks {
+                    ids,
+                    sender: tx_remove_block,
+                })
+                .await
+                .map_err(|err| Status::internal(format!("Send Error: {:?}", err)))?;
```
akichidis, line 61: If I understand correctly this message will be visible to the consumer of the API. Maybe it won't make much sense to expose the "internal" errors (e.x couldn't send a message to a tokio channel) to the client? We could come up with something generic. Not an issue for now, but just highlighting.

primary/src/grpc_server/validator.rs
```diff
@@ -2,40 +2,89 @@
...
+                    Err(remove_block_error) => Err(Status::internal(format!(
+                        "Removal Error: {:?}",
+                        remove_block_error.error
+                    ))),
+                },
+                Ok(None) => Err(Status::internal(
+                    "Removal channel closed, no result has been received.",
+                )),
+                Err(_) => Err(Status::internal(
+                    "Timeout, no result has been received in time",
```
akichidis, line 74: The block remover is also responding with a `timeout` error. We might want to emit only one timeout error.

primary/tests/integration_tests.rs
```diff
@@ -14,22 +14,183 @@ use test_utils::{
...
+    });
+    let response = client.remove_collections(request).await.unwrap();
+    let actual_result = response.into_inner();
+
+    assert_eq!(Empty {}, actual_result);
+    for &block_id in collection_ids.iter() {
+        assert!(
+            store
+                .certificate_store
+                .read(block_id)
```
akichidis, line 162: In case you are not aware you can also read stuff in bulk via the `read_all` method. Nothing wrong of course with this way as well

Example: https://github.com/MystenLabs/narwhal/blob/b7d2dfdd6394afa31cb910be0ff53efe7c6a5578/primary/src/helper.rs#L205

types/proto/narwhal.proto
```diff
@@ -96,8 +96,8 @@ message CollectionRetrievalResult {
     }
 }
 
-message GetCollectionsRequest {
-    // List of collections to be retreived.
+message Collections {
```
akichidis, line 99: I definitely see the point of re-using a message for both functionalities, I am just a bit skeptical on the convention though. Since this is a _message_ the `Collections` on its own doesn't mean much to be honest and in most examples I've seen conventions closer to e.x `GetCollectionsRequest` , `RemoveCollectionsRequest`. I believe keeping those separate might benefit us for future evolution (in case those differentiate as well)

types/src/primary.rs
```diff
@@ -520,21 +520,37 @@ pub struct BatchMessage {
     pub transactions: Batch,
 }
 
+// TODO: refactor BlockRemoverResult and BlockResult into one type shared by get/remove collections.
```
akichidis, line 523: The `block_remover` and the `block_waiter` were not really designed to be aligned on their responses - maybe there is an opportunity to do so - since they are having a different role. I have no problem trying to get those aligned, but let's keep in mind that we don't want to create too much dependency.

akichidis (commented):
This is great! Thanks so much @arun-koshy . Let's sync on the error cases.

Could you please also include the `remove_collections_timeout` property to this [README](https://github.com/MystenLabs/narwhal/blob/main/benchmark/README.md?plain=1#L47) as well?

types/src/primary.rs
arun-koshy: That is fair, I will remove this todo. Do you think we should still return one error per collection in the `block_remover` case? Currently with everything being handled in batches is it possible for some collections to be deleted and some to fail?

primary/tests/integration_tests.rs
arun-koshy: I was not, thanks!

primary/src/grpc_server/validator.rs
arun-koshy: Sorry can you clarify this? Only one timeout will return either the block remover which will get surfaced and return as part of this branch 

```
Err(remove_block_error) => Err(Status::internal(format!(
                        "Removal Error: {:?}",
                        remove_block_error.error
                    ))),
```

or the grpc request timeout which you commented on above. Whichever comes first will be returned.

primary/src/grpc_server/validator.rs
arun-koshy: Hmmm... yeah this might require some more thought. But I will leave the error for now as we are building and maybe go back and make the errors more generic as to not expose too much to the customer.

types/proto/narwhal.proto
arun-koshy: You are correct `GetCollectionRequest`/`RemoveCollectionRequest` is the convention. I will switch back to that format instead of reusing the `Collections` message.

primary/src/grpc_server/validator.rs
akichidis: If you could open an issue for this it would be great so we make sure we track it

primary/src/grpc_server/validator.rs
akichidis: Yeah, basically I meant that the consumer will see two different cases of timeout - either from the block_remover or the endpoint it self. That might be a bit confusing . As a client I would expect a "timeout" error to be a unique one - the operation timed out . Again maybe not big a deal and something we can look into afterwards.

primary/src/grpc_server/validator.rs
arun-koshy: Opened [Issue#234](https://github.com/MystenLabs/narwhal/issues/234)

primary/src/grpc_server/validator.rs
huitseeker: Or we could probe for a `BlockRemoverError` with second field matching `BlockRemoverErrorType::Timeout` above. The pattern would look like this, IIUC:
```
Ok(Some(Err(BlockRemoverError(_,BlockRemoverErrorType::Timeout))))
```

primary/src/grpc_server/validator.rs
```diff
@@ -2,57 +2,97 @@
...
+            let (tx_remove_block, mut rx_remove_block) = channel(1);
+            let ids = parse_certificate_digests(collection_ids)?;
+            self.tx_block_removal_commands
+                .send(BlockRemoverCommand::RemoveBlocks {
+                    ids,
+                    sender: tx_remove_block,
+                })
+                .await
+                .map_err(|err| Status::internal(format!("Send Error: {:?}", err)))?;
+            match timeout(self.remove_collections_timeout, rx_remove_block.recv()).await {
```
huitseeker, line 58: You could match on `timeout(...).await.map_err(|_err| Status::internal("Timeout ..."))` to remove one level of right-nesting from this.

types/src/primary.rs
```diff
@@ -527,21 +527,36 @@ pub struct BatchMessage {
...
+
+#[derive(Clone, Debug)]
+pub struct BlockRemoverError {
+    pub ids: Vec<CertificateDigest>,
+    pub error: BlockRemoverErrorType,
+}
+
+// TODO: refactor BlockError & BlockRemoverError to be one type shared by get/remove collections.
+#[derive(Clone, Debug, PartialEq)]
+pub enum BlockRemoverErrorType {
```
huitseeker, line 540: Nit: Traditionally in Rust codebases, this would be named `BlockRemoverErrorKind`.

primary/tests/integration_tests.rs
```diff
@@ -178,14 +170,200 @@ async fn test_get_collections() {
...
+        store
+            .certificate_store
+            .read(block_to_be_removed)
+            .await
+            .unwrap()
+            .is_none(),
+        "Certificate shouldn't exist"
+    );
+
+    // Test remove remaining collections, one collection has its batches intentionally
```
huitseeker, line 329: Thanks!

types/proto/narwhal.proto
```diff
@@ -121,6 +126,8 @@ message Empty {
 service Validator {
     // Returns the collection contents for each requested collection
     rpc GetCollections (GetCollectionsRequest) returns (GetCollectionsResponse);
+    // Removes each collection provided.
```
huitseeker, line 129: Nit "provided collection"

primary/src/grpc_server/mod.rs
```diff
@@ -15,19 +15,25 @@ pub struct ConsensusAPIGrpc {
     socket_addr: Multiaddr,
     tx_get_block_commands: Sender<BlockCommand>,
     get_collections_timeout: Duration,
```
huitseeker, line 17: Nit: the field order could put the timeouts close to each other.

huitseeker (approved):
I've left a few nits, but this looks great, thank you Arun!

primary/src/grpc_server/validator.rs
arun-koshy: Nice catch, thanks!

primary/src/grpc_server/validator.rs
arun-koshy: IIUC you are both stating that we should only surface only one type of timeout error so I added the matching Francois suggested to do so. If this not what you expected please add a comment to [Issue#234](https://github.com/MystenLabs/narwhal/issues/234) and I will update it after this PR. Thanks!

```
Err(remove_block_error)
       if remove_block_error.error == BlockRemoverErrorKind::Timeout =>
       {
              Err(Status::internal("Timeout, no result has been received in time"))
        }
```

Issue #223, opened 2022-05-10
[block_waiter] use block_synchronizer to synchronize missing blocks


akichidis:
### Story

Parent story: https://github.com/MystenLabs/narwhal/issues/67

### Description

Extend the block synchronizer handler to call the `block_synchronizer` to fetch the missing payloads - if missing - and wire in to `block_waiter`.

Also, the `integration_tests` under the gRPC server should be modified to cover now the full successful scenario for retrieving missing collections.

huitseeker:
Is this still relevant @akichidis ?

akichidis:
That should be closed @huitseeker ...hmm weird, it seems that I moved it to "done" to the task board but didn't close the ticket. Doing this now.

PR #227, opened 2022-05-11
Introduce block_synchronizer_handler and wire to block waiter (part 1) #214

akichidis:
Ref: https://github.com/MystenLabs/narwhal/issues/183

This PR is introducing the following:
* a `handler` to allow us access the `block_synchronizer` functionality in a synchronous way
* a `MockBlockSynchronizer` to allow us test more easily the `handler`. This will be used for follow up work around batch sync as well
* the `mockall` library to allow us easily mock traits and structs
* wire in to the `block_waiter` the `handler` to allow us seamlessly retrieve certificates and wait for causal completion (via `core`) for certificates that have been fetched on the fly from peers.

The `handler` will also be re-used during the `read_causal` calls when we detect missing certificates on the DAG and we need to fetch them (from peers) and ensure that are validated and history has been causally completed.

There is also a separate PR opened against this one which is testing end to end the `get_collections` when missing certificates. 

This is `part 1` . There is follow up work https://github.com/MystenLabs/narwhal/issues/223 to add more functionality to `handler` so we can also seamlessly synchronize the missing `payload` when we request them on `block_waiter`.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -39,7 +39,8 @@ node_params = {
     'block_synchronizer': {
         'certificates_synchronize_timeout': '2000ms',
         'payload_synchronize_timeout': '2000ms',
-        'payload_availability_timeout': '2000ms'
+        'payload_availability_timeout': '2000ms',
+        'handler_certificate_deliver_timeout': '2_000ms'
     },
     "consensus_api_grpc": {
         "socket_addr": "/ip4/127.0.0.1/tcp/0/http",
@@ -63,6 +64,9 @@ They are defined as follows:
 * `socket_addr`: The socket address the consensus api gRPC server should be listening to.
 * `get_collections_timeout`: The timeout configuration when requesting batches from workers.
 * `remove_collections_timeout`: The timeout configuration when removing batches from workers.
+* `handler_certificate_deliver_timeout`: When a certificate is fetched on the fly from peers, it is submitted from the block synchronizer handler for further processing to core 
+to validate and ensure parents are available and history is causal complete. This property is the timeout while we wait for core to perform this processes and the certificate to become 
+available to the handler to consume.
 * `max_concurrent_requests`: The maximum number of concurrent requests for primary-to-primary and worker-to-worker messages.
 
 ### Run the benchmark
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -150,6 +150,14 @@ pub struct BlockSynchronizerParameters {
     /// discover who has the payload available for the dictated certificates.
     #[serde(with = "duration_format")]
     pub payload_availability_timeout: Duration,
+    /// When a certificate is fetched on the fly from peers, it is submitted
+    /// from the block synchronizer handler for further processing to core
+    /// to validate and ensure parents are available and history is causal
+    /// complete. This property is the timeout while we wait for core to
+    /// perform this processes and the certificate to become available to
+    /// the handler to consume.
+    #[serde(with = "duration_format")]
+    pub handler_certificate_deliver_timeout: Duration,
 }
 
 impl Default for BlockSynchronizerParameters {
@@ -158,6 +166,7 @@ impl Default for BlockSynchronizerParameters {
             certificates_synchronize_timeout: Duration::from_millis(2_000),
             payload_synchronize_timeout: Duration::from_millis(2_000),
             payload_availability_timeout: Duration::from_millis(2_000),
+            handler_certificate_deliver_timeout: Duration::from_millis(2_000),
         }
     }
 }
@@ -227,6 +236,12 @@ impl Parameters {
             "Remove collections timeout set to {} ms",
             self.consensus_api_grpc
                 .remove_collections_timeout
+                .as_millis()
+        );
+        info!(
+            "Handler certificate deliver timeout set to {} ms",
+            self.block_synchronizer
+                .handler_certificate_deliver_timeout
                 .as_millis()
         );
         info!(
@@ -399,7 +414,8 @@ mod tests {
              "block_synchronizer": {
                  "certificates_synchronize_timeout": "2s",
                  "payload_synchronize_timeout": "3_000ms",
-                 "payload_availability_timeout": "4_000ms"
+                 "payload_availability_timeout": "4_000ms",
+                 "handler_certificate_deliver_timeout": "1_000ms"
              },
              "consensus_api_grpc": {
                  "socket_addr": "/ip4/127.0.0.1/tcp/0/http",
@@ -489,6 +505,9 @@ mod tests {
         ));
         assert!(logs_contain(
             "Synchronize payload (batches) timeout set to 2000 ms"
+        ));
+        assert!(logs_contain(
+            "Handler certificate deliver timeout set to 2000 ms"
         ));
         assert!(logs_contain(
             "Consensus API gRPC Server listening on /ip4/127.0.0.1"
--- a/primary/Cargo.toml
+++ b/primary/Cargo.toml
@@ -31,7 +31,6 @@ tower = "0.4.12"
 crypto = { path = "../crypto" }
 network = { path = "../network" }
 types = { path = "../types" }
-
 store = { git = "https://github.com/mystenlabs/mysten-infra.git", package = "typed-store", rev = "3b7daedf91fd8937dde26e905b8114cac459b866" }
 mysten-network = { git = "https://github.com/mystenlabs/mysten-infra.git", rev = "3b7daedf91fd8937dde26e905b8114cac459b866" }
 
@@ -42,6 +41,7 @@ tempfile = "3.3.0"
 test_utils = { path = "../test_utils"}
 tracing-test = { version = "0.2.1" }
 worker = { path = "../worker" }
+mockall = "0.11.0"
 
 [features]
 benchmark = []
--- a/primary/src/block_synchronizer/handler.rs
+++ b/primary/src/block_synchronizer/handler.rs
@@ -1,0 +1,250 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use crate::{
+    block_synchronizer::{
+        handler::Error::{BlockDeliveryTimeout, BlockNotFound, Internal},
+        BlockSynchronizeResult, Command,
+    },
+    BlockHeader,
+};
+use async_trait::async_trait;
+use crypto::{traits::VerifyingKey, Hash};
+use futures::future::join_all;
+#[cfg(test)]
+use mockall::*;
+use std::time::Duration;
+use store::Store;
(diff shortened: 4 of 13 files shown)


primary/src/block_synchronizer/mod.rs
```diff
@@ -429,7 +432,7 @@ impl<PublicKey: VerifyingKey> BlockSynchronizer<PublicKey> {
         let key = RequestID::from_iter(to_sync.iter());
 
         let message = PrimaryMessage::<PublicKey>::CertificatesBatchRequest {
-            certificate_ids: block_ids,
+            certificate_ids: to_sync.clone(),
```
akichidis, line 435: A nasty bug that discovered during end to end testing in `get_collections` in gRPC.

primary/src/block_synchronizer/handler.rs
```diff
@@ -0,0 +1,248 @@
...
+            .collect();
+
+        join_all(futures).await
+    }
+
+    #[instrument(level = "debug", skip_all)]
+    async fn wait(&self, block_id: CertificateDigest) -> Result<Certificate<PublicKey>, Error> {
+        if let Ok(result) = timeout(
+            self.certificate_deliver_timeout,
+            self.certificate_store.notify_read(block_id),
```
akichidis, line 126: I am using here the `certificate_store` to wait and get the "signal" the the certificate has been processed. However, I would like to replace this eventually by the [certificate_waiter](https://github.com/MystenLabs/narwhal/blob/main/primary/src/certificate_waiter.rs) which deals already with it. Unfortunately the [certificate_waiter](https://github.com/MystenLabs/narwhal/blob/main/primary/src/certificate_waiter.rs) doesn't offer a blocking API and also doesn't allow us to channel the outcome to a provide channel (only outputs to core). Until we have this available, I am using the `certificate_store`

primary/src/block_synchronizer/mock.rs
```diff
@@ -0,0 +1,172 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
```
akichidis, line 1: It might be considered a bit far fetched building this mock, but I want to avoid writing for every test code to wait on channels and handling sending back results. Will have to test more cases on follow up PRs for the batch synchronization as well and this mock will also be re-used.

In general I would look into building a more generic mock for testing async components that communicate via channels

primary/src/block_synchronizer/mock.rs
```diff
@@ -0,0 +1,172 @@
...
+            })
+            .await
+            .expect("Failed to send mock expectation");
+
+        Self::await_channel(rx).await;
+    }
+
+    /// Asserts that all the expectations have been fulfilled and no
+    /// expectation has been left without having been called.
+    pub async fn assert_expectations(&self) {
```
akichidis, line 150: One of the reasons that I had to separate to `MockBlockSynchronizer` and the `MockBlockSynchronizerCore` is because I wanted to have a way to continue calling methods like this after spawning the async task (which always consumes via `move` command the object and then weren't able to call it further on tests)

benchmark/README.md
```diff
@@ -55,6 +56,7 @@ They are defined as follows:
 * `certificates_synchronize_timeout`: The timeout configuration when requesting certificates from peers.
 * `payload_synchronize_timeout`: Timeout when has requested the payload for a certificate and is waiting to receive them.
 * `payload_availability_timeout`: The timeout configuration when for when we ask the other peers to discover who has the payload available for the dictated certificates.
+* `handler_certificate_deliver_timeout`: The timeout when waiting on the block synchronizer handler for a certificate fetched from peers to be delivered after submitted to core for processing & validation
```
huitseeker, line 59: I think we need to either point to other documentation for a longer explanation, or describe what the waiting period is in this sentence. Something like "a timeout for checking the validity of a retrieved certificate, including a potential missing parent".

Also, ideally, I suspect we will grow to want an end-to-end timeout per asset (Certificate of block) rather than many intermediate ones.

config/src/lib.rs
```diff
@@ -146,6 +146,11 @@ pub struct BlockSynchronizerParameters {
     /// discover who has the payload available for the dictated certificates.
     #[serde(with = "duration_format")]
     pub payload_availability_timeout: Duration,
+    /// When a certificate is submitted from the handler for further
+    /// processing, we wait for it to become available for us to use. This
+    /// is the timeout while we wait for it.
```
huitseeker, line 151: See the comment on benchmark/README: this is another place to explain what is expected to happen in this wait.

primary/src/block_synchronizer/handler.rs
```diff
@@ -0,0 +1,248 @@
...
+            BlockNotFound { block_id }
+            | Internal { block_id }
+            | BlockDeliveryTimeout { block_id } => block_id,
+        }
+    }
+}
+
+#[automock]
+#[async_trait]
+pub trait Handler<PublicKey: VerifyingKey> {
```
huitseeker, line 55: A succint trait documentation could serve as a summary here.

primary/src/block_synchronizer/handler.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    #[error("Block with id {block_id} couldn't be retrieved, internal error occurred")]
+    Internal { block_id: CertificateDigest },
+
+    #[error("Timed out while waiting for {block_id} to become available after submitting for processing")]
+    BlockDeliveryTimeout { block_id: CertificateDigest },
+}
+
+impl Error {
+    #[allow(dead_code)]
```
huitseeker, line 43: I think you might be able to remove this one.

primary/src/block_synchronizer/handler.rs
huitseeker: I think this is worth opening an issue.

primary/src/block_synchronizer/handler.rs
```diff
@@ -0,0 +1,248 @@
...
+            match result {
+                Ok(some_certificate) => {
+                    if let Some(certificate) = some_certificate {
+                        Ok(certificate)
+                    } else {
+                        Err(Error::BlockNotFound { block_id })
+                    }
+                }
+                Err(_) => Err(Error::Internal { block_id }),
+            }
```
huitseeker, line 139: Isn't that just
```rust
result.map_err(|_| Error::Internal { block_id })?.ok_or_else(|| Error::BlockNotFound { block_id })?
```

huitseeker (commented):
I did a very preliminary pass, and will come back once I've dealt with my case of the foggy brain.

primary/src/block_synchronizer/handler.rs
akichidis: Opened here https://github.com/MystenLabs/narwhal/issues/232

benchmark/README.md
akichidis: Agreed, I also believe that there are now too many timeouts here and there. I thin - and this is me not be Rust-oriented now - it would be great if we could have something like Go has with [context & timeouts](https://golang.cafe/blog/golang-context-with-timeout-example.html) which we could propagate through calls as a way to deal with this. In general I believe we can come up with something which would simplify this.

primary/src/block_synchronizer/handler.rs
akichidis: Yes it is! Nice reduction, thanks

primary/src/block_synchronizer/handler.rs
```diff
@@ -0,0 +1,249 @@
...
+        let mut results: Vec<Result<Certificate<PublicKey>, Error>> = Vec::new();
+
+        // send certificates to core for processing and potential
+        // causal completion
+        let mut wait_for = Vec::new();
+
+        for result in sync_results {
+            match result {
+                Ok(block_header) => {
+                    // we need to perform causal completion since this
```
huitseeker, line 173: Nit: I would place the comment a line below, to make sure the scoping helps the reader figure out what condition guarantees its validity.

huitseeker (approved):
I've left a bunch of comments on minor issues. This looks great, thanks Tasos!

config/src/lib.rs
```diff
@@ -146,6 +146,14 @@ pub struct BlockSynchronizerParameters {
...
     #[serde(with = "duration_format")]
     pub payload_availability_timeout: Duration,
+    /// When a certificate is fetched on the fly from peers, it is submitted
+    /// from the block synchronizer handler for further processing to core
+    /// to validate and ensure parents are available and history is causal
+    /// complete. This property is the timeout while we wait for core to
+    /// perform this processes and the certificate to become available to
+    /// the handler to consume.
+    #[serde(with = "duration_format")]
+    pub handler_certificate_deliver_timeout: Duration,
```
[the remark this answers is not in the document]
akichidis: I've got no problem doing this. My only "objection" is that even to explain the whole thing I had/need to get into details (after all the properties are under the `BlockSynchronizerParameters` which is an implementation detail) so don't believe it makes that much of a difference. Having said that, another +1 on your point about having a unified timeout parameter instead of multiple ones here and there

primary/Cargo.toml
```diff
@@ -31,9 +31,9 @@ tower = "0.4.12"
 crypto = { path = "../crypto" }
 network = { path = "../network" }
 types = { path = "../types" }
-
 store = { git = "https://github.com/mystenlabs/mysten-infra.git", package = "typed-store", rev = "3b7daedf91fd8937dde26e905b8114cac459b866" }
 mysten-network = { git = "https://github.com/mystenlabs/mysten-infra.git", rev = "3b7daedf91fd8937dde26e905b8114cac459b866" }
+mockall = "0.11.0"
```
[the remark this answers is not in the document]
akichidis: My bad, wanted to look into this and totally forgot - thanks for flagging.

So here is the thing, in order for `mockall` to work _only for testing_ , apart from including the dependency only the `dev-dependencies`, I had to mark the use of it on the `handler.rs` file only for testing as well:
```
#[cfg(test)]
use mockall::*;
```
otherwise , as expected, it would try to compile it for prod dev as well and of course the dependency would be missing. Didn't know, but learned now!

primary/src/block_synchronizer/handler.rs
```diff
@@ -0,0 +1,249 @@
...
+    ) -> Self {
+        Self {
+            tx_block_synchronizer,
+            tx_core,
+            certificate_store,
+            certificate_deliver_timeout,
+        }
+    }
+
+    #[instrument(level = "debug", skip_all)]
```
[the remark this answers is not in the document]
akichidis: Ah, in order for this to work the method needs to return a `Result<T, E>` , otherwise the macro won't be able to capture the errors - the method returns a `Vec` of results. I'll log the errors on higher level

primary/src/block_synchronizer/handler.rs
akichidis: Good point!

PR #235, opened 2022-05-13
[chore] Maintenance PR

huitseeker:
This mostly picks up https://github.com/MystenLabs/mysten-infra/pull/74, which logs (to debug) on error in case of storage failures.


akichidis (approved):
Thanks for doing this 💯

PR #240, opened 2022-05-13
Introduce block_synchronizer_handler and wire to block waiter (part 2)

### LINKED ISSUE #183: [get_collections] wire block_synchronizer in block_waiter component

akichidis:
Resolves: MystenLabs/narwhal#183 

This the second part of the MystenLabs/narwhal#183  and follow up of MystenLabs/narwhal#227 .

It extends the block synchronizer handler in order to synchronize missing payloads. Also, the functionality is wired into the `block_waiter` so we make sure that if we are missing the payload (batches) for a certificate we can sync on the fly from other peers and serve it back.

### DIFF
--- a/primary/src/block_synchronizer/handler.rs
+++ b/primary/src/block_synchronizer/handler.rs
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 use crate::{
     block_synchronizer::{
-        handler::Error::{BlockDeliveryTimeout, BlockNotFound, Internal},
-        BlockSynchronizeResult, Command,
+        handler::Error::{BlockDeliveryTimeout, BlockNotFound, Internal, PayloadSyncError},
+        BlockSynchronizeResult, Command, SyncError,
     },
     BlockHeader,
 };
@@ -19,7 +19,7 @@ use tokio::{
     sync::mpsc::{channel, Sender},
     time::timeout,
 };
-use tracing::{debug, error, instrument};
+use tracing::{debug, error, instrument, trace};
 use types::{Certificate, CertificateDigest, PrimaryMessage};
 
 #[cfg(test)]
@@ -38,6 +38,12 @@ pub enum Error {
 
     #[error("Timed out while waiting for {block_id} to become available after submitting for processing")]
     BlockDeliveryTimeout { block_id: CertificateDigest },
+
+    #[error("Payload for block with {block_id} couldn't be synchronized: {error}")]
+    PayloadSyncError {
+        block_id: CertificateDigest,
+        error: SyncError,
+    },
 }
 
 impl Error {
@@ -45,7 +51,8 @@ impl Error {
         match *self {
             BlockNotFound { block_id }
             | Internal { block_id }
-            | BlockDeliveryTimeout { block_id } => block_id,
+            | BlockDeliveryTimeout { block_id }
+            | PayloadSyncError { block_id, .. } => block_id,
         }
     }
 }
@@ -78,6 +85,13 @@ pub trait Handler<PublicKey: VerifyingKey> {
         &self,
         block_ids: Vec<CertificateDigest>,
     ) -> Vec<BlockSynchronizeResult<BlockHeader<PublicKey>>>;
+
+    /// Synchronizes the block payload for the provided certificates via the
+    /// block synchronizer and returns the result back.
+    async fn synchronize_block_payloads(
+        &self,
+        certificates: Vec<Certificate<PublicKey>>,
+    ) -> Vec<Result<Certificate<PublicKey>, Error>>;
 }
 
 /// A helper struct to allow us access the block_synchronizer in a synchronous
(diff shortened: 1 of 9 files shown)


primary/src/block_waiter.rs
```diff
@@ -431,16 +435,47 @@ impl<PublicKey: VerifyingKey, SynchronizerHandler: Handler<PublicKey> + Send + S
...
+            .clone()
+            .into_iter()
+            .filter(|(_, c)| c.is_some())
+            .map(|(_, c)| c.unwrap())
+            .collect();
+
+        let sync_result = self
+            .block_synchronizer_handler
+            .synchronize_block_payloads(found_certificates)
+            .await;
```
akichidis, line 449: Not super optimised strategy, as the whole process will block and wait until we have synchronized all the missing payloads.

Ideally, we could follow here a more stream based approach - so use directly the `block_synchronizer` which emits a message for each successful synchronized batch (best case when we do have then in near cache). However, doing this will require quite some refactoring and made the decision to not spend the time and prioritise delivery. Once we have some stress tests around this with worst case scenarios, we can refactor and improve.

primary/src/block_synchronizer/handler.rs
```diff
@@ -38,14 +38,18 @@ pub enum Error {
 
     #[error("Timed out while waiting for {block_id} to become available after submitting for processing")]
     BlockDeliveryTimeout { block_id: CertificateDigest },
+
+    #[error("Payload for block with {block_id} couldn't be synchronized")]
+    PayloadSyncError { block_id: CertificateDigest },
```
huitseeker, line 43: We could be a lot more precise about diagnosing what went wrong here simply by embedding the inner `SyncError` in the `PayloadSyncError` constructor.

primary/src/block_synchronizer/handler.rs
```diff
@@ -247,4 +258,46 @@ impl<PublicKey: VerifyingKey> Handler<PublicKey> for BlockSynchronizerHandler<Pu
 
         results
     }
+
+    #[instrument(level = "debug", skip_all)]
```
huitseeker, line 262: Do we want to emit an event in case of `err`?

primary/src/block_synchronizer/handler.rs
```diff
@@ -247,4 +258,46 @@ impl<PublicKey: VerifyingKey> Handler<PublicKey> for BlockSynchronizerHandler<Pu
...
+
+        // We want to block and wait until we get all the results back.
+        loop {
+            match rx.recv().await {
+                None => {
+                    debug!("Channel closed when getting results, no more messages to get");
+                    break;
+                }
+                Some(result) => {
+                    println!("Received result {:?}", result.clone());
```
huitseeker, line 288: Is the `println` to stdout the best approach to making this observable?

primary/src/block_synchronizer/handler.rs
```diff
@@ -247,4 +258,46 @@ impl<PublicKey: VerifyingKey> Handler<PublicKey> for BlockSynchronizerHandler<Pu
...
+            .expect("Couldn't send message to block synchronizer");
+
+        // now wait to retrieve all the results
+        let mut results = Vec::new();
+
+        // We want to block and wait until we get all the results back.
+        loop {
+            match rx.recv().await {
+                None => {
+                    debug!("Channel closed when getting results, no more messages to get");
```
huitseeker, line 284: 1. could this be interpreted by someone reading logs as an error case, even when it's the normal termination case for this loop?
2. is the `debug` scope or `trace` more appropriate? (come to think of it, this probably applies to `get_block_headers` above)
3. we're probably going to want to run this task under a timeout. I know we have issues open for this, is there one to which we could tack on taking a look at payload synchronization?

huitseeker (commented):
Looks good! I think we could gain a lot by embedding a `SyncError` in the `PayloadSyncError`.

primary/src/block_synchronizer/handler.rs
akichidis: I've added the `debug` statement purely for the testing/debug purposes - I wouldn't expect someone to try and use this log in prod (or at least I wouldn't like someone have to worry about something going wrong here on production - should make this robust so they won't have to worry about). That being said, to increase robustness probably having a timeout would make sense here. The downstream component - `block_synchronizer` - is already covered from timeout semantics, but it's true that we can't rely on that only.

akichidis: https://github.com/MystenLabs/narwhal/issues/247

primary/src/block_synchronizer/handler.rs
huitseeker: Bump

huitseeker (approved):
Thanks a bunch @akichidis !

primary/src/block_synchronizer/handler.rs
akichidis: I've added `error` events inside the method

PR #244, opened 2022-05-16
Add NewNetworkInfo endpoint

arun-koshy:
This endpoint is used to signal a change in networking info. More info can be found in API [docs](https://docs.google.com/document/d/1ogieB-Y-PkvINJLfpS9G5Tga4RrmBef3OO0RSHa1fyU/edit#).

```
struct ValidatorData {
    public_key: PublicKey,
    stake_weight: u64,
    address: Multiaddr,
}
```

```
// signals a change in networking info
fn new_network_info(epoch_number: u32, validators: Vec<ValidatorData>) -> Result((), Error);
```

This PR just tests serializing and deserializing the network info data that will be required. Once the backend to accept this network info is completed I can complete this work. Also removed the unnecessary comments in the proto file.

### DIFF
--- a/types/proto/narwhal.proto
+++ b/types/proto/narwhal.proto
@@ -96,8 +96,13 @@ message CollectionRetrievalResult {
     }
 }
 
-message Collections {
-    // List of collections to be retreived or removed.
+message GetCollectionsRequest {
+    // List of collections to be retreived.
+    repeated CertificateDigest collection_ids = 1;
+}
+
+message RemoveCollectionsRequest {
+    // List of collections to be removed.
     repeated CertificateDigest collection_ids = 1;
 }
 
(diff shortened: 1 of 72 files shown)


types/proto/narwhal.proto
```diff
@@ -101,15 +63,20 @@ message GetCollectionsRequest {
...
     // List of collections to be removed.
     repeated CertificateDigest collection_ids = 1;
 }
 
-message GetCollectionsResponse {
-    // TODO: Revisit this for spec compliance.  
-    // List of retrieval results of collections.
-    repeated CollectionRetrievalResult result = 1;
+message NewNetworkInfoRequest {
+    int32 epoch_number = 1;
```
akichidis, line 78: We might want `uint32` to stay closer to the api description?

types/proto/narwhal.proto
```diff
@@ -5,90 +5,52 @@ syntax = "proto3";
...
- *
- */
  message BatchMessage {
     BatchDigest id = 1;
     Batch transactions = 2;
 }
 
+message MultiAddr {
+    string address = 1;
+}
```
akichidis, line 42: If the `MultiAddr` in the end is just a `string` then I would recommend to just use directly the `string` type. Unless we want to do something with this struct - e.x write some `validation` method on top of the `MultiAddr` to confirm the validity of it - where I see use of it.

e.x
```
impl MultiAddr {
   fn validate(&self) {
       /// Do some validation etc...
   }
}
```

akichidis (approved):
Thanks @arun-koshy , this looks good to me!

types/proto/narwhal.proto
bmwill: Multiaddr also has a more compact binary format so this could also just be a bytes field

types/proto/narwhal.proto
arun-koshy: Good catch! Thanks

PR #248, opened 2022-05-16
Smoke test node startup

huitseeker:
## Issue
As @akichidis recently discovered, when one tries to run the node with the latest changes like this (internal consensus disabled):
```rust
./node run --committee committee.json --keys keys.json --store db0 primary --consensus-disabled
```

Issue is that the execution on the [main.rs](https://github.com/MystenLabs/narwhal/blob/83e8bf2039ccea681c5abdb6f85d5e769b4c8e8d/node/src/main.rs#L122-L161) node reaches the end as it doesn’t block [here](https://github.com/MystenLabs/narwhal/blob/fdd4f0b45b1d9bf499fb7822d8e1d36144b09fe6/node/src/main.rs#L158) anymore. We don’t use the `rx_transaction_confirmation` anymore and it appears that the node was not shutting down immediately before because of this loop inside the analyze method.

## Solution

The main primary / worker task now return explicit join handles which we `await` on.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -16,6 +16,7 @@ use crypto::{ed25519::Ed25519KeyPair, generate_production_keypair, traits::KeyPa
 use executor::{
     ExecutionIndices, ExecutionState, ExecutionStateError, SerializedTransaction, SubscriberResult,
 };
+use futures::future::join_all;
 use node::{Node, NodeStorage};
 use std::sync::Arc;
 use thiserror::Error;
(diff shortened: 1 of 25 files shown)


node/src/main.rs
```diff
@@ -149,14 +151,17 @@ async fn run(matches: &ArgMatches<'_>) -> Result<()> {
...
                 &store,
                 parameters,
-            );
+            )
         }
         _ => unreachable!(),
-    }
+    };
 
     // Analyze the consensus' output.
```
asonnino, line 159: If I get it correctly, this function returns when we run the dag without consensus, right?

asonnino (approved):

node/src/main.rs
huitseeker: Correct, because its input channel is closed.

akichidis (approved):
Great work Francois, thanks for doing this!

Issue #246, opened 2022-05-16
[network] Eliminate double serialization for the internal NW messages


huitseeker:
https://github.com/MystenLabs/sui/pull/1844

We can do something similar for narwhal....except narwhal uses generic types which wont play nicely with the code gen which really wants concrete types. 

The crypto library needs to be generic, but the rest of the code does not => we should make the config of the validator crypto a compile-time parameter.

PR #253, opened 2022-05-17
Introduce the rounds endpoint for the ProposerAPI

akichidis:
Resolves: MystenLabs/narwhal#253 

Introduces the `rounds` endpoint under the validator api. Performed a small refactoring on some test helper methods of the `consensus` where moved functions under the `test_utils` crate.

### DIFF
--- a/primary/src/grpc_server/proposer.rs
+++ b/primary/src/grpc_server/proposer.rs
@@ -1,0 +1,70 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use config::Committee;
+use consensus::dag::Dag;
+use crypto::traits::VerifyingKey;
+use std::sync::Arc;
+use tonic::{Request, Response, Status};
+use types::{Proposer, RoundsRequest, RoundsResponse};
+
+pub struct NarwhalProposer<PublicKey: VerifyingKey> {
+    /// The dag that holds the available certificates to propose
+    dag: Option<Arc<Dag<PublicKey>>>,
+
+    /// The committee
+    committee: Committee<PublicKey>,
+}
+
+impl<PublicKey: VerifyingKey> NarwhalProposer<PublicKey> {
+    pub fn new(dag: Option<Arc<Dag<PublicKey>>>, committee: Committee<PublicKey>) -> Self {
+        Self { dag, committee }
+    }
+
+    /// Extracts and verifies the public key provided from the RoundsRequest.
+    /// The method will return a result where the OK() will hold the
+    /// parsed public key. The Err() will hold a Status message with the
+    /// specific error description.
+    fn get_public_key(&self, request: RoundsRequest) -> Result<PublicKey, Status> {
+        let proto_key = request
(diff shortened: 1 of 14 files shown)


primary/src/grpc_server/proposer.rs
```diff
@@ -0,0 +1,61 @@
...
+        // convert key
+        let key = self
+            .public_key_mapper
+            .map(
+                request
+                    .into_inner()
+                    .public_key
+                    .ok_or_else(|| Status::invalid_argument("Not public key provided"))?,
+            )
+            .map_err(|_| Status::invalid_argument("Couldn't parse provided key"))?;
```
akichidis, line 45: I am having this practical problem of converting a `PublicKeyProto` to an `Ed25519PublicKey`. Since the `NarwhalProposer` accepts the public key as a generic type argument (`<PublicKey: VerifyingKey>`) , is not possible to just parse at this point the key directly like:
```
let key = Ed25519PublicKey::from_bytes(public_key_proto.bytes.as_ref())
```

when I try to use the key on the dag (e.x `dag.rounds(key)`) it complains - correctly - because the `dag` is accepting the generic type `PublicKey` as type. 

Given the above the only solution I could come up with was introducing a generic mapper trait `PublicKeyMapper` and inject all the way up from the node initialisation the concrete implementation of it, the `Ed25519PublicKeyMapper`. That being said, I don't find it very pretty but I don't have another idea of how this could be achieved alternatively. In the same time I have the feeling that I am missing something obvious - thoughts @huitseeker @arun-koshy ?

huitseeker: 1. Did you notice https://github.com/MystenLabs/narwhal/pull/244/files#diff-8efd1201ab5d436d6cf9a7dbe1a579923c267371eada8742705244a670317400R38-R53 ?
2. This will disappear when I tackle https://github.com/MystenLabs/narwhal/issues/246

akichidis: 1. Yes , but it's basically the same thing - more specifically the `impl TryFrom<&PublicKeyProto> for Ed25519PublicKey` will try to convert the input to a `Ed25519PublicKey`, when the `dag` on the `NarwhalProposer` is passed with the generic type `PublicKey`. So same result.
2. Ah yeah, that would definitely help on this!

huitseeker (commented):
On a general note: do we need a separate server to handle the `rounds` call or can we add it along the list of `get_collections` et al?

I see the logic for having a separate server for the `new_network_info` call (and the `new_epoch` call should probably be served alongside that), but I'm not sure the separation of `rounds` or `node_read_causal` is anything but conceptual.

huitseeker (changes requested):
1. On the endpoints thing: pursuant to our meeting today, I have zero objections if you want to go back to splitting the services for Validator API & Proposer API,
2. On the crypto, I think we should not add too much complexity before the adjustments we may make in #246. I've introduced a PR in #257 which I think we should merge here.

huitseeker (approved):
This looks great, thanks @akichidis !

PR #255, opened 2022-05-17
test: Fix flaky dag test

### LINKED ISSUE #243: [dag] Flaky Dag test
https://github.com/MystenLabs/narwhal/runs/6429300153?check_suite_focus=true is a run from the otherwise innocuous CI pr:
https://github.com/MystenLabs/narwhal/pull/242

Which fails `test_path_compression_from_dag`. Since this is a proptest test, we should have a RNG seed to reproduce.

On first look this seems to be a concurrency issue between an assert checking an invariant and the method enforcing it (as the debug output shows a structure with valid invariants).

Saved logs at : 
https://gist.github.com/huitseeker/53de728095928530ea15ecfbd14d5091

huitseeker:
Fixes #243


dag/proptest-regressions/node_dag.txt
```diff
@@ -0,0 +1 @@
+cc fa3519710f9f7dc108a638115d5d25e9086a6feacc75b01f5576b4a09061cc3b
```
akichidis, line 1: Is this needed? Is this a test output?

akichidis (approved):
LGMT 👍  Thanks @huitseeker for explaining the `prop_assume` functionality offline

dag/proptest-regressions/node_dag.txt
huitseeker: This is an Rng seed. It will make sure that we do not regress on this particular case (e.g. removing the prop_assume) by incentivizing the Rng to execute at least the case that found a bug in #243.

PR #257, opened 2022-05-19
fix: remove cryptographic conversion traits, remove KeyMapper

huitseeker:
This is a simplification for #253 (which branch is the target of this PR).
We can use the generic trait requirement on VerifyingKey to make key conversion work w/o concern for key schemes.


primary/src/grpc_server/validator.rs
```diff
@@ -73,12 +65,11 @@ impl<PublicKey: VerifyingKey, KeyMapper: PublicKeyMapper<PublicKey>>
...
-        let key =
-            self.public_key_mapper
-                .map(request.public_key.ok_or_else(|| {
-                    Status::invalid_argument("Invalid public key: no key provided")
-                })?)
-                .map_err(|_| Status::invalid_argument("Invalid public key: couldn't parse"))?;
+        let proto_key = request
+            .public_key
+            .ok_or_else(|| Status::invalid_argument("Invalid public key: no key provided"))?;
+        let key = PublicKey::from_bytes(proto_key.bytes.as_ref())
```
akichidis, line 71: Ah, completely missed this trait implementation for the `PublicKey` ! This is great...my java instincts worked more in this case and followed the `mapper` path. Thanks @huitseeker for this, much nicer!

akichidis (approved):
Great solution, thanks @huitseeker  for this!

PR #259, opened 2022-05-19
fix: refactor log statements to event ones

akichidis:
A quick fix to change the `tracing::log` statements to `tracing` events as those are properly logged via our tracing capability and it's the practice used from the rest of the codebase. Also added a few more event statements.

### DIFF
--- a/primary/src/block_synchronizer/responses.rs
+++ b/primary/src/block_synchronizer/responses.rs
@@ -5,7 +5,7 @@ use config::Committee;
 use crypto::{traits::VerifyingKey, Digest, Hash};
 use std::fmt::{Display, Formatter};
 use thiserror::Error;
-use tracing::log::{error, warn};
+use tracing::{error, warn};
 use types::{Certificate, CertificateDigest};
 
 // RequestID helps us identify an incoming request and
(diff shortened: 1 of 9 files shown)


primary/src/block_synchronizer/responses.rs
```diff
@@ -109,7 +109,17 @@ impl<PublicKey: VerifyingKey> CertificatesResponse<PublicKey> {
...
+                if let Err(err) = c.verify(committee) {
+                    error!(
+                        "Certificate verification failed for id {} with error {:?}",
+                        c.digest(),
+                        err
+                    );
+                    return true;
+                }
+                false
+            })
```
akichidis, line 122: That should be handled more properly rather than just a log message. Added this issue here: https://github.com/MystenLabs/narwhal/issues/260

huitseeker (approved):
Thanks 
I think you may have missed one instance, see:
```
primary/src/block_synchronizer/tests/block_synchronizer_tests.rs
33:use tracing::log::debug;
```

Ideally, we would also remove the `log` feature in every instance of the `tracing` dependency.

asonnino (approved):
LGTM

PR #262, opened 2022-05-19
Ord for execution indices

asonnino:
* Implement `Ord` for `ExecutionIndices`
* Add unit tests for `ExecutionIndices`

### DIFF
--- a/executor/src/state.rs
+++ b/executor/src/state.rs
@@ -67,12 +67,14 @@ impl Ord for ExecutionIndices {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
         if self.next_certificate_index == other.next_certificate_index {
             if self.next_batch_index == other.next_batch_index {
-                self.next_transaction_index.cmp(&other.next_transaction_index)
+                self.next_transaction_index
+                    .cmp(&other.next_transaction_index)
             } else {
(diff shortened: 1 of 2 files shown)


executor/src/state.rs
```diff
@@ -59,3 +62,25 @@ impl ExecutionIndices {
...
+                self.next_transaction_index
+                    .cmp(&other.next_transaction_index)
+            } else {
+                self.next_batch_index.cmp(&other.next_batch_index)
+            }
+        } else {
+            self.next_certificate_index
+                .cmp(&other.next_certificate_index)
+        }
+    }
```
huitseeker, line 79: You could do that ... or you could write
```rust
        (
            self.next_certificate_index,
            self.next_batch_index,
            self.next_transaction_index,
        )
            .cmp(&(
                other.next_certificate_index,
                other.next_batch_index,
                other.next_transaction_index,
            ))
```
which will [generate. the same thing](https://doc.rust-lang.org/src/core/tuple.rs.html#77-83)

huitseeker (approved):
Fair enough

executor/src/state.rs
asonnino: ha that is much better

Issue #261, opened 2022-05-19
[chore] Update serde_reflection dependency to Zef


huitseeker:
Following new release: https://github.com/zefchain/serde-reflection/commits/main

PR #264, opened 2022-05-20
Update serde-reflection dependency & documentation

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/261


huitseeker (approved):
Thanks! And thanks for cleaning the doc warts!

PR #265, opened 2022-05-20
feat: generate signed certificates in Mocks

huitseeker:
- adds mock_signed_certificate, make_signed_certificates, make_optimal_signed_certificates alongside their unsigned variants,
- they all take a row of private keys which will all sign the cert,
- refactors the mocks to avoid code duplication,
- refactors the `HeaderBuilder::build` method to use the `Signer` impl on keypairs rather than a bespoke closure.

(dozens of tests run through these functions)
This should unblock #258.

### DIFF
--- a/test_utils/src/lib.rs
+++ b/test_utils/src/lib.rs
@@ -15,6 +15,7 @@ use multiaddr::Multiaddr;
 use rand::{rngs::StdRng, Rng, SeedableRng as _};
 use std::{
     collections::{BTreeMap, BTreeSet, VecDeque},
+    ops::RangeInclusive,
     pin::Pin,
 };
 use store::{rocks, Store};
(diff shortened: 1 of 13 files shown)


test_utils/src/lib.rs
```diff
@@ -550,12 +551,47 @@ pub fn make_optimal_certificates(
     make_certificates(start, stop, initial_parents, keys, 0.0)
 }
 
-pub fn make_certificates(
+// Outputs rounds worth of certificates with optimal parents, signed
```
akichidis, line 554: I would suggest to specify on the comment whether the `stop` is inclusive or not

test_utils/src/lib.rs
```diff
@@ -550,12 +551,47 @@ pub fn make_optimal_certificates(
...
+    initial_parents: &BTreeSet<CertificateDigest>,
+    keys: &[Ed25519KeyPair],
+) -> (
+    VecDeque<Certificate<Ed25519PublicKey>>,
+    BTreeSet<CertificateDigest>,
+) {
+    make_signed_certificates(start, stop, initial_parents, keys, 0.0)
+}
+
+// Bernouilli-samples from a set of ancestors passed as a argument,
```
akichidis, line 567: Nit: `Bernouilli` --> `Bernoulli`

akichidis (approved):
Super useful methods, thanks a lot @huitseeker !

test_utils/src/lib.rs
huitseeker: Excellent idea, but instead of doing that I'll just convert those ops to use a [`Range`](https://doc.rust-lang.org/std/ops/struct.Range.html)

PR #267, opened 2022-05-21
[refactor] Batch digests and serialized batch digests

### LINKED ISSUE #156: clarify calculation of BatchDigest
Poking around the code it seems that BatchDigests are created via serializing a [`WorkerMessage::Batch`](https://github.com/MystenLabs/narwhal/blob/main/worker/src/batch_maker.rs#L121-L127) and then hashing it instead of what I would have expected which would be hashing the actual `Batch` type. Is there any context to why it's done the way it is today?

Also it seems non-ideal to craft the byte stream that is hashed by serializing via bincode vs some other canonical format. Is there something i'm missing on that front as well?

huitseeker:
- refactors the digest of `Batch` from hashing the serialized batch message (bincode serialization of `WorkerMessage::Batch`) to hashing transactions one by one,
    * while this is not the ideal approach (we'd like to also hash the tx lengths, rlp-style), this framework offers a pathway towards implementing this in the future,
    * this makes errors that come up in hashing of a structurally invalid serialized batch message explicit, but leaves propagating error handling in the handlers to a further PR,
- defines a function `serialized_batch_digest` on the serialized batch message itself, that computes the same digest,
- introduces randomized test to guarantee this invariant (`serialized_batch_digest` and `Batch::digest` output the same),
- proves the new approach is not a performance regression through a micro-benchmark (see also [results in the next-to-last commit message](https://github.com/MystenLabs/narwhal/pull/267/commits/50d850d7d17b2bb4cf2a8f70b26e6482b498e9e2))

Fixes #156

### DIFF
--- a/types/src/worker.rs
+++ b/types/src/worker.rs
@@ -18,3 +18,6 @@ pub enum WorkerMessage<PublicKey: VerifyingKey> {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ClientBatchRequest(pub Vec<BatchDigest>);
+
+/// Indicates a serialized `WorkerMessage::Batch` message.
+pub type SerializedBatchMessage = Vec<u8>;
--- a/types/src/worker.rs
+++ b/types/src/worker.rs
@@ -24,8 +24,35 @@ pub struct ClientBatchRequest(pub Vec<BatchDigest>);
 /// Indicates a serialized `WorkerMessage::Batch` message.
 pub type SerializedBatchMessage = Vec<u8>;
 
+/// Hashes a serialized batch message without deserializing it into a batch.
+///
+/// See the test `test_batch_and_serialized`, which guarantees that the output of this
+/// function remains the same as the [`Hash::digest`] result you would get from [`Batch`].
+/// See also the micro-benchmark `batch_digest`, which checks the performance of this is
+/// identical to hashing a serialized batch.
+///
+/// TODO: remove the expects in the below, making this return a `Result` and correspondingly
+/// doing error management at the callers.
+/// TODO: update batch hashing to reflect hashing fixed sequences of transactions, see #87.
 pub fn serialized_batch_digest<K: AsRef<[u8]>>(sbm: K) -> BatchDigest {
-    BatchDigest::new(crypto::blake2b_256(|hasher| hasher.update(&sbm)))
+    let sbm = sbm.as_ref();
+    let mut offset = 4; // skip the enum variant selector
+    let num_transactions = u64::from_le_bytes(
+        sbm[offset..offset + 8]
+            .try_into()
+            .expect("Invalid serialized batch"),
+    );
+    offset += 8;
+    let mut transactions = Vec::new();
+    for _i in 0..num_transactions {
+        let (tx_ref, new_offset) =
+            read_one_transaction(sbm, offset).expect("Invalid serialized transaction!");
+        transactions.push(tx_ref);
+        offset = new_offset;
+    }
+    BatchDigest::new(crypto::blake2b_256(|hasher| {
+        transactions.iter().for_each(|tx| hasher.update(tx))
+    }))
 }
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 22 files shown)


akichidis (commented):
Changes look good @huitseeker , thanks a lot for the benchmark tests as well!

Now, regarding the need to dig into the bincode serialised format to extract the transaction info the calculate the hash, I believe this is increasing the complexity and creates a coupling with the bincode serialisation. To better understand the need for this, are we doing it to avoid deserialising the message it self? If that's the case, aren't we already do on the `worker to worker` communication when we receive the message? https://github.com/MystenLabs/narwhal/blob/bdd5e7852dc35f525a3c497b6d4fa0b5d189874e/worker/src/worker.rs#L339 .

huitseeker:
> Now, regarding the need to dig into the bincode serialised format to extract the transaction info the calculate the hash, I believe this is increasing the complexity and creates a coupling with the bincode serialisation.

That is correct. Hence the unit tests to ensure we maintain the invariants that we're meant to.

> To better understand the need for this, are we doing it to avoid deserialising the message it self? If that's the case, aren't we already do on the worker to worker communication when we receive the message?

No, we were already not deserializing the message, as you correctly noted. The goal of this is to give us the ability to update the hashing of the `Batch` according to our hashing specifications:
- to allow us better security, in particular prevention against length confusion attacks, and enforcing TX size limits.
- to allow us to hash intents (see spec),
- to allow us to implement per-transaction logic, including Merkelization of transactions.

The first step towards that is to maintain the ability to have two ways to get to the same digest:
- from a `Batch`,
- from a `SerializedBatch`,
but now, instead of doing it on one binary blob, according to a per-transaction logic. In other words, we will have to maintain an ability to do zero-copy deserialization (which is essentially what we implement here), as we evolve our hashing format. The old hashing method, because it does not allow introspection into the transactions, only satisfied the performance objective without allowing us to change the format.

types/src/worker.rs
```diff
@@ -18,3 +20,55 @@ pub enum WorkerMessage<PublicKey: VerifyingKey> {
...
+
+/// Hashes a serialized batch message without deserializing it into a batch.
+///
+/// See the test `test_batch_and_serialized`, which guarantees that the output of this
+/// function remains the same as the [`Hash::digest`] result you would get from [`Batch`].
+/// See also the micro-benchmark `batch_digest`, which checks the performance of this is
+/// identical to hashing a serialized batch.
+///
+/// TODO: remove the expects in the below, making this return a `Result` and correspondingly
+/// doing error management at the callers.
```
asonnino, line 35: Do we have an issue for this?

asonnino (approved):
Nice

akichidis (approved):
> > Now, regarding the need to dig into the bincode serialised format to extract the transaction info the calculate the hash, I believe this is increasing the complexity and creates a coupling with the bincode serialisation.
> 
> That is correct. Hence the unit tests to ensure we maintain the invariants that we're meant to.
> 
> > To better understand the need for this, are we doing it to avoid deserialising the message it self? If that's the case, aren't we already do on the worker to worker communication when we receive the message?
> 
> No, we were already not deserializing the message, as you correctly noted. The goal of this is to give us the ability to update the hashing of the `Batch` according to our hashing specifications:
> 
> * to allow us better security, in particular prevention against length confusion attacks, and enforcing TX size limits.
> * to allow us to hash intents (see spec),
> * to allow us to implement per-transaction logic, including Merkelization of transactions.
> 
> The first step towards that is to maintain the ability to have two ways to get to the same digest:
> 
> * from a `Batch`,
> * from a `SerializedBatch`,
>   but now, instead of doing it on one binary blob, according to a per-transaction logic. In other words, we will have to maintain an ability to do zero-copy deserialization (which is essentially what we implement here), as we evolve our hashing format. The old hashing method, because it does not allow introspection into the transactions, only satisfied the performance objective without allowing us to change the format.

We talked offline about this as well, all makes sense. Thanks @huitseeker for this!

types/src/worker.rs
huitseeker: https://github.com/MystenLabs/narwhal/issues/268

Issue #212, opened 2022-05-04
[deployment] Co-deploy a set of primaries + workers in a whole network using docker-compose


huitseeker:
We want to use images built in #211  in order to stand up the start of a deployed network.

Some inspiration of ports and config can be found here:
https://github.com/MystenLabs/narwhal/blob/06203d926ec214768ffd83477e3606bb4d4855b0/benchmark/fabfile.py#L13-L41

Some inspiration on node-launching binary commands and options is in :
https://github.com/MystenLabs/narwhal/blob/main/benchmark/benchmark/commands.py

We already use docker-compose for Sui.

Issue #211, opened 2022-05-04
[deployment] Draft an initial Docker image for the Narwhal Primary / Worker


huitseeker:
Inspiration and links:
https://alexbrand.dev/post/how-to-package-rust-applications-into-minimal-docker-containers/

We already use this technique for Sui.

PR #269, opened 2022-05-24
[docker] define Docker container + docker-compose to spin up a cluster

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/211
Resolves: https://github.com/MystenLabs/narwhal/issues/212

This PR is introducing:
* a `Dockerfile` to create a narhwal node (depending on the environment parameters can run either as `primary` or `worker`
* a `docker-compose` file to allow setup a cluster of `4 authorities` with one worker node each (4 primary + 4 worker nodes in total)
* a `README` file under the `Docker` folder to explain the usage

The keys for the nodes, the committee and properties file are pre-configured. A follow up PR will allow us via a simple python script to create more nodes via generating the appropriate `docker-compose.yml` file and json files.


Docker/README.md
```diff
@@ -0,0 +1,150 @@
+### Introduction
```
akichidis, line 1: @huitseeker if you could please do a first pass on this README it would be great. Talked with @Clay-Mysten and he'll try to have a look as well later this week

Docker/README.md
```diff
@@ -0,0 +1,150 @@
...
+to user to treat Narwhal as a pure mempool. When that happens, the gRPC server is bootstrapped
+for the primary nodes and that allow someone to interact with the node.
+
+The docker-compose file is exporting for the `primary` nodes the gRPC server's
+port so it can be accessible from the host machine. For the default setup of `4 primary`
+nodes, the `gRPC` ports are the following:
+* `primary_0`: 8000
+* `primary_1`: 8001
+* `primary_2`: 8002
+* `primary_3`: 8003
```
huitseeker, line 66: We should be clear these are local ports, and be clear that they are the exposition of the same inner port (8000), pointing the reader to look into the docker-compose file for details. 

The reason why this is the case is:
1. someone is bound to ask why the ports are different,
2. you'd want to change this config when running docker-compose to deploy [on say, ecs](https://aws.amazon.com/blogs/containers/deploy-applications-on-amazon-ecs-using-docker-compose/)

Docker/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Folder structure
```
huitseeker, line 68: This is just excellent work, thanks!

Docker/README.md
```diff
@@ -0,0 +1,150 @@
...
+#9 398.4 The following warnings were emitted during compilation:
+#9 398.4 
+#9 398.4 warning: c++: fatal error: Killed signal terminated program cc1plus
+#9 398.4 warning: compilation terminated.
+```
+
+it is possible that the Docker engine is running out of memory and there is no capacity to properly
+compile the code. In this case please try to increase the available RAM at least to 2GB and retry.
+
+#### 2. Mounts denied or cannot start service errors
```
huitseeker, line 143: Nit: Another frequent one is that "Docker recently deprecated the devicemapper storage driver. Depending on your configuration, you may have to [migrate to an overlayfs driver](https://docs.docker.com/storage/storagedriver/overlayfs-driver/)."

Docker/entry.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
huitseeker, line 1: I would add an issue for later to add hadolint + shellcheck to CI to lint this script and the Dockerfile, see [this sui-ops PR](https://github.com/MystenLabs/sui-operations/pull/35).

Docker/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+FROM ubuntu:focal
+
+WORKDIR "$WORKDIR/build"
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends libssl1.1 iputils-ping telnet curl \
+    && rm -rf /var/lib/apt/lists/*
+
+# Copy the Narwhal node binary to bin folder
+COPY --from=builder /build/narwhal/target/debug/node bin/
```
huitseeker, line 47:
```suggestion
COPY --from=builder /build/narwhal/target/$BUILD_MODE/node bin/
```

Docker/README.md
```diff
@@ -0,0 +1,149 @@
...
+you can run the docker-compose command as:
+```
+docker-compose build --build-arg BUILD_MODE=--release
+
+# and then run as
+
+docker-compose up
+```
+
+**Note:** by default each authority's directory will be cleaned up between docker-compose runs when each node
```
huitseeker, line 50: Nit: OK, so: 
1. remove the column in `**Note:**` and look at the Github rendering. 🤯  no?
2. Now, I think we should call this one a `**Warning**` 😁

Docker/docker-compose.yml
```diff
@@ -0,0 +1,147 @@
+version: "3.9"
```
huitseeker, line 1: So I would have gone for the docker-compose replicas feature, which would have allowed you to have a much smaller docker-compose file, using port ranges to expose the ports on which to expose the ports of the validators.

Now I freely admit that I would probably have been an idiot to do that because your exact hardcoding of ports and their exposed variants in fact guarantees we will never have to deal with a port conflict in the next 2-5 weeks, which is the result we want to get to.

Can we nonetheless open an issue to track the need to configure this to more replicas than 4, after we're past our deadline?

huitseeker (commented):
This is ✨ amazing ✨  work, and super duper cool, thanks a lot @akichidis !  I left a few comments, the main ones being around the management of `BUILD_MODE`.

.gitignore
```diff
@@ -1,4 +1,7 @@
 *.json
+!Docker/**/*.json
+Docker/authorities/**/db-*
+Docker/authorities/**/logs*
```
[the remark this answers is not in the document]
akichidis: Ah, I tried to stick to the `Dockefile` convention which is first letter is capital...but no strong opinion on this tbh

Docker/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM lukemathwalker/cargo-chef:latest-rust-1.60.0-slim as builder
+
+# Dictates whether the underlying node should be built as a dev or
+# production binary. If empty, then the development mode will be built.
+# If passed as BUILD_MODE=--release, then the production will be built.
+ARG BUILD_MODE=""
```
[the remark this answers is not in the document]
akichidis: Very good point on the benchmarks. I mainly thought that if someone would like to run a benchmark would basically use the setup that the `benchmark` folder offers. That being said, I can certainly have as default the `--release` option and explain on the `README.md` the option to build on debug mode outlining the above benefits you mentioned.

Docker/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+RUN echo "Will build image with mode: ${BUILD_MODE}"
```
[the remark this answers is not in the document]
akichidis: Actually this prints to the console output, and that was my intention, for someone who is building the image to see the option. Example:
```
....
 => [builder 5/7] COPY . .                                                                                                                                                      0.2s
 => [builder 6/7] RUN echo "Will build image with mode: --release"                                                                                                                       0.2s
 => CANCELED [builder 7/7] RUN cargo build  --bin node 
....
```

Docker/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+    curl \
+    build-essential \
+    libssl-dev \
+    pkg-config \
+    clang \
+    cmake \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install the fmt
+RUN rustup component add rustfmt
```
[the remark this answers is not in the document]
akichidis: Compile fails when trying to build the image in non release (debug) mode. It's asking for `rustfmt`

Docker/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+# Copy all the files in the workdir excluding everything
+# from the .dockerignore
+COPY . .
+
+# Build the binary named "node"
+RUN cargo build $BUILD_MODE --bin node
+
+# Creat another layer so we can re-use caching
+FROM ubuntu:focal
```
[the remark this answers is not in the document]
akichidis: I am all for a lighter linux version as well - on previous positions worked mainly with [Alpine](https://hub.docker.com/_/alpine) linux and found it quite good for (micro)services - and in fact wanted to add a related comment here , but tried to stay in alignment with the approach we took with [SUI](https://github.com/MystenLabs/sui-operations/blob/0618d8d0f52ec455ac6d333ef64d536b6291b556/docker/validator/Dockerfile#L37).

On a more general note, to be honest I would even look into creating an image (e.x mysten-rust) with all the known common setup & tools we need to use across our products and use that for our base on our Docker setup. 

CC @tharbert

Docker/Dockerfile
akichidis: Thanks a lot for doing this - a complete oversight

Docker/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+# Install the fmt
+RUN rustup component add rustfmt
+
+# Set the workdir to the build folder
+WORKDIR "$WORKDIR/build/narwhal"
+
+# Copy all the files in the workdir excluding everything
+# from the .dockerignore
+COPY . .
```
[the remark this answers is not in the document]
akichidis: That's a very good point. Initially I had the `Dockefile` on the project's root directory - the usual approach - but especially with the increasing number of files to help on the setup (e.x `authorities` folder, `entry.sh` file etc) I considered moving this on a separate dedicated folder which obviously leads to those edge cases (e.x building via `docker build` directly rather than using docker-compose). 

So for now I'll do two things
1. `open an issue` as you suggested
2. `include on the README file` instructions of how to build the image without docker-compose, via the command `docker build -f Dockerfile ../` , which will basically allow someone to define the context path on the root folder and use the Dockerfile to build the image

Docker/README.md
```diff
@@ -0,0 +1,150 @@
...
+```
+docker-compose -f docker-compose.yml up
+```
+The first time this will run, it will build the narwhal docker image (this can take a few minutes
+since the narwhal node binary needs to be built from the source code) and then it will spin up 
+a cluster for `4 nodes` by doing the necessary setup for `primary` and `worker` nodes. Each
+`primary` node will be connected to `1 worker` node.
+
+The logs for each authority (primary & worker) can be found on the `logs` folder under the corresponding
+authority folder. The `logs` folder will be created once the node is bootstrapped via docker-compose.
```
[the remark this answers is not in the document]
akichidis: I like the idea of using vocabulary that is understood from the broader audience. My issue is that we have nowhere else any reference to the concept of validator and that might introduce some confusion in the end. I'll dedicate a small paragraph on the README file the `validator` term and how this is connected to what we call authority.

Docker/README.md
akichidis: Good point, I'll do that!

Docker/README.md
akichidis: Thanks a lot, I'll add this as well here!

Docker/docker-compose.yml
akichidis: @huitseeker exactly as you said, that's why avoided to use the `replicas` feature. I am not fond of the long docker-compose file either and I was looking more into the direction of generating the compose yml file via a very simple python script accepting number of authorities & workers which would:
1.  generate the `authority keys` and write the necessary files
2. create the `committee.json file`
3. generate the `docker-compose` file

Alternatively we can have the script do 1 & 2 and configure things to avoid port conflict and make the docker-compose very small using the `replicas` future, but to be honest I like the simplicity and predicability of fixed common ports

Docker/entry.sh
akichidis: Issue opened here https://github.com/MystenLabs/narwhal/issues/277

Docker/docker-compose.yml
akichidis: issue here https://github.com/MystenLabs/narwhal/issues/278

Docker/Dockerfile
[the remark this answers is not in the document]
huitseeker: I think on this particular PR, I'd be OK simply going from ubuntu to the base image of rust-latest, and transitively cargo-chef, which I understand to be debian:bullseye. Less moving parts.
https://github.com/docker-library/official-images/blob/master/library/rust

Docker/README.md
```diff
@@ -0,0 +1,187 @@
...
+smaller binary (and image) size, you can run the docker-compose command as:
+```
+$ docker-compose build --build-arg BUILD_MODE=debug
+
+# and then run as
+
+$ docker-compose up
+```
+
+**Warning:** by default each validator's directory will be cleaned up between docker-compose runs when each node
```
huitseeker, line 65: Remove the column, you'll get super fancy Github rendering
```suggestion
**Warning**: by default each validator's directory will be cleaned up between docker-compose runs when each node
```

Docker/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+If you try to spin up the nodes via docker-compose and you come across errors such as `mounts dened`
+or `cannot start service`, please make sure that you allow Docker to share your host's [Docker/validators](validators)
+folder with the containers. If you are using Docker Desktop you can find more information of how to do
+that here: [mac](https://docs.docker.com/desktop/mac/#file-sharing), [linux](https://docs.docker.com/desktop/linux/#file-sharing),
+[windows](https://docs.docker.com/desktop/windows/#file-sharing)
+
+#### 3. Deprecated devicemapper storage driver
+
+If you see errors related to the `Deprecation of the devicemapper storage`, then you might need to
```
huitseeker, line 185: FYI : those present as mount errors. I'd just be vague and point to Docker doc. This is fundamentally an upstream issue, I'm just mentioning it because the deprecation is recent and the forums are ablaze with it.

Docker/Dockerfile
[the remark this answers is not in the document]
huitseeker: Oh, your docker is more fancy than mine it seems.

huitseeker (approved):
Thanks a bunch Tasos!

Docker/docker-compose.yml
```diff
@@ -106,6 +110,8 @@ services:
       - "4000"
       - "4001"
       - "4002"
+    ports:
+      - "7003:4001" # expose the port for the worker to accept transactions from clients
```
huitseeker, line 114: 1. Good catch
2. in general, annotations port number -> purpose of port would be great overall.

huitseeker (approved):
:shipit:

PR #274, opened 2022-05-24
fix error when requesting non found certificate via block waiter

akichidis:
This is a fix of a bug discovered while interacting with a running Narwhal node cluster. To replicate the error should follow the steps:
1. Spin up a Narwhal node cluster
2. Request via the `GetCollections` request a **single** certificate that doesn't exist
3. Primary node should panic with the error `mpsc bounded channel requires buffer > 0` which is raised on the [handler](https://github.com/MystenLabs/narwhal/blob/5cfb317c2a0dfe29190d1ba67281f8e99dc57799/primary/src/block_synchronizer/handler.rs#L270)

After some investigation it appears that when requesting a single certificate via `GetCollections` that doesn't exist (can't even fetch from other peers) , it was ending up passing an empty vector of [found_certificates](https://github.com/MystenLabs/narwhal/blob/5cfb317c2a0dfe29190d1ba67281f8e99dc57799/primary/src/block_waiter.rs#L448) on the block_waiter when fetching the results. The tokio `channel` used on those underlying handler methods are bounded and their size is initialised based on the provided vector of requested certificates - which when `0` it panics.

As a fix I've amended the `handler` methods to immediately return and respond with an empty vector when an empty vector is passed.

### DIFF
--- a/primary/src/block_synchronizer/handler.rs
+++ b/primary/src/block_synchronizer/handler.rs
@@ -175,6 +175,11 @@ impl<PublicKey: VerifyingKey> Handler<PublicKey> for BlockSynchronizerHandler<Pu
         &self,
         block_ids: Vec<CertificateDigest>,
     ) -> Vec<Result<Certificate<PublicKey>, Error>> {
+        if block_ids.is_empty() {
+            trace!("No blocks were provided, will now return an empty list");
+            return vec![];
+        }
+
         let sync_results = self.get_block_headers(block_ids).await;
         let mut results: Vec<Result<Certificate<PublicKey>, Error>> = Vec::new();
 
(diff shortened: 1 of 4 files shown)


huitseeker:
> The tokio channel used on those underlying handler methods are bounded and their size is initialised based on the provided vector of requested certificates - which when 0 it panics.

This seems to be the issue: is there a reason why we would not initialize to the DEFAULT_CHANNEL_SIZE`?

akichidis:
> > The tokio channel used on those underlying handler methods are bounded and their size is initialised based on the provided vector of requested certificates - which when 0 it panics.
> 
> This seems to be the issue: is there a reason why we would not initialize to the DEFAULT_CHANNEL_SIZE`? Answering my own question: We're using them as a queue aren't we?

As I don't expect to receive more messages than the `certificates.len()` one I considered that there is no reason to have a different bound size. What do you mean using _them_ as a queue? We could use alternatively the `DEFAULT_CHANNEL_SIZE` yes. On another note, I consider essential to have that check there though for non empty vector as we don't want to defer to the `block_synchronizer` to ask to synchronize an empty array anyways.

huitseeker (approved):
Thanks for the fix and the context!

PR #275, opened 2022-05-25
fix: adjust BUILD_MODE default & passing

huitseeker:
BUILD_MODE fix for #269


akichidis (approved):
Thanks a lot @huitseeker for doing this!

PR #285, opened 2022-05-26
[refactor] trace only necessary fields

akichidis:
This PR ensures that we don't track excessive information on our tracing which when enabled (on log level DEBUG) is causing the CPU maxing out and memory excessively consumed (in ~1 min more than 1 GB is consumed) as the whole DAG structure was always captured. We recently observed that behaviour during the docker cluster deployment.

### Before:
* Running a cluster of 4 authorities with log level DEBUG and consensus disabled (so the virtual DAG is used. After 1 minute of running the `primary_0` node had the following stats:
Screenshot 2022-05-26 at 13 07 38

### After the changes on similar setup and running time of 5 minutes:

Screenshot 2022-05-26 at 14 48 08

### DIFF
--- a/consensus/src/dag.rs
+++ b/consensus/src/dag.rs
@@ -146,7 +146,7 @@ impl<PublicKey: VerifyingKey> InnerDag<PublicKey> {
         }
     }
 
-    #[instrument(level = "debug", err)]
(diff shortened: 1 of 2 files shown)


huitseeker (approved):
Thanks!

Issue #281, opened 2022-05-25
[genesis] requesting (via get_collections) or removing (via remove_collections) genesis certificates will fail


akichidis:
### Description
Following @huitseeker fix on the issue https://github.com/MystenLabs/narwhal/issues/270 , we are now able to store and read genesis certificates on the DAG, however, if we try to 
1. request a genesis certificate via the `get_collections` request
2. or delete a genesis certificate via the `remove_collections` request

both those operations will fail. All the operations performed are dealing only with certificates that exist on the `certificate_store` - it is reminded that genesis certificates are generated on the fly and kept in memory only.

Even if the genesis block don't really contain particular value for the consensus (in its current form) , the above operations shouldn't fail and in worst case should just be no-ops, meaning:
* on the `get_collections` that response could be an empty payload (what actually is)
* on the `remove_collections` no really delete anything - apart from the DAG entries

Issue #270, opened 2022-05-24
[dag] fail to insert processed certificates in DAG


akichidis:
### Description
When tried to run a small cluster of `4 primary` and `4 worker` nodes, with consensus disabled, I observed on the log files that the insert operations on the DAG were failing. More specifically receiving errors like the one bellow:
```
2022-05-24T14:30:44.464149Z DEBUG primary::core: Processing uPR8oR9AvWWXSHKr: C1(noDjBFfXGqQioHTf6jEIPYthhUWCMsC12ZJ9DMh7Ujk=, I7eRt6/Y5Z0IgQ6L)
2022-05-24T14:30:44.465776Z ERROR insert{self=InnerDag { rx_primary: Receiver { chan: Rx { inner: Chan { tx: Tx { block_tail: 0xaaab101d5010, tail_position: 2 }, semaphore: (Semaphore { permits: 999 }, 1000), rx_waker: AtomicWaker, tx_count: 1, rx_fields: "..." } } }, rx_commands: Receiver { chan: Rx { inner: Chan { tx: Tx { block_tail: 0xaaab101dac70, tail_position: 0 }, semaphore: (Semaphore { permits: 1000 }, 1000), rx_waker: AtomicWaker, tx_count: 1, rx_fields: "..." } } }, dag: NodeDag { node_table: {} }, vertices: RwLock { data: {}, poisoned: false, .. } } certificate=1p1Dtk/cWV5BEYIk: C1(Zy82aSpF8QghKE4wWvyIoTWyLetCuUSfk2gxHEtwdbg=, FTno2BRK45kvbLJn)}: consensus::dag: error=Dag invariant violation No vertex known by these digests: [UXWPnsSff4MWl6KDu4yCsSx/gDW3GSGxWQguXZYs5gk=, Zp+McZcRLR6cXMYoDmLHj3VYNb+qeBJb9pWI/qo+fJ4=, a11qQmG81vCtFbwYkKPsh9k6MTFdau840cfhVSqBEo0=, ch8wEJmFilfbYjl/h3EJ6h3qI8+Gf1UPwSQC+7viC0M=]
```

The above issue observed on all the primary nodes, so it seems to be a common behaviour.

By a quick investigation I believe that this is attributed to the fact that on an empty DAG, when we try to insert the very first certificates, its parents don't exist - meaning the genesis headers - and the insertion fails.

### Steps to reproduce
[...]

PR #287, opened 2022-05-27
refactor: insert genesis certificates in certificate store when using virtual DAG

akichidis:
Related to: https://github.com/MystenLabs/narwhal/issues/281

Both the:
* get_collections
* remove_collections

requests are using as source of truth the `certificate_store`. Following the fix on https://github.com/MystenLabs/narwhal/issues/270, now it would be possible via the `read_causal` request to get back the genesis certificate digests which when queried via the above requests, will lead to failures.

To avoid that, we want to store the genesis certificates on the storage as well - when using the virtual DAG (when using Tusk we won't do that) - so those can be discoverable from the above requests.

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 use config::{Parameters, SharedCommittee, WorkerId};
 use consensus::{dag::Dag, Consensus, SubscriberHandler};
-use crypto::traits::{KeyPair, Signer, VerifyingKey};
+use crypto::{
+    traits::{KeyPair, Signer, VerifyingKey},
+    Hash,
+};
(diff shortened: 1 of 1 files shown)


akichidis:
cc @arun-koshy the issue described above might make your mock client fail if you try to process a genesis block. This PR will fix that.

huitseeker (commented):
Actually, on this one I would just make genesis and empty blocks compressible, following @akichidis 's brillant idea of implementing `Affiliated` for these blocks.
Compressible blocks will never be reported by a `read_causal`, which alleviates the sync problem.

asonnino (approved):

huitseeker:
@akichidis this should be obsolete given #291

akichidis:
Closing this PR as https://github.com/MystenLabs/narwhal/pull/290 is now the more appropriate solution

PR #288, opened 2022-05-27
Mock Demo Client

arun-koshy:
Does not fully work because the workers are crashing on me, but just wanted to get some thoughts on the scenario that I am trying to achieve and if I got anything wrong there.


examples/src/demo_client.rs
```diff
@@ -0,0 +1,154 @@
...
+
+        println!("NodeReadCausalRequest={:?}\n", request);
+
+        let response = client.node_read_causal(request).await;
+
+        println!("NodeReadCausalResponse={:?}\n", response);
+
+        let node_read_causal_response = response.unwrap().into_inner();
+
+        if collection_ids.len() + node_read_causal_response.collection_ids.len() <= gas_limit {
```
akichidis, line 91: I know this is for demonstration, but I would like to outline that consecutive read_causal requests, for example:
* `read_causal(round=1)`
* `read_causal(round=2)` 
* `read_causal(round=3)`
....

it's possible to contain overlaps on their responses, meaning `collection_ids` that exist on multiple responses. Since we are doing a BFS and a certificate we start from might refer to a parent certificate of the previous round of ours, we always have the chance of  referring to same collection ids.

That being said, we might want to show that we are doing some deduping and we only take into account the "unique" collection ids.

CC @huitseeker

akichidis: Also, in practice, wouldn't the consensus layer want to fetch the actual transactions to figure out the gas cost? I mean, I would imagine in high level that they want to do something like (pseudo code):
````
round  = rounds().oldest_round;
newest_round = rounds().newest_round;

while round <= newest_round 
{
    let collection_ids = read_causal(round);

    // keep non overlapping collection ids
    collection_ids = dedupe_collection_ids(collection_ids, all_collection_ids);

    // fetch the actual transactions so we can
    // calculate the gas will need before proposing

    let transactions = get_collections(collection_ids);

    // calculate the gas for each transaction and return
    // the ones that should be included so we stay under
    // the limit - or whatever logic consensus wants to implement.
   
    let (collection_ids_to_include, gas) = collections_under_gas_limit(transactions, total_gas, gas_limit);
   
     total_gas += gas;
     all_collection_ids.extend(collection_ids_to_include);

     if collection_ids_to_include.len() != collection_ids.len() {
         // not all collections included, meaning we reached (or we would reach if included more)
         // the gas limit. Now abort.
         break;
     }
     
     round += 1;
}
````

**Review by akichidis — COMMENTED** [review]
Left a few comments, but will iterate again. Great work @arun-koshy , let's see what are the scenarios we want to implement here

File: examples/doc/docs.md [hunk 2]
```diff
@@ -0,0 +1,513 @@
+# Protocol Documentation
```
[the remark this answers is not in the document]
arun-koshy: Yeah using [protoc-gen-doc](https://github.com/pseudomuto/protoc-gen-doc)

examples/src/demo_client.rs
huitseeker: @arun-koshy @akichidis 
- it's expected for `read_causal(round=n+1)` to indeed include the `read_causal(round=n)`,
- we should indeed show deduplication,
- as far as gas cost is concerned, the fancy calculations should only exist in comments, saying "PLACE YOUR EXECUTION AND GAS ESTIMATION HERE"
- I suggest that to estimate gas we simply emulate it by saying each transaction costs one unit of gas. 
- this does force us to issue a `get_collection` and compute on its results, because the info of the $ of transactions is hidden in the batches and  not accessible from a header.

examples/src/demo_client.rs
```diff
@@ -0,0 +1,154 @@
...
+    println!("GetCollectionsRequest={:?}\n", request);
+
+    let response = client.get_collections(request).await;
+
+    println!("GetCollectionsResponse={:?}\n", response);
+
+    let get_collection_response = response.unwrap().into_inner();
+
+    // TODO: This doesn't work yet, figure out why workers are crashing.
+    println!("Found {} batches", get_collection_response.result.len());
```
huitseeker, line 143: This still does not work, but not because of workers crashing:
https://gist.github.com/huitseeker/8ced40e2e8ebf14e5d1aa73b3fd2bb0f

examples/README.md
```diff
@@ -0,0 +1,31 @@
+### How to run demo client
+
+1. First start up the narwhal cluster via `docker-compose`
```
akichidis, line 3: I would recommend to also point the reader to the full [README](https://github.com/MystenLabs/narwhal/tree/main/Docker) of the docker setup

akichidis, line 3: I would recommend to also point the reader to the full [README](https://github.com/MystenLabs/narwhal/tree/main/Docker) of the docker setup

examples/src/demo_client.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    let rounds_response = response.unwrap().into_inner();
+    let oldest_round = rounds_response.oldest_round;
+    let newest_round = rounds_response.newest_round;
+    let round = oldest_round + 1;
+    println!("\n2) Find collections from earliest round and continue to add collections until gas limit is hit\n");
+    println!("\n---- Use NodeReadCausal endpoint ----\n");
+
+    let mut collection_ids: Vec<CertificateDigest> = vec![];
+    while round < newest_round && collection_ids.len() < gas_limit {
```
akichidis, line 75: Also, it seems that the `round` is not getting increment in the loop

huitseeker:
@arun-koshy We probably want to pick up at #324

PR #289, opened 2022-05-27
fix: block synchronizer should not sync payload for own headers

akichidis:
A primary node does not store in its local `payload_store` the batch digests for the headers that they create on their own (meaning with batches their own workers create). See https://github.com/MystenLabs/narwhal/blob/f5dd2c29ffc131d3dbc5e9b2fca22464baae083a/primary/src/synchronizer.rs#L54

When using `block_synchronizer` to synchronize the block payloads, the above is not respected and it is erroneously looking into the `payload_store` leading to a false negative - meaning that it will try to fetch the payload from the other peers. This request though will fail as we'll try to sync the "missing" batches via our own worker which already has the batches, hence the primary will never be notified for "successful synchronisation" (worker just ignores the request in this case) and the original payload sync request on the `block_synchronizer` will timeout.

This PR is fixing this issue by adding an extra check on the block synchronizer for their own certificates. Also, some extra debug statements have been added (helped to debug the issue and found them useful) + some extra useful tracing.

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -2,10 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 use config::{Parameters, SharedCommittee, WorkerId};
 use consensus::{dag::Dag, Consensus, SubscriberHandler};
-use crypto::{
-    traits::{KeyPair, Signer, VerifyingKey},
-    Hash,
-};
+use crypto::traits::{KeyPair, Signer, VerifyingKey};
 use executor::{ExecutionState, Executor, SerializedTransaction, SubscriberResult};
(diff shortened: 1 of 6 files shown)


huitseeker (approved):
IIUC, this is the extension of #163 to payloads. Thanks a lot!

akichidis:
> IIUC, this is the extension of #163 to payloads. Thanks a lot!

Yes, it could definitely be considered an extension of #163 ! To be honest, if we would like to simplify - and don't consider a huge impact in our performance (need to measure) - we could always store the batch ids on the `payload_store` , so things are uniform. That will save us such checks here and there.

PR #272, opened 2022-05-24
feat: feed the genesis in the Dag by default

huitseeker:
THis aligns the behavior of our Dag with that of the store

- adds the genesis by default to the Dag as part of the constructor,
- unit tests that initial provisioning,
- adapts the behavior of tests to not feed genesis manually.

Fixes #270

PR #290, opened 2022-05-27
feat: insert no-payload certificates as compressible

### LINKED ISSUE #281: [genesis] requesting (via get_collections) or removing (via remove_collections) genesis certificates will fail

huitseeker:
This adds an implementation of `Affiliated`, providing the insertion-time value of DAG vertexes' compressible field, which returns true on emtpy blocks.
Consequences:
- the DAG has genesis, but it's never reported in a read_causal,
- empty blocks get automatically purged by the next read_causal.

This can be seen as an alternative to #287, a continuation of #272.

This partially fixes #281 in that the user will no longer have a reason to do `get_collections` or `remove_collections` on a genesis certificate (but that important detail needs to be better documented)


consensus/src/tests/dag_tests.rs
```diff
@@ -56,8 +56,89 @@ async fn test_dag_new_has_genesis() {
...
+    }
+
+    // genesis is no longer here
     for certificate in genesis {
-        assert!(dag.contains(certificate).await)
+        assert!(!dag.contains(certificate).await);
+    }
+}
+
+// The above test relies on the fact that genesis produces empty blocks: we re-run it with non-genesis empty blocks to
```
akichidis, line 90: Nit: `The above test...`  ...I would suggest to refer to the name directly `test_dag_new_has_genesis_and_its_not_live` , it's always possible someone inserting a new test between those two and then introduce some confusion

akichidis (approved):
Excellent work @huitseeker , and thanks for the extra testing! 

I would recommend to look into the integration tests [here as well ](https://github.com/MystenLabs/narwhal/blob/f5dd2c29ffc131d3dbc5e9b2fca22464baae083a/primary/tests/integration_tests_validator_api.rs#L419)where we seem to manually populate the genesis certificates to the storage. Not writing in storage now shouldn't make them fail.

Also, definitely not for this PR, but we should probably catch up of how we are going to remove from the storage as well (header_store, certificate_store) the "removed" compressible certificates once they are compressed on the DAG. As they won't appear on the `get_collections` they won't be part of any `remove_collections` process.

huitseeker:
> I would recommend to look into the integration tests [here as well ](https://github.com/MystenLabs/narwhal/blob/f5dd2c29ffc131d3dbc5e9b2fca22464baae083a/primary/tests/integration_tests_validator_api.rs#L419)where we seem to manually populate the genesis certificates to the storage. Not writing in storage now shouldn't make them fail.

Yep, all removed in a second commit, and everything passes.

> Also, definitely not for this PR, but we should probably catch up of how we are going to remove from the storage as well (header_store, certificate_store) the "removed" compressible certificates once they are compressed on the DAG. As they won't appear on the get_collections they won't be part of any remove_collections process.

Correct! Opened #291

PR #295, opened 2022-05-27
feat: Use caching through the workspace-hack

huitseeker:
This is a port of the caching technique in https://github.com/MystenLabs/sui/pull/2246

We use the benefit of the workspace-hack crate in order to ensure we can gain the benefit of docker layer caching of dependencies. The second (and later) build only rebuild the source code in the repo.


Docker/Dockerfile
```diff
@@ -1,12 +1,14 @@
...
 # production binary. If empty, then the development mode will be built.
 # If passed as BUILD_MODE=--release, then the production will be built.
-ARG BUILD_MODE
+ARG BUILD_MODE="release"
 
-RUN echo "Will build image with mode: ${BUILD_MODE}"
+#################################################################
+# Stage 0
+#################################################################
+FROM lukemathwalker/cargo-chef:latest-rust-1.60.0-slim as chef
```
bmwill, line 9: Probably worth using the rust container directly instead of the cargo-chef one since we don't use cargo-chef

Docker/Dockerfile
```diff
@@ -21,12 +23,50 @@ RUN apt-get update \
...
+# In a fresh layer, copy in the "plan" generated by the planner
+# and run `cargo build` in order to create a caching Docker layer
+# with all dependencies built.
+FROM chef AS builder 
+ARG BUILD_MODE
+
+COPY --from=planner /narwhal/Cargo.toml Cargo.toml
+COPY --from=planner /narwhal/Cargo.lock Cargo.lock
+COPY --from=planner /narwhal/workspace-hack workspace-hack
+
```
bmwill, line 63: You're missing a step here to actually build and cache the dependencies.

bmwill (changes requested):

Docker/Dockerfile
bmwill: eg:
```
RUN cargo build --${BUILD_MODE}
```

huitseeker: Er, yes, of course, it was supposed to be part of this commit.

bmwill (approved):

akichidis (approved):
Can't wait for faster (re)builds! Thanks @huitseeker 💯

PR #296, opened 2022-05-27
ci: run udeps only on a nightly basis

huitseeker:
This is a more sedate port of https://github.com/MystenLabs/sui/pull/2231/commits/b820f2e1ed50adcf9c93f5a282cf76cb58cbc6d1
See there for rationale, which applies here.

The sedate part is that instead of obliterating this long-running job we still run it once a day.


bmwill:
@huitseeker heh you caught me eh? I saw that once the workspace hack was added for some reason (even when turning it off) udeps started taking 45+ minutes :(

bmwill:
@huitseeker i'll do the same for sui

https://github.com/MystenLabs/sui/pull/2287

bmwill (approved):

PR #324, opened 2022-05-30
Demo client

huitseeker:
A more up-to-date and rebased version of #288 with a few fixes


asonnino (approved):
seed.py and the demo client look good to me, but I am not familiar with docker enough to review the docker file

PR #336, opened 2022-06-01
fix: Expose the failibility of `serialize_batch_digest` & error-handle

### LINKED ISSUE #268: [serialization, hashing] Error management in batch / serializedBatch hashing
#267 introduces a digest methodology for `Batch` and `SerializedBatchMessage` that proceeds per transaction, and correspondingly exposes errors until now hidden in the implicit `bincode` deserialization errors we would obtain when we'd try to deserialize the batch.

Correspondingly, the errors in digesting a serialized message should be propagated to the callers of `serialized_batch_message`.

huitseeker:
- responds to an invalid serialized batch,
    * the primary-to-worker flow is straightforward, as an error should not really happen,
    * the processor case is more complex, this is a place where we could use some alerting à la MystenLabs/sui#5295 
- closes MystenLabs/narwhal#268

### DIFF
--- a/executor/src/batch_loader.rs
+++ b/executor/src/batch_loader.rs
@@ -16,7 +16,7 @@ use tokio::{
     sync::mpsc::{channel, Receiver, Sender},
     task::JoinHandle,
 };
-use tracing::warn;
+use tracing::{error, warn};
 use types::{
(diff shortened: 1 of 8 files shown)


akichidis (approved):
Thanks a lot @huitseeker !

PR #334, opened 2022-05-31
example formatting for requests & responses

akichidis:

PR #337, opened 2022-06-01
Final Demo Changes

arun-koshy:
- Completed part 2 in @huitseeker [request](https://mysten-labs.slack.com/archives/C03GL3M8P39/p1654001174255939). 
- Made some changes to seeder for whenever we get docker working.

Attached demo logs for review - [demo_client.log](https://github.com/MystenLabs/narwhal/files/8810129/demo_client.log)

### DIFF
--- a/benchmark/benchmark/seed.py
+++ b/benchmark/benchmark/seed.py
@@ -3,8 +3,6 @@ import subprocess
 from math import ceil
 from os.path import basename, splitext
 from time import sleep
-from pathlib import Path
-import json
 
 from benchmark.commands import CommandMaker
 from benchmark.config import BenchParameters, ConfigError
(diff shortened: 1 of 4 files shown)


benchmark/benchmark/seed.py
```diff
@@ -34,24 +32,22 @@ def _kill_nodes(self):
         except subprocess.SubprocessError as e:
             raise BenchError('Failed to kill testbed', e)
 
-    def run(self, committee_path):
-        assert isinstance(committee_path, Path)
+    def run(self, starting_data_port):
```
huitseeker, line 35: We probably want to sanitize `starting_data_port` as an `int`, since this will allow us to detect passing the wrong argument without a long-running command launching.

huitseeker (approved):
This looks amazing 🤩 ! Thanks @arun-koshy (and @akichidis that's also part of this PR through #334)!

PR #340, opened 2022-06-01
[test] Extend dag removal test

huitseeker:
Fixes bug in test which was adding additional certificates at the wrong round. 
Cosmetic changes to build `NodeRef`s less manually (i.e. using `NodeRef::from_pointee` rather than `Arc::new`).

This gives us more confidence DAG compression is working as intended (see assert circa "`// certificates are gone`")


asonnino (approved):
Good catch

PR #342, opened 2022-06-02
[dag] fix: fix a Read Your Writes violation in `remove_collections`,  `rounds`

huitseeker:
Our garbage collection is a mark-and-sweep algorithm, where the mark part is in `make_compressible` (called by `remove_collections`) and `read_causal` triggers a sweep.

To make sure we don't return `rounds` as live when they wouldn't be seen as such by any subsequent `read_causal`, we need to trigger the sweep in the first place.

This takes the simple and unoptimized solution of starting a `read_causal` before a call to `rounds`, which is correct but costly.
We outline a better solution as a TODO.


akichidis (approved):
Awesome, thanks a lot @huitseeker !

PR #351, opened 2022-06-08
Add skeleton for NewEpoch endpoint

arun-koshy:
This endpoint is used to signal a new epoch. More info can be found in API [docs](https://docs.google.com/document/d/1ogieB-Y-PkvINJLfpS9G5Tga4RrmBef3OO0RSHa1fyU/edit#).

```
struct ValidatorData {
    public_key: PublicKey,
    stake_weight: u64,
    address: Multiaddr,
}
```

```
/// signals a new epoch
fn new_epoch(epoch_number: u32, validators: Vec<ValidatorData>) -> Result((), Error);
```

This PR just tests serializing and deserializing the new epoch data that will be required. Once the backend to handle an epoch change is completed I will complete this work.

### DIFF
--- a/primary/src/grpc_server/configuration.rs
+++ b/primary/src/grpc_server/configuration.rs
@@ -3,7 +3,7 @@
 use crypto::{ed25519::Ed25519PublicKey, traits::ToFromBytes};
 use multiaddr::Multiaddr;
 use tonic::{Request, Response, Status};
-use types::{Configuration, Empty, NewNetworkInfoRequest};
+use types::{Configuration, Empty, NewEpochRequest, NewNetworkInfoRequest};
 
 #[derive(Debug)]
 pub struct NarwhalConfiguration {}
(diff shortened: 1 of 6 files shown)


primary/src/grpc_server/configuration.rs
```diff
@@ -16,6 +16,45 @@ impl NarwhalConfiguration {
...
+        let mut parsed_input = vec![];
+        for validator in validators.iter() {
+            let proto_key = validator
+                .public_key
+                .as_ref()
+                .ok_or_else(|| Status::invalid_argument("Missing public key"))?;
+            let public_key: Ed25519PublicKey =
+                Ed25519PublicKey::from_bytes(proto_key.bytes.as_ref()).map_err(|err| {
+                    Status::invalid_argument(format!("Could not serialize: {:?}", err))
+                })?;
```
akichidis, line 35: maybe you could introduce a method like this https://github.com/MystenLabs/narwhal/blob/main/primary/src/grpc_server/proposer.rs#L30 so you can reuse it on the `new_network_info` as well?

akichidis (approved):
LGTM, thanks @arun-koshy !

PR #376, opened 2022-06-14
Laura/eliminate recursion

lanvidr:
https://github.com/MystenLabs/sui/issues/5259

Removes the recursion and instead uses a loop in order to compress the DAG.

### DIFF
--- a/dag/src/lib.rs
+++ b/dag/src/lib.rs
@@ -245,47 +245,30 @@ impl<T: Sync + Send + std::fmt::Debug> Node<T> {
         }
 
         // create set of initial compressible and incompressible nodes
-        let (compressible, incompressible): (Vec<NodeRef<T>>, Vec<NodeRef<T>>) = self
+        let (mut compressible, mut incompressible): (Vec<NodeRef<T>>, Vec<NodeRef<T>>) = self
             .raw_parents_snapshot()
             .into_iter()
             .partition(|p| p.is_compressible());
 
-        // recurse to get compressed set
-        let compressed = Self::parents_inner(compressible, incompressible);
-
-        // deduplicate compressed set and save
-        let res: Vec<NodeRef<T>> = compressed
-            .into_iter()
-            .unique_by(|arc| Arc::as_ptr(arc))
-            .collect();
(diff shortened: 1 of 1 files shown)


dag/src/lib.rs
```diff
@@ -244,35 +244,50 @@ impl<T: Sync + Send + std::fmt::Debug> Node<T> {
...
+        // base case: no more compressible nodes
+        loop {
+            if compressible.is_empty() {
+                return incompressible;
+            }
+
+            // for all the compressible nodes, replace it with its parents
+            let new_parents: Vec<NodeRef<T>> = compressible
+                .par_iter()
+                .flat_map_iter(|node| node.parents())
```
akichidis, line 280: I feel that I might be missing something here, but it seems that we are still making a recursion call here. Also, the code seems quite the same as before in the sense that I don't see how we are leveraging iteration here, as this loop should exit in the first iteration - I would expect the `new_parents` to be consisted only of `incompressible` elements

akichidis: Also tried [this test](https://github.com/MystenLabs/narwhal/blob/1a06209f5c59a9b1030cc468edadc1ee7a7f376e/dag/src/lib.rs#L349) with some extra rounds (set to 500) and it still gives a stack overflow error:
```
        #[test]
        fn test_path_compression(
            dag in arb_dag_complete(10, 500)
        ) {
            let first = dag.first().unwrap();
            let initial_height = first.height();
            let _parents = first.parents();
            let final_height = first.height();
            assert!(final_height <= initial_height);
            assert!(first.is_trivial())
        }
```

let's catch up about it

laura-makdah: Oh my gosh, this is a really silly mistake. I meant to have `node.parents` instead of `node.parents()`. I have set this test to 500 and now I see that more work is required here, so I will close the PR for now. Thank you for bringing this test to my attention.

akichidis (approved):
Looks good @laura-makdah thanks a lot for this and apologies for the late review. I believe you are right, we might be looking on an issue with the testing framework as well causing a stackoverflow. Let's see how we could test this buy ensure testing is not the issue here.

Issue #229, opened 2022-05-11
[consensus_round] update consensus round when not using Tusk


akichidis:
### Story

Parent story: #

### Description

When not using Tusk as a consensus and we have external consensus client we still need a way to update the [consensus_round](https://github.com/MystenLabs/narwhal/blob/b7d2dfdd6394afa31cb910be0ff53efe7c6a5578/primary/src/primary.rs#L122) via some new logic.

GC logic is important both for the internal cleaning processes but also as a way to decide whether we need to wait for a certificate's [parents to become available](https://github.com/MystenLabs/narwhal/blob/b7d2dfdd6394afa31cb910be0ff53efe7c6a5578/primary/src/core.rs#L288) before proceeding.

akichidis:
CC @huitseeker one the issues we talked about on Monday

PR #378, opened 2022-06-15
[garbage_collection] link remove_collections endpoint to garbage collection process

akichidis:
resolves: https://github.com/MystenLabs/narwhal/issues/229

So far when using Narwhal without Tusk, the existing garbage collection process was not running as this depends on the `consensus_index` being updated via the last committed certificate - which do don't do.

As part of this PR, we are using the `remove_collections` endpoint as a signal to update the `garbage_collector`. Whenever a certificate is removed, we are sending this to and output channel which is consumed from the [garbage_collector](https://github.com/MystenLabs/narwhal/blob/1a06209f5c59a9b1030cc468edadc1ee7a7f376e/primary/src/garbage_collector.rs#L56), who's in turn updating the `consensus_index` value to set the baseline for the garbage collection processes across the modules.

More information about the design of this can be found [here](https://www.notion.so/mystenlabs/Design-doc-Disk-Cleanup-GC-41c4eab651334ec2bbfd822a0c412c88#b17113f9f0854f8ba985ec22de4e0f2f).

**Note:** there are some known issues with this approach which should be evaluated, but at least for now is providing some way to garbage collect our internal structures.

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -115,7 +115,7 @@ impl Node {
         State: ExecutionState + Send + Sync + 'static,
     {
         let (tx_new_certificates, rx_new_certificates) = channel(Self::CHANNEL_CAPACITY);
-        let (tx_feedback, rx_feedback) = channel(Self::CHANNEL_CAPACITY);
+        let (tx_consensus, rx_consensus) = channel(Self::CHANNEL_CAPACITY);
 
         // Compute the public key of this authority.
         let name = keypair.public().clone();
@@ -132,7 +132,7 @@ impl Node {
                 parameters.clone(),
                 execution_state,
                 rx_new_certificates,
-                tx_feedback,
+                tx_consensus.clone(),
                 tx_confirmation,
             )
             .await?;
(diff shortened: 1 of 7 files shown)


akichidis:
@asonnino please also see my comment [here](https://www.notion.so/Design-doc-Disk-Cleanup-GC-41c4eab651334ec2bbfd822a0c412c88?d=7685f913c76a4af88eef246b405d3c2c#d4dcbc887a18418baf6434065f4bbb45) whether that is a valid concern.

node/src/lib.rs
```diff
@@ -151,6 +151,7 @@ impl Node {
             /* tx_consensus */ tx_new_certificates,
             /* rx_consensus */ rx_feedback,
             /* dag */ dag,
+            tx_feedback,
```
akichidis, line 154: > Shouldn't we modify primary.rs to receive the output pipe from block_remover and feed it into the GarbageCollector module?
@asonnino basically the channel we are passing to the `block_remover` is the `tx_consensus`, which is consumed by the [GarbageCollector](https://github.com/MystenLabs/narwhal/blob/3257759165a2f9ec09157509294146adcc0c2393/primary/src/primary.rs#L199) module - so that's already there and ready to consume the deleted certificates.

It seems that we are also using the channel name `tx_feedback` and `tx_consensus` interchangeably, so I'll refactor this to stick only to one name - `tx_consensus`

PR #379, opened 2022-06-15
Task tracking for reconfiguration and graceful shutdown

sadhansood:
This dif adds a new module which allows for keeping track of tasks that were spawned in the tokio runtime and cleanly invoking shutdown and cleanup of long running tasks as needed. More tests and documentation to follow. The basic idea is we will be able to do something like:
`
let task_tracker = create_new_tracker()
task_tracker.spawn();
..
task_tracker.shutdown();
`

### DIFF
--- a/worker/src/task.rs
+++ b/worker/src/task.rs
@@ -3,11 +3,13 @@ use futures::Future;
 use once_cell::sync::OnceCell;
 use parking_lot::Mutex;
 use std::collections::BTreeMap;
+use std::thread;
 use std::{pin::Pin, sync::Arc};
 use tokio::task::JoinHandle;
 use tracing::{error, info, warn};
 
 #[derive(Clone, PartialEq)]
+#[allow(dead_code)]
 enum TaskTrackerState {
     Running,
     ShutdownStarted,
@@ -15,6 +17,7 @@ enum TaskTrackerState {
 }
 
 #[derive(Clone)]
+#[allow(dead_code)]
 struct Task {
     task_id: u64,
     tag: Option<String>,
@@ -25,6 +28,10 @@ struct Task {
 #[allow(dead_code)]
 pub type Error = tokio::sync::oneshot::error::TryRecvError;
 
+/// Wraps a [`JoinHandle`] to shutdown the underlying task when dropped
+/// unless it is detached first. Also contains a shutdown token to signal
+/// shutdown to the task
+#[allow(dead_code)]
 pub struct TaskHandle<T> {
     id: u64,
     shutdown: Shutdown,
@@ -32,14 +39,21 @@ pub struct TaskHandle<T> {
     detached: bool,
 }
 
+#[allow(dead_code)]
 impl<T> TaskHandle<T> {
+    /// Signal shutdown and return future to the
+    /// caller to wait for shutdown completion
     pub async fn stop(&self) {
         self.shutdown.shutdown();
         self.shutdown.wait_shutdown_complete().await;
     }
+    /// Signal shutdown and return
     pub fn stop_no_wait(&self) {
         self.shutdown.shutdown();
     }
+    /// Run task in detached mode which means
+    /// [`TaskHandle`] can be dropped without shutting
+    /// down the task
     pub fn detach(&mut self) {
         self.detached = true;
     }
@@ -70,12 +84,12 @@ impl<T> Drop for TaskHandle<T> {
 }
 
 /// TaskTracker is a wrapper around tokio for asynchronous tasks and futures.
-/// This class adds the following functionalities which are not present in tokio:
-/// * Tokio does not provide a way to selectively wait for certain tasks to
-///   finish executing while cancelling others during system shutdown in an easy
-///   way without keeping track of all join handles
-/// * Tokio does not provide an api to run some cleanup code when cancelling a task
-/// * TaskTracker provides an easy api to cancel a bunch of spawned tasks with a user
+/// This class adds the following functionalities which are not present in
+/// tokio today:
+/// * Tokio does not provide a way to track all spawned tasks in a central
+///   place where we can signal shutdown to some tasks and wait for their
+///   shutdown sequence to finish executing during system shutdown
+/// * TaskTracker provides an api to cancel a bunch of spawned tasks with a user
 ///   provided tag which would otherwise be difficult to do by keeping track of
 ///   tokio returned join handles
 ///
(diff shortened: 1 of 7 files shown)


worker/src/task.rs
```diff
@@ -0,0 +1,473 @@
...
+        }
+        futures::future::join_all(futures).await;
+    }
+}
+
+impl Drop for TaskTracker {
+    fn drop(&mut self) {
+        let handle = tokio::runtime::Handle::current();
+        handle.enter();
+        futures::executor::block_on(self.cancel_all_and_wait())
```
bmwill, line 444: This can cause deadlocks. In particular it is not safe to block on async inside of a drop function. There's a reason why tokio will panic if you try.

bmwill (changes requested):

laura-makdah:
I think there is a simpler overall approach to this problem that involves creating a oneshot channel and passing it to any tokio task that is spawned, and if we want to shutdown, send an empty value `()` on the channel. For any task that is launched, also listen to the channel for a shutdown in a select statement. This can be done recursively, and has the benefit of not worrying about the sender end of the channel being closed while the receiver is still open, because the sender getting closed will also cause a select statement to exit. 

This approach has the overhead of needing to refactor a lot of code and passing around a channel, but there is no way around needing to refactor to add some sort of signal, and this is the approach recommended in the tokio docs. See https://tokio.rs/tokio/topics/shutdown 

In fact, I think the tricky part of this problem is to find a way to pass along a signal that involves minimal overhead or refactoring, since we have quite a long list of tasks that get spawned.

worker/src/task.rs
sadhansood: What will cause deadlock other than when the tokio runtime is running with a single thread which is not how we are running things in prod? The only other thing that i can think of is when someone creates a bunch of task trackers  (TaskTracker) instances (it should ideally be a singleton) which is not how this is supposed to be used as well and will get caught in code reviews. The above code successfully runs in my tests but maybe I am missing the point in "There's a reason why tokio will panic if you try."
In any case, to prevent us from doing the wrong thing even by mistake, I've removed the blocking logic in drop and made it check for shutdown to have been completed beforehand or panic otherwise.

sadhansood:
@laura-makdah if I understand you correctly what you are describing actually seems more convoluted for the end user than the api's proposed in this diff. Also, there is no hard requirement to spawn every single task through this api but only the ones we care about for clean shutdown. The approach you are describing is similar to what I took but wrapped behind an easy to use api. Please take a look at the example and the doc string and we can discuss specifics offline.

worker/src/task.rs
```diff
@@ -0,0 +1,600 @@
...
+            error!("Task with task_id {task_id} doesn't exist!");
+        } else {
+            let shutdown = task.unwrap().shutdown;
+            shutdown.clone().shutdown();
+            shutdown.clone().wait_shutdown_complete().await;
+        }
+    }
+    pub async fn stop_tagged_tasks(&self, tag: String) {
+        let mut task_ids: Vec<u64> = vec![];
+        let mut futures: Vec<Pin<Box<dyn Future<Output = ()>>>> = vec![];
```
akichidis, line 482: I have a sense that the `<Box<dyn Future....` could be replaced by the [BoxFuture](https://docs.rs/futures/0.3.17/futures/future/type.BoxFuture.html)

worker/src/task.rs
```diff
@@ -0,0 +1,600 @@
...
+            }
+            let task_output = wrapped.unwrap().await;
+            if tx.send(task_output).is_err() {
+                warn!("Task with task_id {task_id} output could not be sent: receiver dropped")
+            }
+            ()
+        });
+        self.get_task_handle::<T>(tag, task_id, cloned_token, fut, detached, rx)
+    }
+    pub fn spawn_cancel_and_wait<T, U>(
```
akichidis, line 340: So if I understand correctly, the clean up here will be executed as a separate future - not related to the one we provided. For the components we use , ex the [block_waiter](https://github.com/MystenLabs/narwhal/blob/c0eafd241c368b7cba73bec9d350a037068c755e/primary/src/block_waiter.rs#L256) , we might need to perform clean ups from inside the component (object) it self. What would be your suggestion in this case? Implementing the drop method for the object to act as a clean up?

worker/src/task.rs
```diff
@@ -0,0 +1,600 @@
...
+/// let task2 = TaskTracker::global_singleton().spawn_cancel(
+///   Some("test".to_string()),
+///   async move {
+///     tokio::time::sleep(std::time::Duration::from_secs(600)).await;
+///     42
+///   }
+/// );
+/// // a task which will be canceled and instead run cleanup on shutdown
+/// let task3 = TaskTracker::global_singleton().spawn_cancel_and_wait(
+///    "test3".to_string(),
```
akichidis, line 111: I think this parameter is redundant here

worker/src/task.rs
```diff
@@ -0,0 +1,600 @@
...
+///
+/// It is recommended to use task tracker as a global singleton while it is
+/// certainly possible to have multiple instances of it. It is possible to clone
+/// and share it across multiple tasks. Each clone uses the same internal state.
+/// When we are ready to shudown, just call shutdown_and_wait() and wait for the
+/// returned future to complete.
+/// WARNING: Dropping a task tracker without invoking `shutdown_and_wait()` first
+/// will result in a panic
+///
+/// # Examples
```
akichidis, line 90: You could probably convert those to actual document tests (ex like we do in https://github.com/MystenLabs/narwhal/blob/c0eafd241c368b7cba73bec9d350a037068c755e/primary/src/block_waiter.rs#L114)

akichidis (commented):
Great work @sadhansood ! I have done a first pass on the PR, let's catch up offline as well and also evaluate this proposal Vs using a signal via channels like @laura-makdah commended. I just wanted to outline that especially for the `epoch` change stuff we might  introduce such a signal anyways for some components that we don't want to restart - but that's probably a slightly different thing anyways than the pure "shutdown" story.

worker/src/task.rs
sadhansood: done

worker/src/task.rs
```diff
@@ -0,0 +1,600 @@
...
+            if state.tasks.contains_key(&state.next_task_id) {
+                panic!(
+                    "Task tracker already running a task with next id: `{{state.next_task_id}}`"
+                );
+            }
+            task_id = state.next_task_id;
+            detached = state.task_tracker_state != TaskTrackerState::Running;
+            if detached {
+                warn!("Task arrived after task tracking is shutdown, will start as detached");
+            }
```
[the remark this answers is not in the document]
sadhansood: I agree with your comment that spawning a task after tracker has shutdown doesn't isn't ideal. We may be able to return an error on the spawn but it makes the api little ugly with Result<>. I'll think more about trade-offs with different approaches.

akichidis (commented):
@sadhansood following our discussions from previous week this proposal for graceful shutdown looks great! If you still want to go ahead and merge, please let's just move all the related `task` stuff out of the `worker` create ( this is the crate for the worker node) as I believe we want this functionality to be shared across all the code base. In this case I would recommend either create a separate create (e.x the `tasks` one) and add everything there, or probably re-use one of the existing (maybe under the  `node` one)

sadhansood:
Let me address the remaining comments and send it for another review, thanks @akichidis

worker/src/task.rs
sadhansood: I think for such a scenario, using Tasktracker::spawn(tag, closure, shutdown_token) will be ideal. User can pass int he token and use it inside the loop like in my test examples below

sadhansood:
@akichidis addressed your comments, thanks!

akichidis (approved):
Awesome work @sadhansood  💯 !

akichidis:
Let's reopen this in https://github.com/MystenLabs/sui if it's still relevant

PR #381, opened 2022-06-16
Nits

asonnino:
Fix Tasos' comments.


akichidis (approved):
Thanks for addressing those 👍

PR #385, opened 2022-06-21
Bullshark

asonnino:
Implement the partially-synchronous path of Bullshark (not the async fallaback). This includes:
* Add the consensus protocol
* Add a partially-synchronous DAG proposer
* Parametrise the node to default on Bullshark (rather than Tusk)
* Paramterise the Primary to select either the asynchronous DAG proposer or the partially-synchronous one
* Unify the common features between Tusk and Bullshark (avoid code duplication where it makes sense)
* Add unit tests for Bullshark


consensus/src/bullshark.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        // Add the new certificate to the local storage.
+        state
+            .dag
+            .entry(round)
+            .or_insert_with(HashMap::new)
+            .insert(certificate.origin(), (certificate.digest(), certificate));
+
+        // Try to order the dag to commit. Start from the highest round for which we have at least
+        // 2f+1 certificates. This is because we need them to reveal the common coin.
```
LefKok, line 49: For BabyShark even f+1 certificates are enough, in case we want to try reduce latency. 

There is also the 3f+1 (or max allowed) path to reduce latency as discussed. Should I add a conversation?

consensus/src/consensus.rs
```diff
@@ -0,0 +1,174 @@
...
+            .entry(certificate.origin())
+            .and_modify(|r| *r = max(*r, certificate.round()))
+            .or_insert_with(|| certificate.round());
+
+        let last_committed_round = *std::iter::Iterator::max(self.last_committed.values()).unwrap();
+        self.last_committed_round = last_committed_round;
+
+        // We purge all certificates past the gc depth
+        self.dag.retain(|r, _| r + gc_depth >= last_committed_round);
+        for (name, round) in &self.last_committed {
```
LefKok, line 62: What are we cleaning here?

consensus/src/consensus.rs
```diff
@@ -0,0 +1,174 @@
...
+        consensus_index: SequenceNumber,
+        // The new certificate.
+        certificate: Certificate<PublicKey>,
+    ) -> StoreResult<Vec<ConsensusOutput<PublicKey>>>;
+}
+
+pub struct Consensus<PublicKey: VerifyingKey, ConsensusProtocol> {
+    /// Receives new certificates from the primary. The primary should send us new certificates only
+    /// if it already sent us its whole history.
+    rx_primary: Receiver<Certificate<PublicKey>>,
```
LefKok, line 90: Have we ever considered output certificates before synchronizing them? Especially if we have an ordering machine and a separate execution machine in the future it might be worth not following our old convention

consensus/src/tests/bullshark_tests.rs
```diff
@@ -0,0 +1,346 @@
+// Copyright (c) 2021, Facebook, Inc. and its affiliates
```
LefKok, line 1: do we keep the FB copyright header?

consensus/src/tests/bullshark_tests.rs
```diff
@@ -0,0 +1,346 @@
...
+    for i in 1..=21 {
+        let output = rx_output.recv().await.unwrap();
+        let expected = ((i - 1) / keys.len() as u64) + 1;
+        assert_eq!(output.certificate.round(), expected);
+    }
+    let output = rx_output.recv().await.unwrap();
+    assert_eq!(output.certificate.round(), 8);
+}
+
+// Run for 5 dag rounds. The leaders of round 2 does not have enough support, but the leader of
```
LefKok, line 164:
```suggestion
// Run for 5 dag rounds. The leader of round 2 does not have enough support, but the leader of
```

consensus/src/tests/bullshark_tests.rs
```diff
@@ -0,0 +1,346 @@
...
+// round 4 does. The leader of rounds 2 and 4 should thus be committed.
```
LefKok, line 165: This scenario is not always true (in this deterministic test it is). Should we add an explanation that leader 2 is part of leader's 4 history?

consensus/src/tests/bullshark_tests.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    let name = &keys[0];
+    parents.insert(leader_2_digest);
+    let (digest, certificate) = test_utils::mock_certificate(name.clone(), 3, parents.clone());
+    certificates.push_back(certificate);
+    next_parents.insert(digest);
+
+    parents = next_parents.clone();
+
+    // Rounds 4: Fully connected graph.
```
LefKok, line 217: this is where we boost leader 2, add a comment for clarification somewhere

primary/src/async_proposer.rs
```diff
@@ -13,11 +13,11 @@ use tracing::info;
 use types::{BatchDigest, Certificate, CertificateDigest, Header, Round};
 
 #[cfg(test)]
-#[path = "tests/proposer_tests.rs"]
-pub mod proposer_tests;
+#[path = "tests/async_proposer_tests.rs"]
+pub mod async_proposer_tests;
```
LefKok, line 17: is proposer the best name? Maybe DAG_builder or block_builder?

primary/src/primary.rs
```diff
@@ -281,16 +289,28 @@ impl Primary {
 
         // When the `Core` collects enough parent certificates, the `Proposer` generates a new header with new batch
```
LefKok, line 290: fix comment

consensus/src/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+                already_ordered.insert(digest);
+            }
+        }
+    }
+
+    // Ensure we do not commit garbage collected certificates.
+    ordered.retain(|x| x.round() + gc_depth >= state.last_committed_round);
+
+    // Ordering the output by round is not really necessary but it makes the commit sequence prettier.
+    ordered.sort_by_key(|x| x.round());
```
LefKok, line 108: MEV risk? Can the validator play with their block to make it be the first of the round?  Maybe in the future we want a predefined deterministic ordering

LefKok (commented):
taking a break, need to check PartSyncProposer before approving

primary/src/part_sync_proposer.rs
```diff
@@ -0,0 +1,234 @@
...
+use tracing::info;
+use tracing::{debug, warn};
+use types::{BatchDigest, Certificate, Header, Round};
+
+#[cfg(test)]
+#[path = "tests/part_sync_proposer_tests.rs"]
+pub mod part_sync_proposer_tests;
+
+/// The proposer creates new headers and send them to the core for broadcasting and further processing.
+pub struct PartiallySyncProposer<PublicKey: VerifyingKey> {
```
LefKok, line 21: do we really need this separation? the difference of async_proposer and partially_sync_proposer is an extra check per round

primary/src/part_sync_proposer.rs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        // Send the new header to the `Core` that will broadcast and process it.
+        self.tx_core
+            .send(header)
+            .await
+            .expect("Failed to send header");
+    }
+
+    /// Update the last leader.
```
LefKok, line 112: update leader block?

LefKok (approved):

consensus/src/consensus.rs
asonnino: Yes we disscussed the for Celo and Somelier but didn't end up in our current roadmap. For Sui it is less important since "execution" simply means incrementing a sequence number (so it is very fast)

consensus/src/tests/bullshark_tests.rs
asonnino: we have to

consensus/src/tests/bullshark_tests.rs
asonnino: good point, we need to ensure that the leader of round 4 has a path to the leader of round 2. It is the case in the test, but I should update the comment.

primary/src/async_proposer.rs
asonnino: nope, I agree that dag_builder is better. We may defer this renaming for later though because most of our technical documentation uses the term "proposer"

asonnino: making an issue

consensus/src/utils.rs
asonnino: yes it is a good point. It seems that even without this validators may play games

primary/src/part_sync_proposer.rs
asonnino: let me see if we can avoid it (it would be nice)

asonnino: it seems we may, trying it in another PR. In the meantime I make an issue

primary/src/part_sync_proposer.rs
asonnino: fixed

PR #386, opened 2022-06-22
Hookup NewNetworkInfo endpoint with backend

arun-koshy:
PR for [Issue#250](https://github.com/MystenLabs/sui/issues/5294). Backend work was completed by Francois to handle the following 

> `new_network_info` is meant to convey change in validator URIs only. The validators array can include only the changed elements. Any element in the validators array which PublicKey and stake weight do not match those recorded for the epoch number will result in an error.

Just needed to hook it up with the gRPC endpoint.

Pending Tasks
- Figure out if we need to add any special logic to verify network info for a specific epoch. Will monitor @sadhansood's epoch work before adding any logic to handle that case.
- Verify addresses are reachable with some wait logic
- Investigate if any connections need to be closed and updated with new addresses. Assumption is that any connections made are done by checking the shared committee information first which will be updated with the new addresses.

### DIFF
--- a/primary/src/grpc_server/configuration.rs
+++ b/primary/src/grpc_server/configuration.rs
@@ -1,9 +1,8 @@
-use std::collections::BTreeMap;
-
-use config::{PrimaryAddresses, SharedCommittee};
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
+use config::{PrimaryAddresses, SharedCommittee};
 use crypto::traits::VerifyingKey;
+use std::collections::BTreeMap;
 use tonic::{Request, Response, Status};
 use types::{Configuration, Empty, NewEpochRequest, NewNetworkInfoRequest, PublicKeyProto};
 
--- a/primary/src/grpc_server/configuration.rs
+++ b/primary/src/grpc_server/configuration.rs
@@ -1,8 +1,9 @@
-use config::SharedCommittee;
+use std::collections::BTreeMap;
+
+use config::{PrimaryAddresses, SharedCommittee};
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 use crypto::traits::VerifyingKey;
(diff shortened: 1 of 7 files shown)


primary/src/grpc_server/configuration.rs
```diff
@@ -76,30 +93,49 @@ impl<PublicKey: VerifyingKey> Configuration for NarwhalConfiguration<PublicKey>
         request: Request<NewNetworkInfoRequest>,
     ) -> Result<Response<Empty>, Status> {
         let new_network_info_request = request.into_inner();
-        let epoch_number = new_network_info_request.epoch_number;
+        // TODO: Figure out how to verify network data against current epoch.
```
akichidis, line 96: The epoch is part of the `Committee` struct, so I would say in the simplest form we could do the comparison as part of the endpoint here, e.x:
```
if epoch_number != self.committee.epoch {
  // return error
}
```

otherwise we could refactor the `update_primary_network_info` method to accept the epoch number as well and do the validation there. For example accept the tuple `(Epoch, Stake, PrimaryAddresses)`

types/src/generated/narwhal.rs
```diff
@@ -51,7 +58,7 @@ pub struct ValidatorData {
     #[prost(int64, tag="2")]
     pub stake_weight: i64,
     #[prost(message, optional, tag="3")]
-    pub address: ::core::option::Option<MultiAddr>,
+    pub primary_addresses: ::core::option::Option<PrimaryAddresses>,
```
akichidis, line 61: Could you also please update the `docs` and `protos` files in the [examples](https://github.com/MystenLabs/narwhal/tree/main/examples) as well?

akichidis (approved):
Changes look good @arun-koshy thanks for doing this! Left a couple of comments. Also let's discuss for next steps on this

primary/src/grpc_server/configuration.rs
arun-koshy: Yeah, this should work for now. Missed the `epoch` field in the `Committee` struct. Thanks!

PR #389, opened 2022-06-22
Unify proposers

asonnino:
Unify the async and partially-sync proposer.


sadhansood:
The cleanup and refactor makes sense to me, although I am not caught up with the details around network model and synchrony details.

sadhansood (approved):

PR #390, opened 2022-06-22
Fix clippy

asonnino:
Fix https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq


types/src/generated/narwhal.rs
```diff
@@ -1,160 +1,159 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
```
arun-koshy, line 1: This file is auto generated via [bootstrap.rs](https://github.com/MystenLabs/narwhal/blob/main/types/tests/bootstrap.rs), so just wanted to check and see if these changes would get squashed if [narwhal.proto](https://github.com/MystenLabs/narwhal/blob/main/types/proto/narwhal.proto) is updated and this file gets regenerated.

arun-koshy (approved):

Issue #356, opened 2022-06-10
[metrics] introduce telemetry metrics for primary & worker nodes


akichidis:
### Description

Currently in Narwhal we do support logging but no telemetry metrics. In order to gain more insight on the nodes (primary & worker) operations, we would like to introduce telemetry metrics. We can use a similar approach that SUI took. Ultimately we would like to do the following:

1. define _which_ metrics would make sense to introduce (e.x number of batches created, latency, endpoints latencies etc)
2. introduce prometheus metrics
3. add an RPC server/gateway to collect the metrics and expose (to Grafana etc)

As reference adding the corresponding SUI [issue](https://github.com/MystenLabs/sui/issues/1455) and [PR](https://github.com/MystenLabs/sui/pull/1910)

allan-bailey:
When you add metrics, please prefix them with something like narwhal_ or sui_ to enable easier discovery for humans.
Also, namespacing to prevent collisions with other metrics.

PR #393, opened 2022-06-27
[metrics] add prometheus metrics

akichidis:
Resolves: #356 

Introduces a first set of metrics for the primary node (worker will be added separately). I've followed the approach that we have followed in SUI as well (with the potential improvement of not needing the prometheus-exporter as we are using our own server framework - but for now this is quite some effort to do). I've used the additions made on the [mysten-infra](https://github.com/MystenLabs/mysten-infra/pull/83) to support metrics for our servers. 

I expect us to open follow up PRs for metrics we want to introduce , as this will also come with experience as we start running / deploy / monitor the system.


primary/src/tests/core_tests.rs
```diff
@@ -193,13 +226,31 @@ async fn process_header_missing_payload() {
...
         tx_consensus,
         /* tx_proposer */ tx_parents,
+        metrics.clone(),
     );
 
-    // Send a header to the core.
-    let header = Header {
-        payload: [(BatchDigest::default(), 0)].iter().cloned().collect(),
-        ..header()
-    };
```
akichidis, line 202: That was a bug on the original test - the signature of the header is invalid. The test was passing, but for the wrong reasons (the header was never processed not because the payload was missing but because was invalid - so a silent error was arisen).

primary/src/metrics.rs
```diff
@@ -0,0 +1,128 @@
...
+};
+static INIT: Once = Once::new();
+
+/// Initialises the metrics. Should be called only once when the primary
+/// node is initialised, otherwise it will lead to erroneously creating
+/// multiple registries.
+pub(crate) fn initialise_metrics() -> Metrics {
+    unsafe {
+        INIT.call_once(|| {
+            let metrics_registry = default_registry();
```
akichidis, line 31: Ideally I would like to create a [custom Registry](https://github.com/tikv/rust-prometheus/blob/935d968333102c16a954a87295ea8e6c355df605/src/registry.rs#L232) here so I could define more properly the metrics `prefix` and standard labels (e.x the primary's name) so those could be applied on any metric. Unfortunately the [prometheus-exporter](https://github.com/AlexanderThaller/prometheus_exporter) is using the default registry to derive the metrics and there is not way to customise it - which is unfortunate and I am thinking that we could probably fork the project and just amend our own - after all is not a super bug lib. Unless we create our own custom server solution which we'll of course make it configurable anyways

node/Cargo.toml
```diff
@@ -33,6 +33,7 @@ primary = { path = "../primary" }
 types = { path = "../types" }
 worker = { path = "../worker" }
 workspace-hack = { version = "0.1", path = "../workspace-hack" }
+prometheus_exporter = "0.8.4"
```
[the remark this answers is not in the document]
akichidis: Done

primary/src/metrics.rs
akichidis: Done 👍  I've also allowed to pass the registry as part of the [Node::spawn_primary](https://github.com/MystenLabs/narwhal/blob/65993a63ffb23ddd6e0cc7253417682930bbf64f/node/src/lib.rs#L92) and extracted the spin up of the prometheus metrics outside of it as the primary is [spawned in SUI as well ](https://github.com/MystenLabs/sui/blob/e33bda526ab2f6714258054bf30c4effd9df05d4/crates/sui-core/src/authority_server.rs#L174) and this would lead to:
1. Spinning up an unnecessary web server to export prometheus metrics under a different port + endpoint
2. Metrics would be reported under a different registry, which probably don't want to

now you can pass the same registry to initialise in SUI node down to Narwhal nodes as well.

asonnino (approved):
LGTM but I never used prometheus (so maybe wait for the review of someone that knows what they're doing)

config/src/lib.rs
```diff
@@ -128,6 +128,24 @@ pub struct Parameters {
...
     /// The maximum number of concurrent requests for messages accepted from an un-trusted entity
     pub max_concurrent_requests: usize,
+    /// Properties for the prometheus metrics
+    pub prometheus_metrics: PrometheusMetricsParameters,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct PrometheusMetricsParameters {
+    /// Socket address the server should be listening to.
+    pub socket_addr: String,
```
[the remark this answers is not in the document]
bmwill: I think having this as a SocketAddr is ok, i'd even use the stdlib type directly instead of a string

primary/src/metrics.rs
bmwill: That's awesome! allows us to register the narwhal metrics from a sui registry

PR #425, opened 2022-06-30
chore: adding docker compose prometheus scrapes for primaries.

allan-bailey:
Also cleaned up the Dockerfile to generate a much smaller image.


akichidis (approved):
Looks great, thanks @allan-bailey !

Issue #423, opened 2022-06-30
[metrics] introduce metrics to monitor buffers


akichidis:
### Description

Following the work on [adding prometheus metrics](https://github.com/MystenLabs/narwhal/pull/393) we are looking to introduce a series of metrics to primarily monitor buffers (ex pending certificates to synchronise) across the code base (both primary & worker). As part of this PR we should introduce those metrics and expose via our standard prometheus mechanism. That will allow us to diagnose (or confirm) potential memory issues and gain more insight in general.

huitseeker:
We would want to keep track of the memory footprint of:
- the consensus store in the Tusk & Bullshark deployments:
https://github.com/MystenLabs/narwhal/blob/e124cb99c1f2a2d4970cd2684d76db24e907e40f/consensus/src/bullshark.rs#L25
https://github.com/MystenLabs/narwhal/blob/7f0b458d9379710131ecb050746978d851f17c2c/consensus/src/tusk.rs#L90
(or whatever replaces them if they get replaced <LINK TO ISSUE HERE>)
- the DAG memory footprint in DAG (NW/external consensus) deployments:
https://github.com/MystenLabs/narwhal/blob/7f0b458d9379710131ecb050746978d851f17c2c/consensus/src/dag.rs#L43-L47

Only one of the above should be active in any deployment.

We should also keep track of:

- the synchronizer buffer in the primary cert synchronizer:
https://github.com/MystenLabs/narwhal/blob/7f0b458d9379710131ecb050746978d851f17c2c/primary/src/header_waiter.rs#L78
(see also [call points in the synchrnizer](https://github.com/MystenLabs/narwhal/blob/7f0b458d9379710131ecb050746978d851f17c2c/primary/src/synchronizer.rs#L123-L124))
- the worker batch cache:
https://github.com/MystenLabs/narwhal/blob/7f0b458d9379710131ecb050746978d851f17c2c/worker/src/processor.rs#L23

/cc @velvia

akichidis:
> We would want to keep track of the memory footprint of:
> 
> * the consensus store in the Tusk & Bullshark deployments:
>   https://github.com/MystenLabs/narwhal/blob/e124cb99c1f2a2d4970cd2684d76db24e907e40f/consensus/src/bullshark.rs#L25
>   
>   https://github.com/MystenLabs/narwhal/blob/7f0b458d9379710131ecb050746978d851f17c2c/consensus/src/tusk.rs#L90
>   
[...]

PR #426, opened 2022-07-01
[metrics] introduce more metrics across primary, worker & consensus

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/423

This PR adds more metrics across the primary, worker & consensus components. Goal is to enhance our insights and help us diagnose potential memory leak issues.


huitseeker (approved):
Thanks a bunch! I've added a few comments, but they are mostly nits.

consensus/src/dag.rs
```diff
@@ -286,12 +294,28 @@ impl<PublicKey: VerifyingKey> InnerDag<PublicKey> {
...
+
+        self.metrics
+            .external_consensus_dag_vertices_elements
+            .with_label_values(&[])
+            .set(vertices.len() as i64);
+
+        self.metrics
+            .external_consensus_dag_size
+            .with_label_values(&[])
+            .set(self.dag.size() as i64)
```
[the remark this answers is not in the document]
akichidis: Absolutely, that should be the case. Unfortunately given the current separation dag structs we can't do much.

consensus/src/metrics.rs
```diff
@@ -0,0 +1,52 @@
...
+use prometheus::{default_registry, register_int_gauge_vec_with_registry, IntGaugeVec, Registry};
+
+#[derive(Clone, Debug)]
+pub struct ConsensusMetrics {
+    /// The number of elements in the Dag (for Tusk or Bullshark)
+    pub consensus_dag_size: IntGaugeVec,
+    /// The last committed round from consensus
+    pub last_committed_round: IntGaugeVec,
+    /// The number of elements from the vertices secondary index (external consensus)
+    pub external_consensus_dag_vertices_elements: IntGaugeVec,
```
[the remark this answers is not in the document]
akichidis: Because we always derive the size of the vertices map and we report them to metrics - if we would use the `IntCounterVec` we would end up always adding up the sizes one on top of the other. Now we just always report the total which `gauge` is for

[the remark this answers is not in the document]
akichidis: Same stands for the other metrics

primary/src/metrics.rs
```diff
@@ -56,6 +56,12 @@ pub struct PrimaryMetrics {
     pub batches_received: IntCounterVec,
     /// Latency to perform a garbage collection in core module
     pub gc_core_latency: HistogramVec,
+    /// The current Narwhal round
+    pub current_round: IntGaugeVec,
+    /// Latency to perform a garbage collection in header_waiter
+    pub gc_header_waiter_latency: HistogramVec,
+    /// Number of elements in pending list of header_waiter
+    pub pending_elements_header_waiter: IntGaugeVec,
```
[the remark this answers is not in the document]
akichidis: Same with the above comment, since we are basically getting a "snapshot" of the pending elements on the list - which doesn't necessarily correlate with the previous report instance (we don't always expect this to go up) I've used the gauge

worker/src/metrics.rs
```diff
@@ -0,0 +1,57 @@
...
+
+            METRICS = Metrics {
+                worker_metrics: Some(node_metrics),
+            }
+        });
+        METRICS.clone()
+    }
+}
+
+#[derive(Clone)]
```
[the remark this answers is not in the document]
akichidis: Because this is used inside the [Option](https://github.com/MystenLabs/narwhal/pull/426/files#diff-82e6ef919a8598919323ec7291bb8146821c64739d319b4ae7b12864fd534a96R8) in Metrics we need to have clone implemented

consensus/src/metrics.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// The number of elements in the dag (external consensus)
+    pub external_consensus_dag_size: IntGaugeVec,
+}
+
+impl ConsensusMetrics {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            consensus_dag_size: register_int_gauge_vec_with_registry!(
```
velvia, line 20: Sorry I'm late to the party.

The reason to use a `vec` is to add labels so you can have different time series, for example it could be for different workers, or something.  If you aren't using any labels, then instead you should just use the simple `int_gauge`, not one with a vec.

consensus/src/metrics.rs
```diff
@@ -0,0 +1,52 @@
...
+                registry
+            ).unwrap(),
+            last_committed_round: register_int_gauge_vec_with_registry!(
+                "last_committed_round",
+                "The most recent round that has been committed from consensus",
+                &[],
+                registry
+            ).unwrap(),
+            external_consensus_dag_vertices_elements: register_int_gauge_vec_with_registry!(
+                "external_consensus_dag_vertices_elements",
```
velvia, line 33: Might want to prefix all Narwhal stats with `narwhal_` just for ease of discovery

primary/src/header_waiter.rs
```diff
@@ -334,7 +343,18 @@ impl<PublicKey: VerifyingKey> HeaderWaiter<PublicKey> {
...
+                self.metrics
+                    .gc_header_waiter_latency
+                    .with_label_values(&[&self.committee.epoch.to_string()])
+                    .observe(now.elapsed().as_secs_f64());
             }
+
+            // measure the pending elements
+            self.metrics
+                .pending_elements_header_waiter
+                .with_label_values(&[&self.committee.epoch.to_string()])
```
velvia, line 356: Are you sure you want to use epoch as a label value?  Aren't epochs something that changes over time?  This will keep creating new time series, meaning that you cannot then easily correlate old epochs with new epochs.  Usually something that evolves over time is kept in the same time series.

velvia (commented):
Just some comments about use of time series

primary/src/header_waiter.rs
allan-bailey: Labels must be bounded in scope and not unlimited.  Otherwise it will break Prometheus.
If the epoch is just a counter, then it could be exposed that way.

consensus/src/metrics.rs
akichidis: We do currently prefix all the metrics when we initialize the Registry for the primary and worker nodes https://github.com/MystenLabs/narwhal/blob/c0e057a1d2e22873238731ea8f053afdd1b01edd/node/src/metrics.rs#L14

I did it this way to ensure that those will be automatically prefixed irrespective of whether we name them that way or not . The only downside is that we we embed Narwhal in SUI and we use a common registry we lose this feature

primary/src/header_waiter.rs
akichidis: I chose to use the echo as part of the metrics primarily for debugging reasons until we get good confidence on how things work (epoch is a fairly new thing). Having this as part of the metric it self it will make it easy to track how node's behaviour change - keep in mind that epochs change on the fly. I am all for removing epoch once we don't feel we need it anymore,

@allan-bailey we expect epoch to change once per day - I am aware of the limitations of the cardinality on the labels - so I would err on keeping as is before we remove it as we don't look on that many different values on the upcoming period.

consensus/src/metrics.rs
akichidis: Ah thanks for this , I am going to have a look on it.

PR #427, opened 2022-07-01
Expose primary handles

asonnino:
These will be useful later.


akichidis (approved):
Changes look good to me. What's the intention, could you please provide more context on the end goal?

asonnino:
The idea is to be able to wait on the handles (Marc seems to have issues with `main` exiting too early when running the memory profiler)

PR #430, opened 2022-07-03
upgrade to Rust 1.62

huitseeker:
+ convert instances of `.then_some()`


akichidis (approved):
Thanks a lot for doing this @huitseeker !

Nit: let's also update the badge [here](https://github.com/MystenLabs/narwhal/blob/dc286af1598e95efc2f63acf1d3a252179d7f88c/README.md?plain=1#L4) & the [Docker image](https://github.com/MystenLabs/narwhal/blob/dc286af1598e95efc2f63acf1d3a252179d7f88c/Docker/Dockerfile#L9) to the `1.62` version as well to make sure we are aligned.

PR #448, opened 2022-07-04
Add DHAT memory profiling to benchmark

velvia:
This PR adds the option to build the benchmark to use the [DHAT](https://docs.rs/dhat/latest/dhat/index.html#viewing) memory profiler.   See the extra section in README for instructions on how to use it - you can simply turn it on via `mem_profiling` benchmark option in `fabfile.py`.

One side effect of profiling is that debug symbols are needed in release binaries.  Maybe there is a way to make this conditional on using the `dhat-heap` feature flag....

There is a hack to write out the profile after n seconds/minutes, welcome to suggestions to make this better or configurable.  The idea is to however write the profile before the process is killed by the benchmark framework.  I know @asonnino had some ideas.


akichidis (approved):
Thanks so much @velvia , super useful!

node/src/lib.rs
```diff
@@ -2,11 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 use config::{Parameters, SharedCommittee, WorkerId};
 use consensus::{bullshark::Bullshark, dag::Dag, Consensus, SubscriberHandler};
-use crypto::traits::{KeyPair, Signer, VerifyingKey};
+use crypto::traits::{EncodeDecodeBase64, KeyPair, Signer, VerifyingKey};
```
huitseeker, line 5: I'd love to see a rebase because of this an many other small errors (this import is unneeded).

primary/Cargo.toml
```diff
@@ -28,6 +28,7 @@ tokio-util = { version = "0.7.3", features = ["codec"] }
 tonic = "0.7.2"
 tower = "0.4.12"
 tracing = "0.1.34"
+dhat = "0.3"
```
huitseeker, line 31: I think it's worth making the dependency optional, and triggered by the feature.
```
[dependencies]
foo = { version = "1.0", optional = true }
bar = { version = "1.0", optional = true }

[features]
fancy-feature = ["foo", "bar"]
```

benchmark/fabfile.py
```diff
@@ -21,7 +21,8 @@ def local(ctx, debug=True):
         'workers': 1,
         'rate': 50_000,
         'tx_size': 512,
-        'duration': 20,
+        'duration': 300,
```
huitseeker, line 24: While I 100% understand the longer run time for memory profiling, I'd love to not change the default run time for non-memory-profiling benches.

Cargo.toml
```diff
@@ -4,6 +4,8 @@ members = ["config", "consensus", "crypto", "dag", "executor", "examples", "netw
 [profile.release]
 codegen-units = 1
 lto = true
+# Important for performance/memory profiling
+debug = 1
```
huitseeker, line 8: Nit: I'm more used to `debug = true`

Cargo.toml
```diff
@@ -4,6 +4,8 @@ members = ["config", "consensus", "crypto", "dag", "executor", "examples", "netw
 [profile.release]
 codegen-units = 1
 lto = true
+# Important for performance/memory profiling
```
huitseeker, line 7: Cargo now allows [custom profiles](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles). How about creating a benchmark profile with debug (e.g. "release-profiled"), which would contain the option and edit `CommandMaker` to do `cargo profile --profile release-profiled`

node/src/main.rs
```diff
@@ -114,6 +118,7 @@ async fn main() -> Result<()> {
         ("run", Some(sub_matches)) => run(sub_matches).await?,
         _ => unreachable!(),
     }
+    println!("Exiting main()...");
```
huitseeker, line 121: Do we want to keep this?

benchmark/benchmark/local.py
```diff
@@ -52,7 +52,8 @@ def run(self, debug=False):
             sleep(0.5)  # Removing the store may take time.
 
             # Recompile the latest code.
-            cmd = CommandMaker.compile().split()
+            cmd = CommandMaker.compile(mem_profiling = self.mem_profile)
+            Print.info(f"About to run {cmd}...")
```
huitseeker, line 56: Please reflect in `benchmark` if you mean to keep this.

huitseeker (changes requested):
This mostly looks good, thanks Evan! 
Main issue: the change to the default runtime of the benchmark is killing CI, and should be conditioned on the `dhat` option.

node/src/lib.rs
velvia: Actually it's used to name the dhat-heap profile filename so it has a unique filename for each node.

Cargo.toml
velvia: Sure.  Although it doesn't seem like profiles enable feature selection, unfortunately.

node/src/main.rs
velvia: Probably not, removed

benchmark/benchmark/local.py
velvia: Not sure what you mean by that.... mem_profile is a new instance variable based on the config, so this change should be fine....

velvia:
@huitseeker feedback addressed and rebased

benchmark/fabfile.py
```diff
@@ -21,7 +21,8 @@ def local(ctx, debug=True):
         'workers': 1,
         'rate': 50_000,
         'tx_size': 512,
-        'duration': 20,
+        'duration': 10,
```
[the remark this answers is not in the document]
velvia: ok this is changed back now

benchmark/benchmark/local.py
huitseeker: I'm sorry, I misspoke: I meant please reflect, in `remote`, I.e. here: https://github.com/MystenLabs/narwhal/blob/3d921e757acdadf3e61fbf32b151e23f08d89ad2/benchmark/benchmark/remote.py#L169-L170

You're documenting the memory options very well in the `README`, but this `README` documents in a single explanation that you can run the benchmarks in both  `local` and `remote` modes, with the assumption those run more or less identically.

It would be great to therefore have the ability to run remote memory profiling ... or document explicitly that this will only work in `local` mode

benchmark/benchmark/commands.py
```diff
@@ -17,8 +17,12 @@ def clean_logs():
...
 
     @staticmethod
-    def compile():
-        return 'cargo build --quiet --release --features benchmark'
+    def compile(mem_profiling):
+        if mem_profiling:
+            params = ["--profile", "bench-profiling", "--features", "benchmark dhat-heap"]
+        else:
+            params = ["--release", "--features", "benchmark"]
+        return ["cargo", "build", "--quiet"] + params
```
huitseeker, line 25: Perfect!

huitseeker (approved):
Thanks a lot for setting this up!
I have just one comment clarifying my earlier confusing remark about "reflecting" the passing of the memory profiling option in `local.py`. I have no doubt you'll tweak the PR to a fine outcome.

velvia:
@huitseeker ahhh thanks.... the wonders of Python, where I have no idea something will break due to dynamic typing.  :)

velvia:
@huitseeker there is something in the CI workflow that tests every feature?  Is there a way to turn off the `dhat-heap` feature I added for testing, as it's not meant to be used in tests?

huitseeker:
> there is something in the CI workflow that tests every feature? 
Yes.
```
huitseeker@Garillots-MBP.localdomain➜~/tmp/narwhal(main)» rg 'all-features' ./.github                                                                                            [19:34:23]
./.github/workflows/nightly.yml
48:          args: --all-features
60:          args: --all-targets --all-features --release --profile=release

./.github/workflows/rust.yml
72:          cargo nextest run --all-features --profile ci
75:          cargo test --doc --all-features
```

> Is there a way to turn off the dhat-heap feature I added for testing, as it's not meant to be used in tests?

You can simply add a commit that changes the `--all-features` argument to test into a `--features celo`

velvia:
@huitseeker thanks for helping merge it!

PR #452, opened 2022-07-05
Restart Primary

asonnino:
Reconfigure the primary by restarting it with a new committee.


huitseeker (approved):
This works for me, thanks for the test! 
You'll want to rebase to take in the modifications around the committee structure as part of #450 and #453.

PR #454, opened 2022-07-05
fix: turn off logging stderr+out to file

allan-bailey:
Removing the log file so we can just log to the container
stdout and stderr to be picked up by Loki or other means.

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -36,23 +36,14 @@ since the narwhal node binary needs to be built from the source code.) And then 
 a cluster for *four nodes* by doing the necessary setup for `primary` and `worker` nodes. Each
 `primary` node will be connected to *one worker* node.
 
-The logs for each validator (primary & worker nodes) can be found in the `logs` folder under the corresponding
-validator folder.
-
-> **Note**: You will find across the codebase the name *Authority* when we describe
-> the entity participating in the Narwhal protocol that verifies and signs the produced headers. An 
-> authority is composed of a `primary` node and one or more `worker` nodes. Since Authority is not a widely
-> used term, we are referring instead to our *primary + worker* node pairs as **Validator**.
-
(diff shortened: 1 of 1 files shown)


huitseeker (commented):
Do we have an issue to bring in shipping the logs to loki, through e.g. the [tracing subscriber](https://github.com/hrxi/tracing-loki)? Do we moreover have some documentation on how to retrieve those logs through those alternate means (loki or otherwise)?

I'm worried about this suppressing functionality without an alternative in place.

allan-bailey:
> 

I'll create a tracking issue for making logs easily accessible via Loki, or if you're running in docker-compose, the logs would be visible via `docker compose logs`

issue: #455

akichidis:
@allan-bailey I've left a comment on the ticket as well, but to make sure it's not missed, let's make sure that [README](https://github.com/MystenLabs/narwhal/blob/6e86be42032de550764368c4a8a66e1472c68445/Docker/README.md?plain=1#L47) is updated and tell the user where they should expect to find the logs.

huitseeker (commented):
> I'll create a tracking issue for making logs easily accessible via Loki, or if you're running in docker-compose, the logs would be visible via docker compose logs

Sounds fair to me. Could you open the Loki issue and update the section on logs in the [Docker README](https://github.com/MystenLabs/narwhal/blob/main/Docker/README.md#quick-start) with a sentence on e.g. `docker compose logs`, that should round up the present PR in a shippable milestone state, I think.

allan-bailey:
Once this new Dockerfile + entry.sh script is on main, I'll be able to fix our workflow to build and push the container image to docker hub.

huitseeker (approved):
Superb, thanks @allan-bailey !

allan-bailey:
Fixes #455

PR #460, opened 2022-07-07
crypto: Add Sui + Aggregation support

punwai:
Added the support for any type of Aggregated Signature.


crypto/src/bls12381.rs
```diff
@@ -338,8 +346,8 @@ impl KeyPair for BLS12381KeyPair {
         &self.name
     }
 
-    fn private(self) -> Self::PrivKey {
-        self.secret
+    fn private(&'_ self) -> &'_ Self::PrivKey {
```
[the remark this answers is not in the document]
huitseeker: Copying a keypair, and by extension a private key, is not something we want to have public APIs for, and we absolutely do not want to have ways to do that outside tests.

One approach you can use if you absolutely need such a copy ability in tests is to serialize and deserialize the `KeyPair`. This sort of function should be gated by [conditional compilation](https://doc.rust-lang.org/reference/conditional-compilation.html).

crypto/src/bls12381.rs
```diff
@@ -358,6 +368,24 @@ impl KeyPair for BLS12381KeyPair {
             },
         }
     }
+
+    fn deterministic_generate(seed: &[u8], id: &[u8], domain: &[u8]) -> Result<Self, signature::Error> {
```
huitseeker, line 372: I'd also like a size check on the seed's length.

crypto/src/traits.rs
```diff
@@ -122,14 +126,44 @@ pub trait Authenticator:
...
 ///
-pub trait KeyPair {
+pub trait KeyPair: Sized {
     type PubKey: VerifyingKey<PrivKey = Self::PrivKey>;
     type PrivKey: SigningKey<PubKey = Self::PubKey>;
     fn public(&'_ self) -> &'_ Self::PubKey;
-    fn private(self) -> Self::PrivKey;
+    fn private(&'_ self) -> &'_ Self::PrivKey;
     fn generate<R: CryptoRng + RngCore>(rng: &mut R) -> Self;
+    fn deterministic_generate(seed: &[u8], id: &[u8], domain: &[u8]) -> Result<Self, signature::Error>;
```
huitseeker, line 140: Some field documentation would help.

The schemes we use do not tie themselves specifically to a KDF solution: do we want to maintain such a tight coupling between a signature scheme and a KDF function?

One possible way to address this would be to KDF generically, there is a trait doing this here for inspiration:
https://github.com/diem/diem/blob/416fdd3b18d02b2a099917467910d5a6865f2fed/crypto/crypto/src/hkdf.rs

crypto/src/ed25519.rs
```diff
@@ -197,6 +291,38 @@ impl KeyPair for Ed25519KeyPair {
             secret: Ed25519PrivateKey(kp.secret),
         }
     }
+
+    fn deterministic_generate(seed: &[u8], id: &[u8], domain: &[u8]) -> Result<Self, signature::Error> {
```
huitseeker, line 295: I like that you're proactive about this and not leaving the user to figure out KDF on their own.

But here we end up essentially copying the implementation from the BLS variant, which hurts maintenability.

Here's one approach that leads to one version of this code:
- a generic variant of the KDF, that's just parametrized in each implementation,
- or simply an `ikm -> KeyPair` map on the trait, that's then used at the end of a [default implementation](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations) leveraging the KDF.

crypto/src/traits.rs
```diff
@@ -122,14 +126,44 @@ pub trait Authenticator:
...
+    fn aggregate(signatures: Vec<Self::Sig>) -> Result<Self, Error>;
+
+    fn add_signature(&mut self, signature: Self::Sig) -> Result<(), Error>;
+
+    fn add_aggregate(&mut self, signature: Self) -> Result<(), Error>;
+
+    /// Borrow a byte slice representing the serialized form of this key
+    fn verify(
+        &self,
+        pks: &[&<Self::Sig as Authenticator>::PubKey],
```
huitseeker, line 160: So there's two approaches here: 
- conservative: require pubkeys to be stored in a contiguous array in memory, in which case you would want a `pks: &[<Self::Sig as Authenticator>::PubKey]`,
- liberal: require pubkeys to be something you can iterate over, without constraints on memory location.

In the liberal case, note that you could use an iterator type here. Consider using [`Borrow`](https://doc.rust-lang.org/std/borrow/trait.Borrow.html#examples) for its item type. 

There's one example of this [here](https://github.com/MystenLabs/mysten-infra/blob/7ce3f9bba93c93d87d02ec921ce7312c6f6dd0c4/crates/typed-store/src/rocks/mod.rs#L209-L212).

crypto/src/traits.rs
```diff
@@ -122,14 +126,44 @@ pub trait Authenticator:
...
     fn generate<R: CryptoRng + RngCore>(rng: &mut R) -> Self;
+    fn deterministic_generate(seed: &[u8], id: &[u8], domain: &[u8]) -> Result<Self, signature::Error>;
+}
+
+pub trait AggregateAuthenticator:
+    Display + Default + Serialize + DeserializeOwned + Send + Sync + 'static + Clone
+{
+    type Sig: Authenticator;
+    type PubKey: VerifyingKey<Sig = Self::Sig>;
+    type PrivKey: SigningKey<Sig = Self::Sig>;
```
huitseeker, line 148: Filling in associated type constraints on `Sig` here (`<PubKey = Self::PubKey ...>`) should allow you to unify the types and not need a double projection later (in `verify`).

huitseeker (changes requested):
This is promising and full of good ideas! The main thing I'd push back on is the access to the keypair's private key bytes, and then we may want to chat about the way you're coupling the KDF with signing.

crypto/src/traits.rs
punwai: I've changed everything to fit with the conservative approach!

crypto/src/traits.rs
```diff
@@ -119,17 +124,91 @@ pub trait Authenticator:
...
+    + Eq
+    + std::hash::Hash
+    + Copy
+    + Ord
+    + PartialOrd
+    + ToFromBytes
+{
+    type PubKey: VerifyingKey<Bytes = Self>;
+
+    fn to_vec(&self) -> Vec<u8>;
```
huitseeker, line 213: I'm not sure this is an improvement from `pub_key.as_ref().to_vec()`, but even if it is, such a default implementation would avoid us the trouble of implementing this in each scheme, no?

crypto/src/bls12381.rs
```diff
@@ -258,10 +261,57 @@ impl Display for BLS12381Signature {
...
+        S: serde::Serializer,
+    {
+        if serializer.is_human_readable() {
+            self.encode_base64().serialize(serializer)
+        } else {
+            let mut seq = serializer.serialize_tuple(BLS_SIGNATURE_LENGTH)?;
+            for element in self.sig.to_bytes() {
+                seq.serialize_element(&element)?;
+            }
+            seq.end()
```
huitseeker, line 287: OK, so for both serialization and deserialization of bytes, you'll want to have a look at this explanation:
https://github.com/serde-rs/bytes#explanation
And then for the actual implementation, you'd want to use the approach in `serde_with`, which is leveraging the same technique:
https://docs.rs/serde_with/latest/serde_with/struct.Bytes.html
(serde_with is already in crate imports)

crypto/src/traits.rs
```diff
@@ -102,6 +106,7 @@ pub trait VerifyingKey:
 pub trait SigningKey: ToFromBytes + Serialize + DeserializeOwned + Send + Sync + 'static {
     type PubKey: VerifyingKey<PrivKey = Self>;
     type Sig: Authenticator<PrivKey = Self>;
+    const LENGTH: usize;
```
huitseeker, line 109: What's the intent of this LENGTH? I understand it may afford you checks when manipulating an aggregate signature, but I would expect some uses of `Self::LENGTH` in the code to reveal the utility.

crypto/src/traits.rs
```diff
@@ -119,17 +124,91 @@ pub trait Authenticator:
     + Sync
     + 'static
     + Clone
+    + EncodeDecodeBase64
```
huitseeker, line 127: 👍

crypto/Cargo.toml
```diff
@@ -30,7 +35,6 @@ blst = "0.3.10"
 once_cell = "1.12.0"
 readonly = "0.2.1"
 serde_bytes = "0.11.6"
-serde_with = { version = "1.14.0", optional = true }
```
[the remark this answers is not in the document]
punwai: I just moved it somewhere else because I needed it.

crypto/src/traits.rs
punwai: It's pretty useful in the Sui repo.

punwai (commented):
e

crypto/src/traits.rs
punwai: This is true

crypto/src/bls12381.rs
punwai: I tried this but was miserably failing, I will attempt it again.

punwai: Fixed.

crypto/src/tests/ed25519_tests.rs
```diff
@@ -143,6 +143,155 @@ fn verify_invalid_batch() {
...
+    pubkeys[0] = keys()[3].public().clone();
+
+    // // Verify the batch.
+    let res = aggregated_signature.verify(&pubkeys[..], &digest.0);
+    assert!(res.is_err(), "{:?}", res); 
+}
+
+#[test]
+fn verify_batch_aggregate_signature() {
+    // Make signatures.
```
punwai, line 213: The tests in all 3 files are the same.

crypto/src/bls12381.rs
```diff
@@ -258,10 +266,22 @@ impl Display for BLS12381Signature {
...
+    fn encode_base64(&self) -> String {
+        base64ct::Base64::encode_string(self.as_bytes())
+    }
+
+    fn decode_base64(value: &str) -> Result<Self, eyre::Report> {
+        let bytes = base64ct::Base64::decode_vec(value).map_err(|e| eyre!("{}", e.to_string()))?;
+        BLS12381Signature::from_bytes(&bytes).map_err(|e| e.into())
+    }
+}
+
```
huitseeker, line 279: This should go away, see my comment at the top of `crypto/src/traits.rs`

crypto/src/ed25519.rs
```diff
@@ -131,10 +156,22 @@ impl<'de> Deserialize<'de> for Ed25519PrivateKey {
     }
 }
 
+impl EncodeDecodeBase64 for Ed25519Signature {
```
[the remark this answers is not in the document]
huitseeker: You can. This should go away, see my comment at the top of `crypto/src/traits.rs`

huitseeker:
https://github.com/punwai/narwhal-1/pull/3

huitseeker (commented):
There's still more simplifications. Committing my feedback gradually so you don't have to wait.

huitseeker:
https://github.com/punwai/narwhal-1/pull/4

crypto/src/traits.rs
```diff
@@ -122,14 +135,102 @@ pub trait Authenticator:
...
+        const SHA3_OUTPUT_SIZE: usize = <Sha3_256 as OutputSizeUser>::OutputSize::USIZE;
+        let mut okm = [0u8; SHA3_OUTPUT_SIZE];
+        hk.expand(info, &mut okm)
+            .map_err(|_| signature::Error::new())?;
+
+        let secret_key = Self::PrivKey::from_bytes(ikm).map_err(|_| signature::Error::new())?;
+
+        let keypair = Self::from(secret_key);
+        Ok(keypair)
+    }
```
huitseeker, line 187: I have solved the problem that this `Sha3` based implementation may not produce hashes of the suitable length for the byte representation of the private key, in https://github.com/punwai/narwhal-1/pull/5

crypto/src/traits.rs
```diff
@@ -122,14 +142,123 @@ pub trait Authenticator:
...
+
+    let keypair = K::from(secret_key);
+    Ok(keypair)
+}
+
+/// Trait impl'd by aggregated signatures in asymmetric cryptography.
+///
+/// The trait bounds are implemented to allow the aggregation of multiple signatures,
+/// and to verify it against multiple, unaggregated public keys. For signature schemes
+/// where aggregation is not possible, a trivial implementation is provided.
```
huitseeker, line 222: Here's one main thing that's bugging me:
- in this trait, you have access to all the facilities implemented in the lower traits: Authenticator, PubKey, Privkey, ...
- You mention "where aggregation is not possible, a trivial implementation is provided."

You do provide this trivial implementation on ed25519: Is it possible to define one as a default implementation right here in the trait?

crypto/src/traits.rs
```diff
@@ -122,14 +142,123 @@ pub trait Authenticator:
...
+/// # fn main() {
+///     let ikm = b"some_ikm";
+///     let domain = b"my_app";
+///     let salt = b"some_salt";
+///     let my_keypair = hkdf_generate_from_ikm::<Sha3_256, Ed25519KeyPair>(ikm, salt, Some(domain));
+///
+///     let my_keypair_default_domain = hkdf_generate_from_ikm::<Sha3_256, Ed25519KeyPair>(ikm, salt, None);
+/// # }
+/// ```
+pub fn hkdf_generate_from_ikm<'a, H: OutputSizeUser, K: KeyPair>(
```
joyqvq, line 184: might have missed this elsewhere, this looks like a utility function to generate keypair, should this live inside `pub trait KeyPair`?

crypto/src/serde_helpers.rs
```diff
@@ -0,0 +1,151 @@
...
+        ed25519_dalek::Signature::from_bytes(&bytes).map_err(to_custom_error::<'de, D, _>)
+    }
+}
+
+pub trait Encoding {
+    fn decode(s: &str) -> Result<Vec<u8>, anyhow::Error>;
+    fn encode<T: AsRef<[u8]>>(data: T) -> String;
+}
+
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
```
joyqvq, line 92: how is the JsonSchema used?

huitseeker (commented):
This should simplify the `PublicKeyBytes` part:
https://github.com/punwai/narwhal-1/pull/6/

I would really like to get the test coverage percentage a bit higher today, and merge this.

crypto/src/traits.rs
```diff
@@ -122,14 +142,123 @@ pub trait Authenticator:
...
-pub trait KeyPair {
+pub trait KeyPair: Sized + From<Self::PrivKey> {
     type PubKey: VerifyingKey<PrivKey = Self::PrivKey>;
     type PrivKey: SigningKey<PubKey = Self::PubKey>;
+    type Sig: Authenticator<PubKey = Self::PubKey>;
+
     fn public(&'_ self) -> &'_ Self::PubKey;
     fn private(self) -> Self::PrivKey;
+
+    #[cfg(feature = "copy_key")]
```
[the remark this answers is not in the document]
huitseeker: We still have a trailing need to copy some private keys  ( 🙄 ) in Sui, notably for testing (Narwhal testing is instead seeding a Rng deterministically to solve the same issue). I'd like to merge the two crypto libs and then fix that.

crypto/src/serde_helpers.rs
```diff
@@ -0,0 +1,151 @@
...
+            let s = String::deserialize(deserializer)?;
+            Base64::decode(&s).map_err(to_custom_error::<'de, D, _>)?
+        } else {
+            Bytes::deserialize_as(deserializer)?
+        };
+        ed25519_dalek::Signature::from_bytes(&bytes).map_err(to_custom_error::<'de, D, _>)
+    }
+}
+
+pub trait Encoding {
```
huitseeker, line 87: I'm confused as to why we need to re-import all of this, which IIUC is essentially a copy of the base64 crate with a slightly different API: what am I missing?

punwai: Removed the reimport and just used native functions - I ported this over from the Sui codebase and it was useful there because it shortened a lot of other code. No real need here.

punwai:
@huitseeker Also, would it be possible to remove the `cargo test` from the Code Coverage check? AFAIK it's already being done on another workflow and just makes code coverage checking directly on Github annoyingly long.

huitseeker:
> Also, would it be possible to remove the cargo test run from the Code Coverage check? AFAIK it's already being done on another workflow and just makes code coverage checking directly on Github annoyingly long.

Code coverage computations are done through code instrumentation and can't be computed without running tests.

crypto/src/hkdf.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// Creation of a keypair using the [RFC 5869](https://tools.ietf.org/html/rfc5869) HKDF specification.
+/// This requires choosing an HMAC function of the correct length (conservatively, the size of a private key for this curve).
+/// Despite the unsightly generics (which aim to ensure this works for a wide range of hash functions), this is straightforward to use.
+///
+/// Example:
+/// ```rust
+/// use sha3::Sha3_256;
+/// use crypto::ed25519::Ed25519KeyPair;
+/// use crypto::traits::hkdf_generate_from_ikm;
```
huitseeker, line 21:
```suggestion
/// use crypto::hkdf::hkdf_generate_from_ikm;
```

crypto/src/hkdf.rs
```diff
@@ -0,0 +1,85 @@
...
+///     let domain = b"my_app";
+///     let salt = b"some_salt";
+
+///     let my_keypair = hkdf_generate_from_ikm::<Sha3_256, BLS12381KeyPair>(ikm, salt, Some(domain)).unwrap();
+///     let native_sk = blst::min_sig::SecretKey::key_gen_v4_5(ikm, salt, domain).unwrap();
+
+///     assert_ne!(native_sk.to_bytes(), my_keypair.private().as_bytes());
+/// # }
+/// ```
+pub fn hkdf_generate_from_ikm<'a, H: OutputSizeUser, K: KeyPair>(
```
huitseeker, line 51: 👍

node/tests/staged/narwhal.yaml
```diff
@@ -44,10 +44,7 @@ Committee:
 Ed25519PublicKey:
   NEWTYPESTRUCT: STR
 Ed25519Signature:
-  NEWTYPESTRUCT:
-    TUPLEARRAY:
-      CONTENT: U8
-      SIZE: 64
+  NEWTYPESTRUCT: BYTES
```
huitseeker, line 47: So this might be a slight issue, but we can certainly address it later.

huitseeker (approved):
This is a herculean amount of excellent, quality work! Thank you so much! 🎉 

PS: checked, no negative impact on benches.

PR #462, opened 2022-07-08
fix: update timestamp of requests when we re-issue them

huitseeker:
This addresses a too-frequent-retry behavior, whereby:
- we set a time of request when queuing one,
- in a loop, reexamine the requests with a high examination frequency,
- reissue the requests older than a retry delay,
- and don't modify the queued request in any way.

The intent of this code was to reissue the requests at the retry frequency, but since we never update the request's issuance time, it ends up reissuing all requests on every iteration, at the (much higher) examination frequency.

### DIFF
--- a/primary/src/header_waiter.rs
+++ b/primary/src/header_waiter.rs
@@ -295,10 +295,12 @@ impl<PublicKey: VerifyingKey> HeaderWaiter<PublicKey> {
                         .as_millis();
 
                     let mut retry = Vec::new();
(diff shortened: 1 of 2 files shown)


arun-koshy (approved):

akichidis (approved):
Thanks for the fix @huitseeker and for fixing the worker synchronizer as well! 

Let's have a follow up PR at some point to add tests around the intended behaviour as well

PR #463, opened 2022-07-08
Limit number of tasks spawned for sending messages

mystenmark:


network/src/primary.rs
```diff
@@ -72,17 +81,19 @@ impl PrimaryNetwork {
...
-                })
-                .map(|response| {
-                    response.expect("we retry forever so this shouldn't fail");
-                }),
-        );
+        let handle = self
+            .spawn(
+                self.retry_config
+                    .retry(move || {
+                        let mut client = client.clone();
```
velvia, line 88: How heavyweight is this `client.clone()`?

velvia (approved):
Can we add a gauge to track the number of concurrent tasks?  That would be super useful.

network/src/primary.rs
mystenmark: I don't know, but hopefully it bottoms out in an Arc::clone somwhere, and this precedes my change

mystenmark:
> Can we add a gauge to track the number of concurrent tasks? That would be super useful.

I added a log for when we hit the limit and start waiting

network/src/lib.rs
```diff
@@ -40,3 +46,35 @@ impl<T> std::future::Future for CancelHandler<T> {
...
+) -> JoinHandle<()> {
+    match acquire_permit(semaphore).await {
+        Some(permit) => tokio::spawn(async move {
+            let _permit = permit; // hold permit until future completes
+            fut.await;
+        }),
+        // semaphore has been closed, return an empty task and do nothing
+        None => tokio::spawn(async move {}),
+    }
+}
```
huitseeker: The issue I have with this compared to e.g. the [BoundedExecutor](https://github.com/diem/diem/blob/416fdd3b18d02b2a099917467910d5a6865f2fed/common/bounded-executor/src/lib.rs) is the silent failure, which does not allow retry/failure management policies on top.

`BoundedExecutor` offers an alternative to the forcible queueing, thanks for looking at it.

mystenmark:
> > > Can we add a gauge to track the number of concurrent tasks? That would be super useful.
> > 
> > 
> > I added a log for when we hit the limit and start waiting
> 
> I think a gauge has a ton of value on its own, which we can easily look at in a dashboard and see and visually correlate patterns. It's much faster for diagnosing problems of this kind in the future, and we can like alert on it (for example that it hit the max for longer than a certain amount of time)

Sounds reasonable, but I don't have time to do that right now - feel free to replace my log with a gauge if you'd like to!

network/src/primary.rs
huitseeker: It's a decorator around a tokio [`Channel`](https://github.com/hyperium/tonic/blob/9ea03c2c44ad6bf76e4141e6fb40870b9b30ab63/tonic/src/transport/channel/mod.rs#L45-L71), TL;DR: not heavyweight.

network/src/primary.rs
```diff
@@ -72,17 +75,20 @@ impl PrimaryNetwork {
...
+                        let message = message.clone();
+                        async move { client.send_message(message).await.map_err(Into::into) }
+                    })
+                    .map(|response| {
+                        response.expect("we retry forever so this shouldn't fail");
+                    }),
```
huitseeker, line 89: Taking notes for a further PR (and @asonnino | @bmwill's eyes): now we can start splitting this out between both callers that land here: `broadcast` (reliable sends we must make sure we hear back from) and `unreliable_(send|broadcast)` (which we don't).

The retry is really here to make sure we execute the future >= 1 in the case of the unreliable send, so we could envision a 2-lane system: the `broadcast`'s `send_message` latch on the semaphore directly as they do now, because their not getitng through would block the protocol, whereas the `unreliable...` would use `try_spawn` and `try_acquire_owned` in a general (mutualized) retry, so as to only queue when there are permits left on the semaphore.

huitseeker (approved):
This looks good to me, thanks!

network/src/lib.rs
mystenmark: Well, BoundedExecutor just unwraps the permit result to avoid this issue

PR #464, opened 2022-07-08
fix: fix entry.sh to derive validator/worker_id from hostname in k8s

allan-bailey:
Currently there is no way with k8s kustomize templates or manifests to know the member number of a stateful set.
However, the hostname does have a -# number appended that is sequential for the number of replicas.

### DIFF
--- a/Docker/entry.sh
+++ b/Docker/entry.sh
@@ -3,6 +3,14 @@
 # Capture stack trace
 export RUST_BACKTRACE=1
 
+if [ -z "$VALIDATOR_ID" -a "$KUBERNETES_PORT" ]; then
+    export VALIDATOR_ID=${HOSTNAME##*-}
+    # assuming that WORKER_ID isn't also set.
+    # currently they match the validator it is assigned to.
+    export WORKER_ID=${VALIDATOR_ID}
+fi
+
+
 # Environment variables to use on the script
 NODE_BIN="./bin/node"
 KEYS_PATH="/validators/validator-$VALIDATOR_ID/key.json"


Docker/entry.sh
```diff
@@ -3,6 +3,14 @@
 # Capture stack trace
 export RUST_BACKTRACE=1
 
+if [ -z "$VALIDATOR_ID" -a "$KUBERNETES_PORT" ]; then
+    export VALIDATOR_ID=${HOSTNAME##*-}
+    # assuming that WORKER_ID isn't also set.
+    # currently they match the validator it is assigned to.
+    export WORKER_ID=${VALIDATOR_ID}
```
huitseeker, line 10: You could instead (and more safely, i.e. without clobbering a preexisting `WORKER_ID`) make `VALIDATOR_ID` the default value for `WORKER_ID`:
```bash
WORKER_ID=${WORKER_ID:=$VALIDATOR_ID}
```

allan-bailey: Thanks! good catch.

Docker/entry.sh
```diff
@@ -3,6 +3,14 @@
 # Capture stack trace
 export RUST_BACKTRACE=1
 
+if [ -z "$VALIDATOR_ID" -a "$KUBERNETES_PORT" ]; then
+    export VALIDATOR_ID=${HOSTNAME##*-}
```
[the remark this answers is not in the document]
allan-bailey: I'm assuming that they IDs will be "unique" to the cluster and starting at 0.

huitseeker (approved):
lgtm!

akichidis (approved):

PR #465, opened 2022-07-08
[proposer] Early-signal rounds from valid certificates to the proposer

huitseeker:
## Context

 The core's `process_certificates` lets the proposer know about full sets of parents certificates at a round r and at an epoch e. This set is full in the sense that it contains enough (2f+1) parents for the proposer to advance its round to r+1 and then propose (`make_header`) at that round r+1.

## Issues

- the proposer will propose at a round beyond that of any round for which it has a full set of parents, which is good,
- but it will also propose at that round even if it has evidence that such a proposal has zero chance of succeeding (to certification), which is bad.

Indeed, when a node receives a (well-signed) certificate C at round r, it knows that no proposal at a strictly lesser round can succeed, since a majority (>= f+1) of honest validators (of the 2f+1 signers of C) is committed to vote at least at this round. Note that we do not need to retrieve the payload or antecedents of the certificate C to assert this.

This leads our validator to make useless proposals when it is not fully caught up with the rest of the network (it reaches a proposal deadline without full sets of parents for every round).

## The Fix

When we learn of a (cryptographically valid) certificate bearing round r, without waiting, we let the proposer know of the smallest round at which we could ever provide a set of parents usable for a proposal: r-1.

This tells the validator to not bother proposing strictly below round r.

### DIFF
--- a/primary/src/core.rs
+++ b/primary/src/core.rs
@@ -339,6 +339,19 @@ impl<PublicKey: VerifyingKey> Core<PublicKey> {
     async fn process_certificate(&mut self, certificate: Certificate<PublicKey>) -> DagResult<()> {
         debug!("Processing {:?}", certificate);
 
+        // Let the proposer draw early conclusions from a certificate at this round and epoch, without its
+        // parents or payload (which we may not have yet).
+        //
+        // Since our certificate is well-signed, it shows a majority of honest signers stand at round r,
+        // so to make a successful proposal, our proposer must use parents at least at round r-1.
+        //
+        // This allows the proposer not to fire proposals at rounds strictly below the certificate we witnessed.
+        let minimal_round_for_parents = certificate.round().saturating_sub(1);
+        self.tx_proposer
+            .send((vec![], minimal_round_for_parents, certificate.epoch()))
+            .await
+            .map_err(|_| DagError::ShuttingDown)?;
+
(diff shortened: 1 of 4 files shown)


primary/src/core.rs
```diff
@@ -339,6 +339,19 @@ impl<PublicKey: VerifyingKey> Core<PublicKey> {
...
         debug!("Processing {:?}", certificate);
 
+        // Let the proposer draw early conclusions from a certificate at this round and epoch, without its
+        // parents or payload (which we may not have yet).
+        //
+        // Since our certificate is well-signed, it shows a majority of honest signers stand at round r,
+        // so to make a successful proposal, our proposer must use parents at least at round r-1.
+        //
+        // This allows the proposer not to fire proposals at rounds strictly below the certificate we witnessed.
+        let minimal_round_for_parents = certificate.round().saturating_sub(1);
```
LefKok, line 349: what does saturating_sub do?

huitseeker: `certificate.round().checked_sub(1).unwrap_or(0)`
Or informally x ↦max(x-1, 0)

LefKok (approved):

akichidis (approved):
Thanks a lot @huitseeker for this, makes total sense!

PR #472, opened 2022-07-11
[metrics] add metrics to monitor concurrent tasks in network

akichidis:
Following the work on https://github.com/MystenLabs/narwhal/pull/463 I believe that @velvia made a good point on adding metrics to monitor when we reach the maximum of capacity on concurrent tasks. Took the opportunity to add those since we are still hot on adding metrics and keep them until we stabilise things. Once we gain strong confidence on our nodes and we consider this an implementation detail we can probably remove.

### DIFF
--- a/network/src/metrics.rs
+++ b/network/src/metrics.rs
@@ -1,0 +1,155 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use prometheus::{default_registry, register_int_gauge_vec_with_registry, IntGaugeVec, Registry};
+use std::sync::Arc;
+
+pub trait NetworkMetrics {
+    fn network_available_tasks(&self) -> &IntGaugeVec;
+}
+
+#[derive(Clone, Debug)]
+pub struct PrimaryNetworkMetrics {
+    /// The number of executor available tasks
+    pub network_available_tasks: IntGaugeVec,
+}
+
+impl PrimaryNetworkMetrics {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            network_available_tasks: register_int_gauge_vec_with_registry!(
+                "primary_network_available_tasks",
+                "The number of available tasks to run in the network connector",
+                &["module", "network", "address"],
+                registry
+            )
+            .unwrap(),
+        }
+    }
+}
+
+impl NetworkMetrics for PrimaryNetworkMetrics {
+    fn network_available_tasks(&self) -> &IntGaugeVec {
+        &self.network_available_tasks
+    }
+}
+
+impl Default for PrimaryNetworkMetrics {
+    fn default() -> Self {
+        Self::new(default_registry())
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct WorkerNetworkMetrics {
+    /// The number of executor available tasks
+    pub network_available_tasks: IntGaugeVec,
+}
+
+impl WorkerNetworkMetrics {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            network_available_tasks: register_int_gauge_vec_with_registry!(
+                "worker_network_concurrent_tasks",
+                "The number of available tasks to run in the network connector",
+                &["module", "network", "address"],
+                registry
+            )
+            .unwrap(),
+        }
+    }
+}
+
+impl NetworkMetrics for WorkerNetworkMetrics {
+    fn network_available_tasks(&self) -> &IntGaugeVec {
+        &self.network_available_tasks
+    }
+}
+
+impl Default for WorkerNetworkMetrics {
+    fn default() -> Self {
+        Self::new(default_registry())
+    }
+}
+
+pub struct Metrics<N: NetworkMetrics> {
+    /// The handler to report the metrics.
+    metrics_handler: Arc<N>,
+    /// A tag used for every reported metric to trace the module where this has been used from
+    module_tag: String,
+    /// The network type - this shouldn't be set outside of this crate but is meant to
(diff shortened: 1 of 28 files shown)


network/src/metrics.rs
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use prometheus::{default_registry, register_int_gauge_vec_with_registry, IntGaugeVec, Registry};
+use std::sync::Arc;
+
+pub trait NetworkMetrics {
```
akichidis, line 7: Ideally I would use only one struct to report the metrics both for worker and primary nodes as those would normally be deployed as separate nodes. However, currently the worker & primary are deployed as part of the same binary in SUI and this would lead to re-registering the same metric (which would lead to an error). So had to split those but to minimise the implementation I used a common interface .

huitseeker (commented):
I'd be OK with the theory of what we're doing here (keeping a finger on the pulse of concurrency), but I'm wary of the *very extensive* code changes this is introducing and actually concerned we may never manage to either remove those metrics or maintain them well.

1. is there another tack we could take to reduce the maintenance burden, using e.g. the recently introduced tokio-metrics? https://tokio.rs/blog/2022-02-announcing-tokio-metrics
2. if not, is there a reduced, simpler variant of the metrics reporting that would e.g. wake up a background task every second and report the concurrency value in the bounded executors of each of the worker & primary? this would not require instrumentation of the whole code base I feel.

velvia:
@huitseeker @akichidis just had a look at this.  Here's my take:

I think in terms of manually maintaining our own metrics -- most of the change in this PR really has to do with adding an inner Metrics struct to each network facility.  My guess is that this is going to be hard to avoid with this approach.
I had a thought where in the `BoundedExecutor` itself where it is acquiring and releasing we could possibly update the gauge within there, which would only help a little bit.

Is there a higher level place which could collect this metric periodically?  Then each network eg primary, etc only has to expose a method, and the higher level thing could call down to get the metric.  This would be less intrusive in that one would not need to pass in a separate Metrics struct just for monitoring this one thing.

As for Tokio Metrics, I think that is a good approach, though that requires some investment too.  That should return much more metrics overall and be really good and rich, but that will likely be a few days.

akichidis:
@huitseeker @velvia thanks both for your comments. So my thoughts regarding the above:

1. Tokio metrics they seem to be super useful and something we can/should evaluate. From a quick look though they don't seem to offer a way to inject customer metrics, but rather report a series of standard ones (more info [here](https://github.com/tokio-rs/tokio-metrics#task-metrics)). Here we want to monitor custom values which don't fit in the realm of task / runtime metrics - or at least in the obvious way we need them
2. @huitseeker your idea around having a separate task to periodically report the values sounds definitely a way for us to avoid injecting all this "inner" code to the handlers. The thing is that in the end we'll need to build and maintain this mechanism which to be honest I don't know how much less hassle will be compared to what we currently have. We'll need pretty much the tagging I've introduced (a bounded executor needs to be identified by the corresponding networking handler) and we'll definitely need a way to "push" the values to this background collector. I mean now it seems that this PR is introducing many changes, but keep in mind that all the boilerplate now is there. If want to add additional metrics this will be only a few lines of change.

That being said I am open to give a quick try on the approach (2) mentioned above and compare the complexity

network/src/metrics.rs
```diff
@@ -0,0 +1,143 @@
...
+impl PrimaryNetworkMetrics {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            network_concurrent_tasks: register_int_gauge_vec_with_registry!(
+                "primary_network_concurrent_tasks",
+                "The number of concurrent tasks running in the network connector",
+                &["module", "network"],
+                registry
+            )
+            .unwrap(),
```
huitseeker, line 26: I'd like to not panic the node if this fails: metrics are not enough to justify a crash, IMHO.

network/src/metrics.rs
```diff
@@ -0,0 +1,143 @@
...
+impl WorkerNetworkMetrics {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            network_concurrent_tasks: register_int_gauge_vec_with_registry!(
+                "worker_network_concurrent_tasks",
+                "The number of concurrent tasks running in the network connector",
+                &["module", "network"],
+                registry
+            )
+            .unwrap(),
```
huitseeker, line 58: see above re: the avoidable panic

huitseeker (commented):
OK, after the experience of #559 I recognize how wise this is and I am now in favor of introducing this PR with minor changes. 
I think this mostly needs a rebase.

network/src/metrics.rs
akichidis: Ah that should happen primarily if [we have already registered the same metric ](https://docs.rs/prometheus/latest/src/prometheus/registry.rs.html#257)in the codebase somewhere else. Having this there will protect us from accidentally declaring the same metric in two different places and go unnoticed - which I believe is quite important. So my expectation is that this will be caught early in our e2e tests before it even hits staging - or production worst case.

network/src/metrics.rs
akichidis: Similar to this https://github.com/MystenLabs/narwhal/pull/472#discussion_r928783850

huitseeker:
@akichidis you may need to rebase this one to pass checks: we changed a linter in #485

akichidis:
@huitseeker @asonnino @bmwill @velvia I've rebased the PR in the latest master, if you could please re-review

huitseeker (approved):
Thank you very much 🎉

Issue #115, opened 2022-03-29
Consensus crash-recovery


asonnino:
### Description
To ensure crash-recovery of consensus, it is not enough to have the `CertificatesWaiter` to send to consensus every certificate for which their ancestors are in the storage. After crash-recovery, we may also need to send all parents certificates with rounds greater than `last_committed`. This means we need to boot the `CertificatesWaiter` with knowledge of the last committed round and modify its workings to potentially re-serve some past certificates as well.

Note that the Primary core also need to persist `last_voted` to not equivocate on certificates.

huitseeker:
See also #17.

asonnino:
fixed by https://github.com/MystenLabs/narwhal/pull/481

gdanezis:
I do not think this is entirely done. As per the discussion with @akichidis - the interface between client <-> Bullshark has to be adapted to ensure upon recovery the stream of agreed messages is correct. See also (new) issue #730

asonnino:
You are right! sorry for that

PR #481, opened 2022-07-12
construct ConsensusState from ConsensusStore

laura-makdah:
Part of https://github.com/mystenlabs/narwhal/issues/115

It looks like we are already reloading the contents of the ConsensusStore from disk on startup, but not using it.

Left to do: reconstruct the ``` Dag<PublicKey> =
    HashMap<Round, HashMap<PublicKey, (CertificateDigest, Certificate<PublicKey>)>>;```
One way I see to do this is to also store this struct on disk and reload it, however, there may be a possibility to instead reconstruct it from the existing structs that are written / loaded from disk?
``` pub header_store: Store<HeaderDigest, Header<PublicKey>>,
    pub certificate_store: Store<CertificateDigest, Certificate<PublicKey>>,
    pub payload_store: Store<(BatchDigest, WorkerId), PayloadToken>,
    pub batch_store: Store<BatchDigest, SerializedBatchMessage>,
    pub consensus_store: Arc<ConsensusStore<PublicKey>>,
```

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -52,6 +52,36 @@ impl<PublicKey: VerifyingKey> ConsensusState<PublicKey> {
         }
     }
 
+    pub fn new_from_store(
+        genesis: Vec<Certificate<PublicKey>>,
+        metrics: Arc<ConsensusMetrics>,
+        last_committed: HashMap<PublicKey, Round>,
+    ) -> Self {
+        let last_committed_round = *last_committed
+            .iter()
+            .max_by(|a, b| a.1.cmp(b.1))
+            .map(|(_k, v)| v)
+            .unwrap_or_else(|| &0);
+        if last_committed_round == 0 {
+            return Self::new(genesis, metrics);
+        }
+
+        let genesis = genesis
+            .into_iter()
+            .map(|x| (x.origin(), (x.digest(), x)))
+            .collect::<HashMap<_, _>>();
+
+        Self {
+            last_committed_round,
+            last_committed,
+            dag: [(0, genesis)]
+                .iter()
+                .cloned()
+                .collect::<HashMap<_, HashMap<_, _>>>(), // todo Laura: reconstruct using store
+            metrics,
+        }
+    }
+
     /// Update and clean up internal state base on committed certificates.
     pub fn update(&mut self, certificate: &Certificate<PublicKey>, gc_depth: Round) {
         self.last_committed
@@ -119,6 +149,9 @@ pub struct Consensus<PublicKey: VerifyingKey, ConsensusProtocol> {
 
     /// Metrics handler
     metrics: Arc<ConsensusMetrics>,
+
+    /// The latest committed round of each validator.
+    last_committed: HashMap<PublicKey, Round>,
 }
 
 impl<PublicKey, Protocol> Consensus<PublicKey, Protocol>
@@ -137,6 +170,7 @@ where
     ) -> JoinHandle<StoreResult<()>> {
         tokio::spawn(async move {
             let consensus_index = store.read_last_consensus_index()?;
+            let last_committed = store.read_last_committed();
             let committee: &Committee<PublicKey> = &committee.load();
             let genesis = Certificate::genesis(committee);
             Self {
@@ -147,6 +181,7 @@ where
                 consensus_index,
                 protocol,
                 metrics,
+                last_committed,
             }
             .run()
             .await
@@ -155,7 +190,11 @@ where
 
     async fn run(&mut self) -> StoreResult<()> {
         // The consensus state (everything else is immutable).
-        let mut state = ConsensusState::new(self.genesis.clone(), self.metrics.clone());
+        let mut state = ConsensusState::new_from_store(
+            self.genesis.clone(),
+            self.metrics.clone(),
+            self.last_committed.clone(),
+        );
 
         // Listen to incoming certificates.
         while let Some(certificate) = self.rx_primary.recv().await {


consensus/src/consensus.rs
```diff
@@ -137,8 +198,10 @@ where
     ) -> JoinHandle<StoreResult<()>> {
         tokio::spawn(async move {
             let consensus_index = store.read_last_consensus_index()?;
+            let last_committed = store.read_last_committed();
```
huitseeker, line 201: I would rename `last_committed` to something that makes it explicit this comes from a store, e.g. `recovered_last_committed`, here, and in all the places this is used. 

The soundness of your reconstructed `ConsensusState` only works if this table has been reconstructed from disk, so I think it's wise to use the name to invite the reader of `construct_dag_from_cert_store` to go read the very different place where this value is computed.

consensus/src/consensus.rs
```diff
@@ -119,6 +173,12 @@ pub struct Consensus<PublicKey: VerifyingKey, ConsensusProtocol> {
 
     /// Metrics handler
     metrics: Arc<ConsensusMetrics>,
+
+    /// The latest committed round of each validator.
+    last_committed: HashMap<PublicKey, Round>,
+
+    /// Certificates to load into the DAG
+    certs: HashMap<CertificateDigest, Certificate<PublicKey>>,
```
huitseeker, line 181: I'm confused: you seem to read those from the certificate store when constructing `ConsensusState` in the first place, why do you need those here as well?

huitseeker (changes requested):
You're reading certificates in memory and putting some of them in the `dag` `HashMap` when creating `ConsensusState`, and this makes sense to me.

It seems you're also reading them in memory and carrying those certificates in the `Consensus` struct. Why is that?

consensus/src/consensus.rs
```diff
@@ -52,6 +53,59 @@ impl<PublicKey: VerifyingKey> ConsensusState<PublicKey> {
...
+                let inner = dag.get_mut(&cert.header.round);
+                match inner {
+                    Some(m) => {
+                        m.insert(cert.header.author.clone(), (*digest, cert.clone()));
+                    }
+                    None => {
+                        dag.insert(cert.header.round, HashMap::new());
+                        dag.get_mut(&cert.header.round)
+                            .unwrap()
+                            .insert(cert.header.author.clone(), (*digest, cert.clone()));
```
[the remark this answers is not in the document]
laura-makdah: 👍

consensus/src/consensus.rs
laura-makdah: removing this

consensus/src/consensus.rs
```diff
@@ -155,7 +220,13 @@ where
 
     async fn run(&mut self) -> StoreResult<()> {
         // The consensus state (everything else is immutable).
-        let mut state = ConsensusState::new(self.genesis.clone(), self.metrics.clone());
+        let mut state = ConsensusState::new_from_store(
+            self.genesis.clone(),
+            self.metrics.clone(),
+            self.last_committed.clone(),
+            self.certs.clone(),
+            50, // todo Laura: get parameter from bullshark / tusk
```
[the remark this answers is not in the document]
laura-makdah: 👍

consensus/src/tests/bullshark_tests.rs
```diff
@@ -75,20 +75,23 @@ async fn commit_one() {
...
         committee: Arc::new(ArcSwap::from_pointee(mock_committee(&keys[..]))),
         store: store.clone(),
-        gc_depth: 50,
+        gc_depth,
     };
     let metrics = Arc::new(ConsensusMetrics::new(&Registry::new()));
     Consensus::spawn(
         Arc::new(ArcSwap::from_pointee(mock_committee(&keys[..]))),
         store,
+        make_certificate_store(&test_utils::temp_dir()),
```
huitseeker, line 88: Nit: it would be nice, here and in other tests, to follow the pattern of `consensus_store` and declare an intermediate variable for this one above, just to see the stores setup beside each other.

consensus/src/consensus.rs
```diff
@@ -129,14 +188,17 @@ where
...
         rx_primary: Receiver<Certificate<PublicKey>>,
         tx_primary: Sender<ConsensusPrimaryMessage<PublicKey>>,
         tx_output: Sender<ConsensusOutput<PublicKey>>,
         protocol: Protocol,
         metrics: Arc<ConsensusMetrics>,
+        gc_depth: Round,
     ) -> JoinHandle<StoreResult<()>> {
         tokio::spawn(async move {
             let consensus_index = store.read_last_consensus_index()?;
+            let recovered_last_committed = store.read_last_committed();
```
huitseeker, line 201: Nit: while the name `recovered_last_committed` makes sense to me, `recover_last_committed` less so.

node/src/lib.rs
```diff
@@ -199,11 +199,13 @@ impl Node {
         Consensus::spawn(
             committee.clone(),
             store.consensus_store.clone(),
+            store.certificate_store.clone(),
             /* rx_primary */ rx_new_certificates,
             /* tx_primary */ tx_feedback,
             /* tx_output */ tx_sequence,
             ordering_engine,
             consensus_metrics.clone(),
+            parameters.gc_depth,
```
huitseeker, line 208: 👍

huitseeker (approved):
This looks rather splendid to me! Thanks for this & the fast turnaround!

I've left a couple of nits you can perfectly ignore.

consensus/src/consensus.rs
```diff
@@ -52,6 +53,68 @@ impl<PublicKey: VerifyingKey> ConsensusState<PublicKey> {
...
+        }
+    }
+
+    pub async fn construct_dag_from_cert_store(
+        cert_store: Store<CertificateDigest, Certificate<PublicKey>>,
+        last_committed_round: Round,
+        gc_depth: Round,
+    ) -> Dag<PublicKey> {
+        let mut dag: Dag<PublicKey> = HashMap::new();
+        let min_round = last_committed_round - gc_depth;
```
asonnino, line 94: Can this underflow?

akichidis: Very good eye! Indeed is possible and this has been fixed here https://github.com/MystenLabs/narwhal/pull/497

PR #484, opened 2022-07-12
crypto: Implement KeyPair for ECDSA secp256k1

joyqvq:
https://github.com/MystenLabs/narwhal/issues/483

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -15,7 +15,7 @@ hex = "0.4.3"
 rand = { version = "0.7.3", features = ["std"] }
 serde = { version = "1.0.139", features = ["derive"] }
 zeroize = "1.5.6"
-signature = "1.5.0"
+signature = "1.2.0"
 tokio = { version = "1.20.0", features = ["sync", "rt", "macros"] }
 ark-bls12-377 = { version = "0.3.0", features = ["std"], optional = true }
 hkdf = { version = "0.12.3", features = ["std"] }
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -22,6 +22,7 @@ hkdf = { version = "0.12.3", features = ["std"] }
 serde_with = "1.14.0"
 serde_json = "1.0.80"
 schemars ="0.8.10"
+k256 = { version = "0.7.2", features = ["ecdsa", "sha256"] }
 
 # TODO: switch to https://github.com/celo-org/celo-bls-snark-rs 
 # when https://github.com/celo-org/celo-bls-snark-rs/issues/228 is solved
--- a/crypto/benches/crypto.rs
+++ b/crypto/benches/crypto.rs
@@ -13,6 +13,7 @@ mod ed25519_benches {
         bls12377::{BLS12377KeyPair, BLS12377Signature},
         bls12381::{BLS12381KeyPair, BLS12381Signature},
         ed25519::*,
+        secp256k1::{Secp256k1KeyPair, Secp256k1Signature},
         traits::{KeyPair, VerifyingKey},
         Verifier,
     };
@@ -24,6 +25,7 @@ mod ed25519_benches {
         let ed_keypair = Ed25519KeyPair::generate(&mut csprng);
         let bls_keypair = BLS12377KeyPair::generate(&mut csprng);
         let blst_keypair = BLS12381KeyPair::generate(&mut csprng);
+        let secp256k1_keypair = Secp256k1KeyPair::generate(&mut csprng);
         let msg: &[u8] = b"";
 
         c.bench_function("Ed25519 signing", move |b| b.iter(|| ed_keypair.sign(msg)));
@@ -33,6 +35,9 @@ mod ed25519_benches {
         c.bench_function("BLS12381 signing", move |b| {
             b.iter(|| blst_keypair.sign(msg))
         });
+        c.bench_function("Secp256k1 signing", move |b| {
+            b.iter(|| secp256k1_keypair.sign(msg))
+        });
     }
 
     fn verify(c: &mut Criterion) {
@@ -40,15 +45,18 @@ mod ed25519_benches {
         let ed_keypair = Ed25519KeyPair::generate(&mut csprng);
         let bls_keypair = BLS12377KeyPair::generate(&mut csprng);
         let blst_keypair = BLS12381KeyPair::generate(&mut csprng);
+        let secp256k1_keypair = Secp256k1KeyPair::generate(&mut csprng);
 
         let ed_public = ed_keypair.public();
         let bls_public = bls_keypair.public();
         let blst_public = blst_keypair.public();
+        let secp256k1_public = secp256k1_keypair.public();
 
         let msg: &[u8] = b"";
         let ed_sig: Ed25519Signature = ed_keypair.sign(msg);
         let bls_sig: BLS12377Signature = bls_keypair.sign(msg);
         let blst_sig: BLS12381Signature = blst_keypair.sign(msg);
+        let secp256k1_sig: Secp256k1Signature = secp256k1_keypair.sign(msg);
 
         c.bench_function("Ed25519 signature verification", move |b| {
             b.iter(|| ed_public.verify(msg, &ed_sig))
@@ -58,6 +66,9 @@ mod ed25519_benches {
         });
         c.bench_function("BLS12381 signature verification", move |b| {
             b.iter(|| blst_public.verify(msg, &blst_sig))
+        });
+        c.bench_function("Secp256k1 signature verification", move |b| {
+            b.iter(|| secp256k1_public.verify(msg, &secp256k1_sig))
         });
     }
 
@@ -133,6 +144,9 @@ mod ed25519_benches {
         });
         c.bench_function("BLS12381 keypair generation", move |b| {
             b.iter(|| BLS12381KeyPair::generate(&mut csprng))
+        });
+        c.bench_function("Secp256k1 keypair generation", move |b| {
+            b.iter(|| Secp256k1KeyPair::generate(&mut csprng))
         });
     }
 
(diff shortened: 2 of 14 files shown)


crypto/src/ecdsa.rs
```diff
@@ -0,0 +1,305 @@
...
+use signature::{Signature, Signer, Verifier};
+use std::fmt::{self, Debug, Display};
+
+pub const ECDSA_PRIVATE_KEY_LENGTH: usize = 32;
+pub const ECDSA_PUBLIC_KEY_LENGTH: usize = 33;
+pub const ECDSA_SIGNATURE_LENGTH: usize = 64;
+
+#[readonly::make]
+#[derive(Debug, Clone)]
+pub struct EcdsaPublicKey {
```
punwai, line 20: I'd probably be more specific about the curve that we are using - perhaps Francois has more insight into that.

punwai: If I'm not mistaken, we might be looking to implement secp256r1 as well.

crypto/src/tests/secp256k1_tests.rs
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2021, Facebook, Inc. and its affiliates
```
huitseeker, line 1: I think this one is not necessary.

crypto/src/tests/secp256k1_tests.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    let signature = kp.sign(&digest.0);
+
+    // Verify the signature.
+    let bad_message: &[u8] = b"Bad message!";
+    let digest = bad_message.digest();
+
+    assert!(kp.public().verify(&digest.0, &signature).is_err());
+}
+
```
huitseeker, line 96: It would be great to test the main cryptographic functions using [secp256k1](https://docs.rs/secp256k1/latest/secp256k1/), the rust bindings to the Bitcoin lib. One way to engineer those tests is with [proptest](https://docs.rs/proptest/latest/proptest/index.html) which will generate random inputs, let you define how to generate a random key with each lib, then test the comparisons.
The nice thing about this being that secp256k1 only appears in dev-dependencies of this package, and that it's probably not hard to compare that implementation to secp256k1 in micro-benchmarks.

crypto/benches/crypto.rs
```diff
@@ -13,6 +13,7 @@ mod ed25519_benches {
         bls12377::{BLS12377KeyPair, BLS12377Signature},
         bls12381::{BLS12381KeyPair, BLS12381Signature},
         ed25519::*,
+        secp256k1::{Secp256k1KeyPair, Secp256k1Signature},
```
huitseeker, line 16: 🎉

huitseeker (commented):
I think this is implementing agains the old traits and will need a rebase to take on the changes in #460 (and lead to a test in Sui). This globally looks good, I've left a few suggestions.

crypto/benches/crypto.rs
```diff
@@ -134,6 +145,9 @@ mod ed25519_benches {
         c.bench_function("BLS12381 keypair generation", move |b| {
             b.iter(|| BLS12381KeyPair::generate(&mut csprng))
         });
+        c.bench_function("Secp256k1 keypair generation", move |b| {
+            b.iter(|| Secp256k1KeyPair::generate(&mut csprng))
+        });
     }
 
     criterion_group! {
```
huitseeker, line 153: Not that this is part of your PR, but this criterion group needs a better name 😁

crypto/src/serde_helpers.rs
```diff
@@ -78,6 +78,6 @@ impl<'de> DeserializeAs<'de, ed25519_dalek::Signature> for Ed25519Signature {
         } else {
             Bytes::deserialize_as(deserializer)?
         };
-        ed25519_dalek::Signature::from_bytes(&bytes).map_err(to_custom_error::<'de, D, _>)
+        <ed25519_dalek::Signature as signature::Signature>::from_bytes(&bytes).map_err(to_custom_error::<'de, D, _>)
```
huitseeker, line 81: Same remark on the `Signature` import.

crypto/src/secp256k1.rs
```diff
@@ -0,0 +1,353 @@
...
+
+pub const SECP256K1_PRIVATE_KEY_LENGTH: usize = 32;
+pub const SECP256K1_PUBLIC_KEY_LENGTH: usize = 33;
+pub const SECP256K1_SIGNATURE_LENGTH: usize = 64;
+
+#[readonly::make]
+#[derive(Debug, Clone)]
+pub struct Secp256k1PublicKey {
+    pub pubkey: k256::ecdsa::VerifyingKey,
+    pub bytes: OnceCell<[u8; SECP256K1_PUBLIC_KEY_LENGTH]>,
```
huitseeker, line 21: I wonder if this is actually useful. Is there a way to implement `AsRef<[u8]>` for `k256::ecdsa::VerifyingKey`? If so, we would not need this `OnceCell`

crypto/src/tests/secp256k1_tests.rs
```diff
@@ -0,0 +1,135 @@
...
+    
+        let signature = kp.sign(&digest.0);
+        let secp = rust_secp256k1::Secp256k1::new();
+        // let s = secp.
+        // let (secret_key, public_key) = secp.ge
+// ! let message = Message::from_hashed_data::<sha256::Hash>("Hello World!".as_bytes());
+// !
+// ! let sig = secret_key.sign_ecdsa(&message, &secret_key);
+// ! assert!(sig.verify(&message, &public_key).is_ok());
+// !
```
huitseeker, line 98: Looking forward to seeing more of this !

huitseeker (commented):
This is going in an exciting direction 🦾

crypto/benches/crypto.rs
kchalkias: Yes let's do in an upcoming cleanup PR

kchalkias (commented):
Great work! Let's fix the most basic comments and we reiterate on the rest (ie extensive tests and obvious cleanups) in upcoming PRs. Ah please ensure you run `cargo fmt` and `clippy` before pushing.

There are also extra features we should consider in the final deliverable, like ecrecover, ECDSA-NM (non-malleable), but again not in this PR.

crypto/Cargo.toml
```diff
@@ -15,13 +15,14 @@ hex = "0.4.3"
 rand = { version = "0.7.3", features = ["std"] }
 serde = { version = "1.0.139", features = ["derive"] }
 zeroize = "1.5.6"
-signature = "1.5.0"
```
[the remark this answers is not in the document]
joyqvq: in order to implement this function `fn generate<R: CryptoRng + RngCore>(rng: &mut R) -> Self;
` in the trait KeyPair, all implementations of it (ed25519 and secp256k1) needs to be on the same RngCore version. due to ed255-dalek being capped at RngCore 0.5.1, we need to use a version of k256 that also depends on rand-core 0.5.1. this lower version of k256 resulted in a lower version of `signature` it depends on. 

hopefully, once we get rid of dalek, this will be an non-issue for us to use the latest rand-core

crypto/src/tests/secp256k1_tests.rs
joyqvq: almost had it working, but due to the different representations of a signature in the c library, comparing the actual signature result was a bit challenging to massage them into. 

will follow up with the prop test PR

crypto/src/secp256k1.rs
joyqvq: > I wonder if this is actually useful. Is there a way to implement `AsRef<[u8]>` for `k256::ecdsa::VerifyingKey`? If so, we would not need this `OnceCell`

there wasnt anyway to get the underlying bytes out of `k256::ecdsa::VerifyingKey` (its a private struct and no interface to retrieve this).

closest I can find, but this will not be AsRef
```
    /// Serialize this [`VerifyingKey`] as a SEC1-encoded bytestring
    /// (with point compression applied)
    pub fn to_bytes(&self) -> CompressedPoint {
        let mut result = [0u8; 33];
        result.copy_from_slice(EncodedPoint::from(self).as_ref());
        result
    }
```

crypto/src/traits.rs
```diff
@@ -134,7 +134,6 @@ pub trait Authenticator:
 {
     type PubKey: VerifyingKey<Sig = Self>;
     type PrivKey: SigningKey<Sig = Self>;
-    type AggregateSig: AggregateAuthenticator<Sig = Self>;
```
joyqvq, line 137: since BLS12377 and Ed25519 has `impl AggregateAuthenticator for` and its needed functionality, and this type is not applicable to Ecdsa for now, removing this from the Authenticator trait which means to be the common denominator for all key types

kchalkias: Maybe we shouldn't remove existing trait functionality (unless we're 100% it's not used elsewhere - for which we don't know) as this might be a breaking change elsewhere. 

We could just implement a "concatenate" sigs for ECDSA (as we do with ed25519) and naively verify (in a for-loop) all sigs; because we don't have batch_verify for ECDSA.

crypto/Cargo.toml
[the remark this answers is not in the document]
huitseeker: Please open the issue: we use the `signature` trait is used in `Sui` as well!
Note also that some crates re-export the rand crate : is this the case for `secp256k1`? You could leave this and pass in a `secp256k1::rand::rng::OsRng` or some such?

crypto/src/secp256k1.rs
huitseeker: Entirely fair, then, to use the memo pattern.Thanks for looking into this.

crypto/src/traits.rs
huitseeker: We're 100% used this isn't used anywhere, this makes sense to me. The `AggregateAuthenticator` already points to this in its associated types anyway. We would only need this if we decided to use ECDSA for validator sigs.

huitseeker (approved):
Thanks a lot for this ! This LGTM, please open an issue for proptests if you intend to complete them later.

PR #485, opened 2022-07-12
[Lint] Run clang-format for protobuf

mwtian:
There seems to be no formatting for protobuf right now, so add them to the CI workflow.

There are some existing lint issues, e.g. proto enums should start with a `UNSPECIFIED` / `NOT_SET` tag at 0. This can be fixed in a subsequent PR if people agree with them. Then proto linting can be enabled.

Also, switch to [`super-linter`](https://github.com/github/super-linter) which has support for all existing linters, and can hopefully reduce effort for adding linters in future.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,5 @@
 # For proto file format
 BasedOnStyle: Google
 IndentWidth: 4
+ColumnLimit: 100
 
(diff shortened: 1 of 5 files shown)


huitseeker (approved):
Thanks @mwtian ! Sincere apologies for the late review, this is a very nice contribution!

mwtian:
No problem, and thanks!

PR #486, opened 2022-07-13
Fix panic in process_header, test CertificateWaiter

huitseeker:
The first commit is purely technical: replaces an `mpsc::channel` by a lighter `oneshot::channel` for waiter cancellations.

The second (see message) introduces a long test firing up the `CertificateWaiter` and all dependent components,
and feeding it valid certificates in inverse (decreasing) round order, simulating catching up as a late node.

Along the way, this *discovers a panic in this catch-up path*:
- `process_certificate` calls process_header,
- `process_header` can return with `Ok(())` in case the node has missing parents / payload,
- but in the case it doesn't, we will try to vote for the header if not in a `last_voted` table,
- that vote may be a panic-inducing duplicate vote if
    + the node was already among the voters for this certificate,
    + a restart has blown the node's in-memory `last_voted` table,
 - this panic blows up all the way to the process,
- we change this panic to instead generate the following message at the error level:
```
2022-07-13T12:36:25.607333Z ERROR process_header: primary::core: Failed to process our own vote: Authority vq2gYSbHjZi0oaafbuYYlpTw9HUVONqCTxrcixShtWI= appears in quorum more than once
```

The behavior-changing fix is ca. lines 275 of `primary/src/core.rs`

This also tweaks a few mock functions to generate those valid certificates and headers in the first place.

### DIFF
--- a/primary/src/certificate_waiter.rs
+++ b/primary/src/certificate_waiter.rs
@@ -29,6 +29,10 @@ use types::{
     Certificate, CertificateDigest, HeaderDigest, Reconfigure, Round,
 };
 
+#[cfg(test)]
+#[path = "tests/certificate_waiter_tests.rs"]
+pub mod certificate_waiter_tests;
+
 /// Waits to receive all the ancestors of a certificate before looping it back to the `Core`
 /// for further processing.
 pub struct CertificateWaiter<PublicKey: VerifyingKey> {
--- a/primary/src/certificate_waiter.rs
+++ b/primary/src/certificate_waiter.rs
@@ -18,8 +18,8 @@ use std::{
 use store::Store;
 use tokio::{
     sync::{
-        mpsc::{channel, Receiver, Sender},
-        watch,
+        mpsc::{Receiver, Sender},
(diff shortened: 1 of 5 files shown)


asonnino:
Ha I finally understand what you tried to tell me yesterday!

Just to be sure, when you say the following:
> that vote may be a panic-inducing duplicate vote if
> * the node was already among the voters for this certificate,
> * a restart has blown the node's in-memory last_voted table,

You mean that the panic happens when both conditions are trigger (not only one of them)? If so, the panic was expected. We have to persist `last_voted` or we may loose safety. Narwhal looses many of its properties if Byzantine nodes can trick crash-recovering nodes to equivocate on certificates (ie. vote for two different headers from the same (author, round))

asonnino (approved):
I agree that the PR doesn't introduce any additional safety issues. Let's just remember to convert back the Result into a panic once we can safely crash-recover

huitseeker:
Amended to mention changes which need to happen after https://github.com/MystenLabs/narwhal/issues/488

PR #487, opened 2022-07-13
[testing] introduce the Cluster struct & integration test for causal completion

akichidis:
This PR is introducing:
* The `Cluster` struct to allow us easily bring up a cluster of primary nodes (up to 4 for now) , stop single nodes, start nodes independently (so we can simulate late comers) etc.
* Allow confirm the nodes internals by using the prometheus Registries (setup individual ones by each node)
* An integration test that showcases the issue of the causal completion and the `Failed to process our own vote` error that has been presented on the https://github.com/MystenLabs/narwhal/pull/486 as well.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -10,20 +10,17 @@
 
 use anyhow::{Context, Result};
 use arc_swap::ArcSwap;
-use async_trait::async_trait;
 use clap::{crate_name, crate_version, App, AppSettings, ArgMatches, SubCommand};
 use config::{Committee, Import, Parameters, WorkerId};
 use crypto::{ed25519::Ed25519KeyPair, generate_production_keypair, traits::KeyPair};
-use executor::{
-    ExecutionIndices, ExecutionState, ExecutionStateError, SerializedTransaction, SubscriberResult,
-};
+use executor::{SerializedTransaction, SubscriberResult};
 use futures::future::join_all;
 use node::{
+    execution_state::SimpleExecutionState,
     metrics::{primary_metrics_registry, start_prometheus_server, worker_metrics_registry},
     Node, NodeStorage,
 };
 use std::sync::Arc;
-use thiserror::Error;
 use tokio::sync::mpsc::{channel, Receiver};
 use tracing::{info, subscriber::set_global_default};
 use tracing_subscriber::filter::{EnvFilter, LevelFilter};
--- a/primary/tests/epoch_change.rs
+++ b/primary/tests/epoch_change.rs
@@ -6,7 +6,7 @@ use crypto::traits::KeyPair;
 use futures::future::join_all;
 use node::NodeStorage;
 use primary::{NetworkModel, Primary, CHANNEL_CAPACITY};
-use prometheus::default_registry;
+use prometheus::Registry;
 use std::{collections::BTreeMap, sync::Arc};
 use test_utils::{keys, make_authority, pure_committee_from_keys, temp_dir};
 use tokio::sync::mpsc::channel;
@@ -51,7 +51,7 @@ async fn test_simple_epoch_change() {
             /* dag */ None,
             NetworkModel::Asynchronous,
             /* tx_committed_certificates */ tx_feedback,
-            default_registry(),
+            &Registry::new(),
         );
     }
 
@@ -138,7 +138,7 @@ async fn test_partial_committee_change() {
             /* dag */ None,
             NetworkModel::Asynchronous,
             /* tx_committed_certificates */ tx_feedback,
-            default_registry(),
+            &Registry::new(),
         );
     }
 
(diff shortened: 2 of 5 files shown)


node/src/execution_state.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
```
akichidis, line 1: Just moved this code out of the `main` file

primary/src/metrics.rs
```diff
@@ -16,35 +16,23 @@ pub(crate) struct Metrics {
     pub(crate) node_metrics: Option<PrimaryMetrics>,
 }
 
-static mut METRICS: Metrics = Metrics {
-    endpoint_metrics: None,
-    primary_endpoint_metrics: None,
-    node_metrics: None,
-};
-static INIT: Once = Once::new();
```
akichidis, line 24: No need to keep the metrics behind a `Once` - since we can pass a Registry we can always pass a separate Registry per test so we don't bump into conflict issues on the default registry

primary/tests/causal_completion_tests.rs
```diff
@@ -0,0 +1,98 @@
...
+    let mut cluster = Cluster::new(None);
+
+    // start the cluster
+    let nodes = cluster.start(4).await;
+
+    // Let primaries advance little bit
+    tokio::time::sleep(Duration::from_secs(10)).await;
+
+    // Ensure all nodes advanced
+    for node in nodes {
```
akichidis, line 26: I believe at the moment the metrics is the most powerful tool we have to understand what's going on with a node since we are lacking of any endpoints - at least to test the cases that have to do with node advancement, late joiners etc

test_utils/src/cluster.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        let mut k = keys(None);
+        let keypair = k.remove(id);
+        let name = keypair.public().clone();
+
+        let registry = primary_metrics_registry(name.clone());
+
+        // Make the data store.
+        let store: NodeStorage<Ed25519PublicKey> = NodeStorage::reopen(temp_dir());
```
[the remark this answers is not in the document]
laura-makdah: That is precisely the functionality we need for testing loading consensus from disk.

primary/tests/causal_completion_tests.rs
```diff
@@ -0,0 +1,98 @@
...
+        let metric_family = node.registry.gather();
+
+        for metric in metric_family {
```
laura-makdah, line 29: Before this point, don't we need to add certificates to the nodes for processing? Or rather, I think it would be helpful for testing more specific scenarios, but don't quite know how to do that.

test_utils/src/cluster.rs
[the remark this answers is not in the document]
akichidis: I have opened a follow up PR that allow us to preserve the disk between node restarts (reading from the same RocksDB folder) as currently we always create a new one we a node starts. I believe that would be useful for your tests https://github.com/MystenLabs/narwhal/pull/492

huitseeker (commented):
This seems like a good start! Two questions:
- we're reusing the `tracing_subscriber` manually, as we did before in this repo. Is this a good time to upgrade to [telemetry-subscribers](https://github.com/MystenLabs/mysten-infra/tree/main/crates/telemetry-subscribers) as Sui does, hopefully minimizing some maintenance burden? e.g. I can definitely see myself running those tests under tokio-console 
- the absence of a worker seems like a severe limitation given some workflows wait on worker interaction, how do we plan to address this component?

primary/tests/causal_completion_tests.rs
huitseeker: The test introduced here may include useful scaffolding: https://github.com/MystenLabs/narwhal/pull/486

test_utils/src/cluster.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        // The channel returning the result for each transaction's execution.
+        let (tx_transaction_confirmation, _) = channel(Node::CHANNEL_CAPACITY);
+
+        let h = Node::spawn_primary(
```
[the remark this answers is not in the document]
akichidis: On follow up PR I'll spin up worker(s) for each authority - they'll communicate via network as expected. The current test doesn't really need workers as we are testing against empty headers (they don't have payloads) where no contact is required to workers at any point. Very precisely as you said this will be an issue when we make tests where we pump in transactions - that's why I'll amend to spin up worker nodes.

akichidis:
Thanks a lot for the feedback @huitseeker .
> * we're reusing the `tracing_subscriber` manually, as we did before in this repo. Is this a good time to upgrade to [telemetry-subscribers](https://github.com/MystenLabs/mysten-infra/tree/main/crates/telemetry-subscribers) as Sui does, hopefully minimizing some maintenance burden? e.g. I can definitely see myself running those tests under tokio-console

ah thanks for that. Let me have a quick look and see how this could be incorporated in this PR. Otherwise I'll address on a follow up one.

> * the absence of a worker seems like a severe limitation given some workflows wait on worker interaction, how do we plan to address this component?

I've replied on the related comment. I'll include that on follow up PR.

PR #489, opened 2022-07-13
Reconfig consensus

asonnino:
## What this PR does:

Reconfigure the consensus crate and interface it with the Primary and worker. The strategy for reconfiguration is the following:
* The executor (run by the Sui authority) or any external entity (in case of Celo/Somelier) sends a ReconfigureNotification to the primary and the workers
* This reconfigure notification can be of two kinds. It can either provide a new committee, in which case all tasks perform internal cleanups and start using the new committee, or it tell all tasks to shutdown.
* End to end reconfiguration test 

## Next PRs:
* Currently the workers and the primary receive ReconfigureNotification messages in the same grpc service that listen to Primary Messages. @bmwill tells me that there is a cleaner way to do that without requring an extra TCP port to run Narwhal.
* Support committee changes when the epoch does not change. This include checking that the stake distribution stays the same and that only network addresses changed.


node/src/lib.rs
```diff
@@ -140,6 +150,8 @@ impl Node {
                 store,
                 parameters.clone(),
                 execution_state,
+                tx_reconfigure.subscribe(),
+                rx_reconfigure,
```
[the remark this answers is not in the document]
asonnino: make nicer names for the last two arguments or can we pass a pointer to the sender instead?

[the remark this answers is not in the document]
huitseeker: I think the pointer to the sender will be more clear, as long as you don't have to ferry that pointer across a thread boundary. If you do I'd revert to the first strategy.

akichidis (commented):
Great work @asonnino . Left a few comments - before I approve please have a look on them.

huitseeker (commented):
> In order to properly shutdown consensus without shutting down the executor, we need to have the consensus-executor connection to go through TCP (rather than a tokio channel).

There is inspiration to be found in the code of e.g. https://github.com/Finomnis/tokio-graceful-shutdown/tree/749b30adfcd5ecbf495cbfeaacf8e42c580edd85

consensus/src/consensus.rs
```diff
@@ -153,44 +165,94 @@ where
...
+                            let message = self.rx_reconfigure.borrow_and_update().clone();
+                            match message  {
+                                ReconfigureNotification::NewCommittee(new_committee) => {
+                                    state = self.reconfigure(new_committee)?;
+                                },
+                                ReconfigureNotification::Shutdown => return Ok(()),
+                            }
+                        }
+                        Ordering::Less => {
+                            // We already updated committee but the core is slow.
```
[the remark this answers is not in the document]
akichidis: Agreed, it might be spammy - well for a short period of time under healthy circumstances - but especially until we gain good trust on the stability of our nodes I would err on adding more logging rather less (doesn't have to be `info` - it could be a `debug`). For this specific part I would probably like to know if for some reason this component is dropping certificates and not doing any processing. We might be looking situations where the node is not making any progress and I would like to answer the question "why is my consensus component is not making progress?"

node/src/lib.rs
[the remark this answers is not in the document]
asonnino: good point, fixed

consensus/src/consensus.rs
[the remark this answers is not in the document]
asonnino: fixed

consensus/src/tests/subscriber_tests.rs
```diff
@@ -203,3 +213,40 @@ async fn subscribe_sync() {
...
+    while let Some(certificate) = certificates.pop_front() {
+        tx_consensus_input.send(certificate).await.unwrap();
+    }
+
+    // Ensure the first 4 ordered certificates have the expected consensus index. Note that we
+    // need to feed 5 certificates to consensus to trigger a commit.
+    for i in 0..=4 {
+        let output = rx_consensus_to_client.recv().await.unwrap();
+        assert_eq!(output.consensus_index, i);
+    }
```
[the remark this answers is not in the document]
asonnino: Yes it is just to ensure we can shutdown the subscriber while it is running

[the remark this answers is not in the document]
asonnino: it is not strictly needed

executor/src/core.rs
```diff
@@ -176,10 +204,14 @@ where
...
         &mut self,
+        consensus_output: &ConsensusOutput<PublicKey>,
         serialized: SerializedTransaction,
         total_transactions: usize,
         total_batches: usize,
-    ) -> SubscriberResult<Vec<u8>> {
+    ) -> SubscriberResult<(
+        <State as ExecutionState>::Outcome,
+        Option<Committee<PublicKey>>,
+    )> {
```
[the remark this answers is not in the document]
asonnino: good idea

primary/src/state_handler.rs
```diff
@@ -81,30 +85,39 @@ impl<PublicKey: VerifyingKey> StateHandler<PublicKey> {
...
-                    break;
+
+                    // Exit only when we are sure that all the other tasks received
+                    // the shutdown message.
+                    if shutdown {
+                        while self.tx_reconfigure.receiver_count() != 0 {
+                            tokio::time::sleep(tokio::time::Duration::from_millis(10)).await;
+                        }
+                        return;
+                    }
```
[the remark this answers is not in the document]
asonnino: good idea to use the closed approach, trying that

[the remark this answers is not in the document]
asonnino: done, much nicer indeed!

primary/src/primary.rs
```diff
@@ -487,13 +491,20 @@ impl<PublicKey: VerifyingKey> PrimaryToPrimary for PrimaryReceiverHandler<Public
...
+                .tx_state_handler
+                .send(notification)
+                .await
+                .map_err(|_| DagError::ShuttingDown),
             _ => self
                 .tx_primary_messages
                 .send(message)
                 .await
-                .expect("Failed to send certificate"),
+                .map_err(|_| DagError::ShuttingDown),
```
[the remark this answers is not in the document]
asonnino: you are completely right. I will move this message to the primary-worker channel (that is not exposed to WAN) until a next PR does what Brandon suggested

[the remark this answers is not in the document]
bmwill: The suggestion I had was to have a new gRPC service that only handles these types of messages. Multiple gRPC services can hang off of the same gRPC server so we wouldn't need to use a whole new interface if we didn't want to.

consensus/src/consensus.rs
```diff
@@ -153,44 +165,94 @@ where
...
-                    .expect("Failed to send certificate to primary");
+                        if let Err(e) = self.tx_output.send(output).await {
+                            tracing::warn!("Failed to output certificate: {e}");
+                        }
+                    }
+                },
 
-                if let Err(e) = self.tx_output.send(output).await {
-                    tracing::warn!("Failed to output certificate: {e}");
+                // Check whether the committee changed.
```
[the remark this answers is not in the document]
asonnino: yes we can pull the new committee upon seeing a certificate with a higher epoch number. this works for most tasks but then they can't be shut down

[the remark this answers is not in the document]
asonnino: mmh unless we gracefully shutdown when the channel drops. That would be nice

primary/src/primary.rs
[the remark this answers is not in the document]
akichidis: > you are completely right. I will move this message to the primary-worker channel (that is not exposed to WAN) until a next PR does what Brandon suggested

I am ok with this as a trade off to make this PR progress but I would like to ask to prioritise the next PR to address this properly.

@bmwill If we use a new gRPC service and attached it under the same gRPC server then this will still be exposed under the same port, right? If that's true then it's not clear to me how we could prevent other primaries from accessing the endpoints of this service (given that we don't have any kind of authentication - yet) . Or did you have in mind some other way?

node/src/lib.rs
```diff
@@ -266,3 +295,104 @@ impl Node {
         handles
     }
 }
+
+/// Module to start a node (primary, workers and default consensus), keep it running, and restarting it
+/// every time the committee changes.
+pub struct NodeController;
+
+impl NodeController {
```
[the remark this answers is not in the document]
asonnino: Good idea, NodeRestarter is better name (and it can go in its own module)

huitseeker (changes requested):
You're looking for the `BATCH_TIMEOUT` in `primary/src/block_waiter.rs`. Sadly I have not much more to say than we can't afford to `#[ignore]` the concerned tests, as they are load-bearing in our integration w/ partners.

akichidis:
> The following test have been disabled and should be fixed and re-enabled:
> 
> ```
> test test_get_collections ... ignored, Strange timeout error
> test test_get_collections_with_missing_certificates ... ignored, Strange timeout error
> ```

@asonnino I had a quick look on it and it seems that there is some trouble between the worker ---> primary when trying to send at once more batches than one. The `get_collections` is retrieving the payload (batches) for each provided collection id (certificate id). In this test each certificate (collection) has only one batch. Requesting a single collection works fine. When requesting multiple in the same time, it seems that eventually only one sent back from the worker to the primary. I tried to trace the path and I managed to see that [this is being called only once](https://github.com/MystenLabs/narwhal/blob/6868535bc0b2ad64bd53042d1c699610e933f556/primary/src/primary.rs#L579). Given that we had some changes in the worker (on the [primary_connector](https://github.com/MystenLabs/narwhal/pull/489/files#diff-05fb9bcbdbd0ddcd923afde36b1131ebb85a1aa3c1ce879bc4ca4ea562d22191)) maybe we want to have a look there

worker/src/primary_connector.rs
```diff
@@ -3,94 +3,135 @@
...
-            .await
-            .map_err(Into::into)
-            .map(|_| ())
+impl WorkerToPrimaryNetwork {
+    pub async fn send<PublicKey: VerifyingKey>(
+        &mut self,
+        address: Multiaddr,
+        message: &WorkerPrimaryMessage<PublicKey>,
+    ) -> CancelHandler<()> {
+        if self.client.is_none() {
```
akichidis, line 113: I think it's quite implicit the initialisation of the client here. This method once the client is created it will not initialise a new one even if we need to communicate to a different address. I am aware that we don't expect the primary's address to change, but since this is the case I would refactor to initialise the client once the `WorkerToPrimaryNetwork` is created. This will make clear that when we `send` a message it will always be sent to the right address.

Otherwise, if we expect the primary's address to change via reconfiguration - then again this should be refactored as the client will keep pointing to the old address

worker/src/primary_connector.rs
```diff
@@ -3,94 +3,135 @@
...
             tokio::select! {
                 // Send the digest through the network.
                 Some(digest) = self.rx_digest.recv() => {
-                    // We don't care about the error
-                    let _ = self.primary_client.send(&digest).await;
+                    let address = self.committee
+                        .primary(&self.name)
+                        .expect("Our public key is not in the committee")
+                        .worker_to_primary;
+                    _handle = self.primary_client.send(address, &digest).await;
```
akichidis, line 64: Very interest, when I do this:
```
 let handle = self.primary_client.send(address, &digest).await;
 handle.await;
```

the issues observed here https://github.com/MystenLabs/narwhal/pull/489#issuecomment-1192490074 are resolved and the tests are able to pass correctly. So we are probably looking to some sort of concurrency issue? Need to look further into it.

akichidis: Ok , so this is not a concurrency issue. The `CancelHandler` returned by the `send` method is aborting the task when dropped: https://github.com/MystenLabs/narwhal/blob/main/network/src/lib.rs#L29

So based on the info I posted here https://github.com/MystenLabs/narwhal/pull/489#issuecomment-1192490074 (about handlers getting overwritten) , indeed when the previous handler is dropped it has a consequence for the underlying task to get cancelled. That explains what we see here.

So we probably have two ways forward:
1. Don't use the `CancelHandle` struct, but rather return just a JoinHandle
2. Save the handles until they are finished - maybe we don't want to do that though, since the approach is fire and forget

arun-koshy: This makes sense, when I was investigating the failure I tried running the demo client against this PR and it would fail on calls to endpoints other than `GetCollections`. (i.e. `Rounds`, `ReadCausal`, etc.). Consistently seeing connections being dropped in the primaries.

```
2022-07-21T22:02:27.082935Z DEBUG tonic::codec::decode: decoder inner stream error: Status { code: Unknown, message: "error reading a body from connection: stream error received: stream no longer needed", source: Some(hyper::Error(Body, Error { kind: Reset(StreamId(1), CANCEL, Remote) })) }
```
[primary-1.log](https://github.com/MystenLabs/narwhal/files/9169689/primary-1.log)
[demo_client.log](https://github.com/MystenLabs/narwhal/files/9169694/demo_client.log)

asonnino: 🫢 wow that was silly! thanks for your help, it was right under my nose. it probably means I need a vacation

akichidis: To be honest its not apparent the functionality of the `CancelHandler` (aborting the task when dropped) - I mean someone could easily miss that if not familiar with it (I wasn't as well). We'll probably rename this in a separate PR and add some comments as well so it becomes more apparent what someone should expect.

asonnino:
Thanks everyone! I fixed the comments and all tests, it is much better now

worker/src/primary_connector.rs
huitseeker: I would point out the reason we wrap the `.broadcast` and `.send` operations on primaries and workers is that they are on an infinite retry from a network PoV -> if you're not baby-sitting the cancellation handlers, you should probably not use those operations, but rather their non-infinite-retry variants (`unreliable_broadcast`, `unreliable_send`).

worker/src/primary_connector.rs
```diff
@@ -3,94 +3,152 @@
...
+                    .retry(move || {
+                        let mut client = client.clone();
+                        let message = message.clone();
+                        async move { client.send_message(message).await.map_err(Into::into) }
+                    })
+                    .map(|response| {
+                        response.expect("we retry forever so this shouldn't fail");
+                    }),
+            )
+            .await;
```
huitseeker, line 151: This is effectively bypassing the changes brought in in #559 and should rather build on top of them. We've fixed some nasty behavior there, so that PR may need review.

asonnino: Ah! I though this PR would be merged before #559

asonnino: let me see how to fix that

asonnino: fixed

network/src/worker.rs
```diff
@@ -167,3 +169,84 @@ impl WorkerNetwork {
...
+            retrying_max_elapsed_time: None,
+            ..Default::default()
+        };
+
+        Self {
+            address: None,
+            client: Default::default(),
+            config: Default::default(),
+            retry_config,
+            executor: BoundedExecutor::new(MAX_TASK_CONCURRENCY, Handle::current()),
```
huitseeker, line 194: This breaks the primitive that `BoundedExecutor` is now meant to be per address. Please give `WorkerToPrimaryNetwork` a comment indicating it's meant to one-to-one communication.

network/src/worker.rs
```diff
@@ -167,3 +169,84 @@ impl WorkerNetwork {
...
+        //             .retry(move || {
+        //                 let mut client = client.clone();
+        //                 let message = message.clone();
+        //                 async move { client.send_message(message).await.map_err(Into::into) }
+        //             })
+        //             .map(|response| {
+        //                 response.expect("we retry forever so this shouldn't fail");
+        //             }),
+        //     )
+        //     .await;
```
huitseeker, line 249: Please remove.

network/src/worker.rs
```diff
@@ -167,3 +169,84 @@ impl WorkerNetwork {
...
+            client: Default::default(),
+            config: Default::default(),
+            retry_config,
+            executor: BoundedExecutor::new(MAX_TASK_CONCURRENCY, Handle::current()),
+        }
+    }
+}
+
+impl WorkerToPrimaryNetwork {
+    pub async fn send<PublicKey: VerifyingKey>(
```
huitseeker, line 200: Please reproduce & adapt the "Safety" comment you'll find on send above. Ideally, we would be able to justify of a bound on the # of messages sent per round.

consensus/src/consensus.rs
```diff
@@ -248,40 +276,76 @@ where
...
+                            }
+                        }
+                        Ordering::Less => {
+                            // We already updated committee but the core is slow.
+                            tracing::debug!("Already moved to the next epoch");
+                            continue
+                        },
+                        Ordering::Equal => {
+                            // Nothing to do, we can proceed.
+                        }
```
huitseeker, line 301: This is essentially the same as the proposer, right? There's probably a trait lurking under this.

node/src/restarter.rs
```diff
@@ -0,0 +1,126 @@
...
+        execution_state: Arc<State>,
+        parameters: Parameters,
+        mut rx_reconfigure: Receiver<(Keys, Committee<PublicKey>)>,
+        tx_output: Sender<ExecutorOutput<State>>,
+    ) where
+        PublicKey: VerifyingKey,
+        State: ExecutionState + Send + Sync + 'static,
+        State::Outcome: Send + 'static,
+        State::Error: Debug,
+        Keys: KeyPair<PubKey = PublicKey> + Signer<PublicKey::Sig> + Send + 'static,
```
huitseeker, line 33: On the one hand, I'm glad you came around to trait contraints. On the other hand, please open an issue for me to mop this up later (you should be able to grab the most-constrained trait, i.e. KeyPair, and obtain the rest through projection, i.e. it's easier to do
```rust
foo<Keys: KeyPair>(
    ...
    Bar<Keys::PubKey>
    ...
)
```
rather than 
```rust 
foo<Keys: KeyPair<PubKey = PublicKey> + Signer<PublicKey::Sig> + Send + 'static, PublicKey: VerifyingKey> (
    ...
   Bar<PublicKey>
    ...
)
```
Moreover, `KeyPair` includes the Signer constraint, which you should not need to repeat)

node/src/restarter.rs
```diff
@@ -0,0 +1,126 @@
...
+            join_all(worker_cancel_handles).await;
+            tracing::debug!("Committee reconfiguration message successfully sent");
+
+            // Wait for the components to shut down.
+            join_all(handles.drain(..)).await;
+            tracing::debug!("All tasks successfully exited");
+
+            // Give it an extra second in case the last task to exit is a network server. The OS
+            // may need a moment to make the TCP ports available again.
+            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
```
huitseeker, line 117: There is a better method for doing this in the network crate if you know which ports you'll claim. Do you?

primary/tests/epoch_change.rs
```diff
@@ -67,17 +72,30 @@ async fn test_simple_epoch_change() {
...
+            .collect();
+        let message = WorkerPrimaryMessage::Reconfigure(ReconfigureNotification::NewCommittee(
+            new_committee.clone(),
+        ));
+        let mut _do_not_drop = Vec::new();
+        for address in addresses {
+            _do_not_drop.push(
+                WorkerToPrimaryNetwork::default()
+                    .send(address, &message)
+                    .await,
```
huitseeker, line 97: You could type the vector `let mut _do_not_drop: Vec<CancelHandle<()>> = Vec::new()`, that would help, and I have an upcoming refactor that helps identifying the behavior by renaming `CancelHandle` -> `CancelOnDropHandle`.

huitseeker (commented):
Thanks! It looks like this herculean work is pretty close to done.

The main think I would ask: would it be possible for us to have per-component logging of acting on a reconfiguration notification at the debug level? If for any reason one reconfig doesn't work out I'd like us to have one log line that we should notice the absence of. A span for the component in question (perhaps in the run method) would probably help.

I want us to have some conventions around someone writing a new component and forgetting to manage the `Reconfiguration`. A place where a few comments would give them directions would help.

executor/src/core.rs
```diff
@@ -150,27 +174,46 @@ where
...
 
-                // Output the result (eg. to notify the end-user);
-                let output = (result.clone(), transaction);
-                if self.tx_output.send(output).await.is_err() {
-                    debug!("No users listening for transaction execution");
-                }
+                let (bail, result) = match result {
+                    Ok((outcome, committee)) => {
+                        if let Some(_committee) = committee {
+                            // todo!();
```
huitseeker, line 188: Could we leave a note about what's left to do here? Perhaps clarify the conditions in which we would return `bail`? Log the epoch change, which may condition the validity of the sequencing output by consensus?

worker/src/primary_connector.rs
asonnino: Fixed

worker/src/primary_connector.rs
asonnino: yes I agree, but it is probably a bad idea to rename all CancelHandler in this PR (it is already quite big)

consensus/src/consensus.rs
asonnino: That could be nice indeed. I will think about it in a next PR

asonnino:
> Possible follow-ups:
> 
> * Did you look into using [Cluster](https://github.com/MystenLabs/narwhal/blob/main/test_utils/src/cluster.rs) to write an e2e test? I found #560 inspiring

Not yet because this PR was concurrent to this one. I will have a look and see how to refactor the integration tests. 

> * I would like us to have some conventions around the scenario of someone writing a new component and forgetting to manage the `Reconfiguration`. A place where we'd have a few comments that would give them directions would help.

I agree, I am not a fan of having to repeat the following pattern over and over:
```
result = self.rx_reconfigure.changed() => {
     result.expect("Committee channel dropped");
     let message = self.rx_reconfigure.borrow().clone();
    match message {
```

Although I would favour to land this PR fast because it touches a lot of files and people will have to rebase on it pretty aggressively

asonnino:
> Thanks! It looks like this herculean work is pretty close to done.
> 
> The main think I would ask: would it be possible for us to have per-component logging of acting on a reconfiguration notification at the debug level? i.e. can we add `debug!("Just reconfigured to committee: { X, Y, Z }")` to the many blocks :
> 
> ```rust
>     ReconfigureNotification::NewCommittee(new_committee) => {
>     ...
>     },
>     ReconfigureNotification::Shutdown  =>
> ```
> 
> throughout the PR?
> 
> If for any reason one reconfig doesn't work out I'd like us to have one log line that we should notice the absence of. A span for the component in question (perhaps in the run method) would probably help.

Good idea, it is done

network/src/worker.rs
asonnino: fixed

network/src/worker.rs
asonnino: oops

network/src/worker.rs
asonnino: fixed

consensus/src/consensus.rs
asonnino: That could be nice indeed. I will think about it in a next PR

node/src/restarter.rs
asonnino: I also though so but if I remove the Signer trait I get the following error:
```
--> node/src/restarter.rs:53:35
   |
53 |             let primary_handles = Node::spawn_primary(
   |                                   ^^^^^^^^^^^^^^^^^^^ expected crypto::traits::VerifyingKey::Sig, found crypto::traits::KeyPair::Sig
   |
   = note: expected associated type `<PublicKey as VerifyingKey>::Sig`
              found associated type `<Keys as KeyPair>::Sig`
    ```

asonnino: Fixed #579

primary/tests/epoch_change.rs
asonnino: good point, fixed

node/src/restarter.rs
asonnino: I don't really know which port that is sadly, it can be any of them (on the Primary or any worker)

asonnino: what is the method?

huitseeker (approved):
Thanks a ton, Alberto!

PR #492, opened 2022-07-14
[testing] return joinhandles and add store preserve support in cluster

akichidis:
A follow up of https://github.com/MystenLabs/narwhal/pull/487 , is doing the following:

1) Making the code return JoinHandles from all the parts that we spawn new tasks
2) Because of the above we are now able to properly shutdown all the running parts of the primary when we stop a node and _properly close the database_ (automatically closed once all references are dropped)
3) Refactoring the `Cluster` struct to allow us preserve storage when node is restarted. That will allow us to test our network under different assumptions

**Note:** Interestingly enough the test now passes when we preserve the storage - the restarted node manages to perform the causal completion successfully, catch up the other nodes and make progress to the same commit round.

### DIFF
--- a/executor/src/subscriber.rs
+++ b/executor/src/subscriber.rs
@@ -52,7 +52,7 @@ impl<PublicKey: VerifyingKey> Subscriber<PublicKey> {
         tx_batch_loader: Sender<ConsensusOutput<PublicKey>>,
         tx_executor: Sender<ConsensusOutput<PublicKey>>,
         next_consensus_index: SequenceNumber,
-    ) -> JoinHandle<SubscriberResult<()>> {
+    ) -> JoinHandle<()> {
         tokio::spawn(async move {
             Self {
                 store,
@@ -64,6 +64,7 @@ impl<PublicKey: VerifyingKey> Subscriber<PublicKey> {
             }
             .run()
             .await
+            .unwrap();
         })
     }
 
--- a/test_utils/src/cluster.rs
+++ b/test_utils/src/cluster.rs
@@ -4,12 +4,11 @@ use crate::{committee, keys, temp_dir};
 use arc_swap::ArcSwap;
 use config::{Committee, Parameters};
 use crypto::{ed25519::Ed25519PublicKey, traits::KeyPair};
-use executor::SubscriberResult;
 use node::{
     execution_state::SimpleExecutionState, metrics::primary_metrics_registry, Node, NodeStorage,
 };
 use prometheus::Registry;
-use std::{collections::HashMap, sync::Arc, time::Duration};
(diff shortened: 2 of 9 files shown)


test_utils/src/cluster.rs
```diff
@@ -34,13 +45,25 @@ impl Cluster {
...
         for (index, key) in k.into_iter().enumerate() {
             info!("Key {index} -> {}", key.public().clone());
+
+            nodes.insert(
+                index,
+                Arc::new(NodeDetails {
+                    id: index,
+                    name: key.public().clone(),
+                    registry: Registry::new(),
+                    store_path: Default::default(),
```
huitseeker, line 59: Since this is a `String`, isn't there a risk that if you then call `start_node` with `preserve_store`, this will lead to a start at the system root (`/`) ? 

One way to get out of this conundrum, I think is to use [`env::temp_dir`](https://doc.rust-lang.org/std/env/fn.temp_dir.html) - i.e. on a Unixoid the equivalent of the $TMPDIR env variable - for the default value of the path. In many cases, this will resolve to `/tmp`. This is not `tempdir` from the associated crate, which on an Unixoid resolves to an `mktemp`.

huitseeker (commented):
> Because of the above we are now able to properly shutdown all the running parts of the primary when we stop a node and properly close the database (automatically closed once all references are dropped)

Love to see it, this is excellent! 🚀

test_utils/src/cluster.rs
akichidis: Ah very good catch, I'll fix this

test_utils/src/cluster.rs
```diff
@@ -83,7 +114,15 @@ impl Cluster {
...
-        let store: NodeStorage<Ed25519PublicKey> = NodeStorage::reopen(temp_dir());
+        let store_path = if preserve_store {
+            node.store_path.clone()
+        } else {
+            temp_dir()
+        };
+
+        info!("Node {} will use path {:?}", id, store_path.clone());
+
+        let store: NodeStorage<Ed25519PublicKey> = NodeStorage::reopen(store_path.clone());
```
laura-makdah, line 125: 👏

executor/src/subscriber.rs
```diff
@@ -64,6 +64,7 @@ impl<PublicKey: VerifyingKey> Subscriber<PublicKey> {
             }
             .run()
             .await
+            .unwrap();
```
laura-makdah, line 67: If we are going to have unwraps on failure to run components, we might as well have them be `.unwrap_or_else(..)` with log a message that includes the error?

akichidis:
Closing as [this has been merged ](https://github.com/MystenLabs/narwhal/pull/498) which already contains the above changes

PR #493, opened 2022-07-14
refactor: making entry.sh more generic for k8s, etc

allan-bailey:
This is so I can override the data and key paths for the k8s templates.


allan-bailey:
Correction: this is so I can put the committee.json, parameters.json and key.json files into a configmap mounted at /config

huitseeker (approved):
👍

PR #496, opened 2022-07-15
Make the certificate waiter fire on a timeout

huitseeker:
## Context

The Certificate Waiter operates in a loop that checks:
- whether we were told to process a new certificate, checking which of its missing parents are to be added to a pending queue,
- whether we received any of these awaited certificates,
- whether we were notified of an epoch change,
and fires a GC routine that prunes certificates past a round boundary, once we've processed either of these cases.
Indeed, certificates past the "GC bound" are no longer available on the network, and so waiting to request them from others is pointless.

## The issue:

It may be that none of the three causes above fire:
- the synchronizer only receives certificates with no missing ancestors, thereby never contacting us,
- we don't hear back about any of our requests,
- there is no epochchange,

Typically this happens if we are catching up, as we pile up a ton of certificates in pending that are not so old for us, but which expire past the GC bound as we catch up and realize the network is much further ahead in terms of rounds:
None of those certificates are served by the network (for others, they are past the GC bound), and as we catch up, we do not trigger the GC per se.

## The fix:

Trigger the GC logic periodically, on a `GC_RESOLUTION` basis.

## Commit organization
1. a test that does not pass,
2. the fix that makes the test pass,
3. a small commit ignoring this long-running test until we have time to improve it.


primary/src/certificate_waiter.rs
```diff
@@ -158,14 +165,43 @@ impl<PublicKey: VerifyingKey> CertificateWaiter<PublicKey> {
...
+    fn clean_pending(
+        round: Round,
+        gc_depth: u64,
+        pending: &mut HashMap<HeaderDigest, (Round, oneshot::Sender<()>)>,
+    ) -> Option<HashMap<HeaderDigest, (Round, oneshot::Sender<()>)>> {
+        // Cleanup internal state. Deliver the certificates waiting on garbage collected ancestors.
+        if round > gc_depth {
+            let gc_round = round - gc_depth;
+            Some(
+                pending
```
akichidis, line 204: Since the `pending` is passed as a &mut , once we drain, isn't this going to drain the map in any case? I mean shouldn't we re-assign whatever left to the the pending list?

akichidis (approved):
Great work @huitseeker 💯  Left a few comments, but approved to unblock

PR #498, opened 2022-07-15
[testing] add cluster workers support

akichidis:
Follow up of the #492 .

This PR is giving a generous refactoring in the `Cluster` struct so we have proper organisation of the nodes. Now the cluster is spinning up Authority nodes. Each Authority has `one primary` and `multiple workers`. Now each node type (worker & primary) are responsible for start /stop and report running status - which makes the organisation much cleaner. Helper methods and input configuration has been provided to allow start / stop nodes in a flexible way. Also the transactions addresses of the workers can now be accessed in order to be able to send transactions to the cluster.

As before , the prometheus Registries for the primary & worker nodes can be accessed to observe the metrics.

Next steps: 
1. expose the receiver of the `tx_transaction_confirmation` consensus channel of the primary so tests can be written to monitor the consensus output transactions
2. enable starting the cluster with consensus off (help us test cases with external consensus)

Apart from the above an important next step is to create tests around the bootstrap sequence of nodes to start addressing [this](https://www.notion.so/mystenlabs/Narwhal-Storm-July-11-4f3cc6c4698448c5820d3e949bc3764e#b7b1cec7f6dc42df9fa6938935c9c506)

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -10,6 +10,7 @@ use tokio::{
     sync::mpsc::{Receiver, Sender},
     task::JoinHandle,
 };
+use tracing::info;
 use types::{
     Certificate, CertificateDigest, ConsensusPrimaryMessage, ConsensusStore, Round, StoreResult,
 };
@@ -91,6 +92,9 @@ impl<PublicKey: VerifyingKey> ConsensusState<PublicKey> {
         gc_depth: Round,
     ) -> Dag<PublicKey> {
         let mut dag: Dag<PublicKey> = HashMap::new();
+
+        info!("Last committed round: {}", last_committed_round);
+
         let min_round = last_committed_round.saturating_sub(gc_depth);
         let cert_map = cert_store
             .iter(Some(Box::new(move |(_dig, cert)| {
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -10,7 +10,7 @@ use tokio::{
     sync::mpsc::{Receiver, Sender},
     task::JoinHandle,
 };
-use tracing::info;
+use tracing::{info, instrument};
 use types::{
     Certificate, CertificateDigest, ConsensusPrimaryMessage, ConsensusStore, Round, StoreResult,
 };
@@ -54,6 +54,7 @@ impl<PublicKey: VerifyingKey> ConsensusState<PublicKey> {
         }
     }
 
+    #[instrument(level = "info", skip_all)]
     pub async fn new_from_store(
         genesis: Vec<Certificate<PublicKey>>,
         metrics: Arc<ConsensusMetrics>,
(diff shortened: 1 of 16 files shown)


laura-makdah (approved):
If you find a way to expose `tr_transaction_confirmation` before merging this that would be 💯 otherwise I'll take it from here

test_utils/src/cluster.rs
```diff
@@ -2,174 +2,514 @@
...
+#[derive(Clone)]
+pub struct WorkerNodeDetails {
+    pub id: WorkerId,
+    pub transactions_address: Multiaddr,
+    pub registry: Registry,
+    name: Ed25519PublicKey,
+    committee: SharedCommittee<Ed25519PublicKey>,
+    parameters: Parameters,
+    store_path: PathBuf,
+    handlers: Arc<ArcSwap<Vec<JoinHandle<()>>>>,
```
huitseeker, line 308: This is probably a nit, I'll leave it to you to consider whether it's worth opening an issue to follow-up on this:

It is possible to me that the `ArcSwap` here might be an overkill: the `ArcSwap` is most useful to me when you have at least one mutation of a value from a non-empty state, i.e. `None` -> `Some(1)` -> `Some(42)`. (the `Option` is figurative here and just there to let me express the notion of "empty" state, not actually involved in this).

If I understand what you're trying to do correctly, you're more probably trying to initialize something from an empty state, and then bring it back to that empty state before setting it again, i.e. `None` -> `Some(1)`-> `None` -> `Some(42)`. For this consider that the `OnceCell` type has an initialization (and a `get_or_try_init` function) but also a `take` function that brings things back to this empty state.

huitseeker (approved):
Thanks a bunch!

primary/tests/causal_completion_tests.rs
```diff
@@ -46,12 +46,24 @@ async fn test_read_causal_signed_certificates() {
...
+
+    // show case the transactions channel
+    /*
+    let mut receiver = node.primary.tx_transaction_confirmation.subscribe();
+    loop {
+        if let Ok(result) = receiver.recv().await {
+            // do something here
+        }
+    }
+     */
```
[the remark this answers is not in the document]
akichidis: Ah I've added this to showcase @laura-makdah how could use the `tx_transaction_confirmation` to receive the consensus output

test_utils/src/cluster.rs
akichidis: Ah thanks for the above. To be honest my intuition was also that probably the `ArcSwap` might be an overkill. Unfortunately I can't use `OnceCell` as `PrimaryNodeDetails` needs to be cloned (`#[derive(Clone)]` is on top) which ends up in linter errors. I've switched to using an `Rc<RefCell>` instead as I believe is what I really need (interior mutability behind a way to share across clones)

consensus/src/consensus.rs
```diff
@@ -91,7 +92,9 @@ impl<PublicKey: VerifyingKey> ConsensusState<PublicKey> {
         gc_depth: Round,
     ) -> Dag<PublicKey> {
         let mut dag: Dag<PublicKey> = HashMap::new();
-        let min_round = last_committed_round - gc_depth;
+        info!("Last committed round: {}", last_committed_round);
```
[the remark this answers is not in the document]
akichidis: Good point, I'll amend this!

test_utils/src/cluster.rs
```diff
@@ -2,174 +2,514 @@
...
+
+impl PrimaryNodeDetails {
+    fn new(
+        id: usize,
+        key_pair: Ed25519KeyPair,
+        parameters: Parameters,
+        committee: SharedCommittee<Ed25519PublicKey>,
+    ) -> Self {
+        // used just to initialise the struct value
+        let (tx, _) = tokio::sync::broadcast::channel(1);
```
laura-makdah, line 196: tr isn't being used

PR #499, opened 2022-07-15
fix: Make the table of pending certificates concurrency-safe

huitseeker:
Post #496 and in fact post the oneshot channel changes introduced in #486, the pending table of the `CertificateWaiter` can be accessed concurrently.

In some cases, our benchmark has revealed that the drop triggered by a call to `clear()` can interact with the `drain()` operation to create a double-send on the same `oneshot::Sender`, which panics.
(see comments on https://github.com/MystenLabs/narwhal/commit/29e853d2d65e543fa0f1dbc634d01576c0e8e60e for evidence and re-run the bench locally at that commit to repro)

This is now visible, because the GC actually fires more often post #496. Prior to #486, this issue was invisible, since we were using `mpsc` channels (which allow double sends).

This fixes the benchmark.

### DIFF
--- a/primary/src/certificate_waiter.rs
+++ b/primary/src/certificate_waiter.rs
@@ -4,16 +4,15 @@
 use crate::metrics::PrimaryMetrics;
 use config::Committee;
 use crypto::traits::VerifyingKey;
+use dashmap::DashMap;
 use futures::{
     future::try_join_all,
     stream::{futures_unordered::FuturesUnordered, StreamExt as _},
 };
-use std::{
-    collections::HashMap,
-    sync::{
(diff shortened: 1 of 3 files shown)


lanvidr (approved):

primary/src/certificate_waiter.rs
```diff
@@ -178,44 +180,29 @@ impl<PublicKey: VerifyingKey> CertificateWaiter<PublicKey> {
...
+                        true
+                    }
+                });
             }
 
             self.update_metrics();
         }
     }
 
-    fn clean_pending(
```
laura-makdah, line 191: Are all places that called this function also updated?

PR #507, opened 2022-07-18
PublicKeyBytes trait

punwai:
Connected PublicKeyBytes to existing associated traits to enable generic functions on Sui. Please do suggest any other way I could do this, atleast before the rust devs get back to work.

### DIFF
--- a/crypto/src/bls12377/mod.rs
+++ b/crypto/src/bls12377/mod.rs
@@ -42,7 +42,7 @@ pub struct BLS12377PublicKey {
     pub bytes: OnceCell<[u8; CELO_BLS_PUBLIC_KEY_LENGTH]>,
 }
 
-pub type BLS12377PublicKeyBytes = PublicKeyBytes<BLS12377PublicKey, { CELO_BLS_PUBLIC_KEY_LENGTH }>;
+pub type BLS12377PublicKeyBytes = PublicKeyBytes<BLS12377PublicKey, { BLS12377PublicKey::LENGTH }>;
 
 #[readonly::make]
 #[derive(Debug)]
@@ -259,7 +259,6 @@ impl<'a> From<&'a BLS12377PrivateKey> for BLS12377PublicKey {
 impl VerifyingKey for BLS12377PublicKey {
     type PrivKey = BLS12377PrivateKey;
     type Sig = BLS12377Signature;
-    type Bytes = BLS12377PublicKeyBytes;
     const LENGTH: usize = CELO_BLS_PUBLIC_KEY_LENGTH;
 
     fn verify_batch(msg: &[u8], pks: &[Self], sigs: &[Self::Sig]) -> Result<(), signature::Error> {
@@ -549,19 +548,6 @@ impl TryFrom<BLS12377PublicKeyBytes> for BLS12377PublicKey {
 
 impl From<&BLS12377PublicKey> for BLS12377PublicKeyBytes {
     fn from(pk: &BLS12377PublicKey) -> Self {
-        let pk_bytes =
-            pk
-            .bytes
-            .get_or_try_init::<_, eyre::Report>(|| {
-                let mut bytes = [0u8; CELO_BLS_PUBLIC_KEY_LENGTH];
-                pk.pubkey
-                    .as_ref()
-                    .into_affine()
-                    .serialize(&mut bytes[..])
-                    .unwrap();
-                Ok(bytes)
-            })
-            .expect("OnceCell invariant violated");
-        BLS12377PublicKeyBytes::new(pk_bytes)
+        BLS12377PublicKeyBytes::from_bytes(pk.as_ref()).unwrap()
     }
 }
--- a/crypto/src/bls12377/mod.rs
+++ b/crypto/src/bls12377/mod.rs
@@ -543,20 +543,25 @@ impl TryFrom<BLS12377PublicKeyBytes> for BLS12377PublicKey {
     type Error = signature::Error;
 
     fn try_from(bytes: BLS12377PublicKeyBytes) -> Result<BLS12377PublicKey, Self::Error> {
-        // TODO(https://github.com/MystenLabs/sui/issues/101): Do better key validation
-        // to ensure the bytes represent a poin on the curve.
         BLS12377PublicKey::from_bytes(bytes.as_ref()).map_err(|_| Self::Error::new())
     }
 }
 
 impl From<&BLS12377PublicKey> for BLS12377PublicKeyBytes {
-    fn from(s: &BLS12377PublicKey) -> Self {
-        let mut bytes = [0u8; CELO_BLS_PUBLIC_KEY_LENGTH];
-        s.pubkey
-            .as_ref()
-            .into_affine()
-            .serialize(&mut bytes[..])
-            .unwrap();
-        BLS12377PublicKeyBytes::new(bytes)
+    fn from(pk: &BLS12377PublicKey) -> Self {
+        let pk_bytes =
+            pk
+            .bytes
+            .get_or_try_init::<_, eyre::Report>(|| {
+                let mut bytes = [0u8; CELO_BLS_PUBLIC_KEY_LENGTH];
+                pk.pubkey
+                    .as_ref()
+                    .into_affine()
+                    .serialize(&mut bytes[..])
+                    .unwrap();
+                Ok(bytes)
+            })
+            .expect("OnceCell invariant violated");
+        BLS12377PublicKeyBytes::new(pk_bytes)
     }
 }
--- a/crypto/src/bls12377/mod.rs
+++ b/crypto/src/bls12377/mod.rs
@@ -42,7 +42,7 @@ pub struct BLS12377PublicKey {
     pub bytes: OnceCell<[u8; CELO_BLS_PUBLIC_KEY_LENGTH]>,
 }
 
-pub type BLS12377PublicKeyBytes = PublicKeyBytes<BLS12377PublicKey, { BLS12377PublicKey::LENGTH }>;
+pub type BLS12377PublicKeyBytes = PublicKeyBytes<BLS12377PublicKey, { CELO_BLS_PUBLIC_KEY_LENGTH }>;
 
 #[readonly::make]
 #[derive(Debug)]
@@ -259,6 +259,7 @@ impl<'a> From<&'a BLS12377PrivateKey> for BLS12377PublicKey {
 impl VerifyingKey for BLS12377PublicKey {
     type PrivKey = BLS12377PrivateKey;
     type Sig = BLS12377Signature;
+    type Bytes = PublicKeyBytes<BLS12377PublicKey, { CELO_BLS_PUBLIC_KEY_LENGTH }>;
     const LENGTH: usize = CELO_BLS_PUBLIC_KEY_LENGTH;
 
     fn verify_batch(msg: &[u8], pks: &[Self], sigs: &[Self::Sig]) -> Result<(), signature::Error> {
@@ -538,20 +539,24 @@ impl AggregateAuthenticator for BLS12377AggregateSignature {
 /// Implement VerifyingKeyBytes
 ///
 
-impl TryInto<BLS12377PublicKey> for BLS12377PublicKeyBytes {
+impl TryFrom<BLS12377PublicKeyBytes> for BLS12377PublicKey {
     type Error = signature::Error;
 
-    fn try_into(self) -> Result<BLS12377PublicKey, Self::Error> {
+    fn try_from(bytes: BLS12377PublicKeyBytes) -> Result<BLS12377PublicKey, Self::Error> {
         // TODO(https://github.com/MystenLabs/sui/issues/101): Do better key validation
         // to ensure the bytes represent a poin on the curve.
-        BLS12377PublicKey::from_bytes(self.as_ref()).map_err(|_| Self::Error::new())
+        BLS12377PublicKey::from_bytes(bytes.as_ref()).map_err(|_| Self::Error::new())
     }
 }
 
-impl From<BLS12377PublicKey> for BLS12377PublicKeyBytes {
-    fn from(pk: BLS12377PublicKey) -> BLS12377PublicKeyBytes {
+impl From<&BLS12377PublicKey> for BLS12377PublicKeyBytes {
+    fn from(s: &BLS12377PublicKey) -> Self {
         let mut bytes = [0u8; CELO_BLS_PUBLIC_KEY_LENGTH];
-        pk.pubkey.serialize(&mut bytes[..]).unwrap();
+        s.pubkey
+            .as_ref()
+            .into_affine()
+            .serialize(&mut bytes[..])
+            .unwrap();
         BLS12377PublicKeyBytes::new(bytes)
     }
 }
--- a/crypto/src/bls12377/mod.rs
+++ b/crypto/src/bls12377/mod.rs
@@ -259,7 +259,7 @@ impl<'a> From<&'a BLS12377PrivateKey> for BLS12377PublicKey {
 impl VerifyingKey for BLS12377PublicKey {
     type PrivKey = BLS12377PrivateKey;
     type Sig = BLS12377Signature;
-    type Bytes = PublicKeyBytes<BLS12377PublicKey, { CELO_BLS_PUBLIC_KEY_LENGTH }>;
+    type Bytes = BLS12377PublicKeyBytes;
     const LENGTH: usize = CELO_BLS_PUBLIC_KEY_LENGTH;
 
     fn verify_batch(msg: &[u8], pks: &[Self], sigs: &[Self::Sig]) -> Result<(), signature::Error> {
--- a/crypto/src/bls12381.rs
+++ b/crypto/src/bls12381.rs
@@ -182,7 +182,7 @@ impl<'a> From<&'a BLS12381PrivateKey> for BLS12381PublicKey {
 impl VerifyingKey for BLS12381PublicKey {
     type PrivKey = BLS12381PrivateKey;
     type Sig = BLS12381Signature;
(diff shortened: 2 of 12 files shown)


crypto/src/ed25519.rs
```diff
@@ -339,7 +341,7 @@ impl KeyPair for Ed25519KeyPair {
     fn copy(&self) -> Self {
         Self {
             name: Ed25519PublicKey::from_bytes(self.name.as_ref()).unwrap(),
-            secret: Ed25519PrivateKey::from_bytes(self.name.as_ref()).unwrap(),
+            secret: Ed25519PrivateKey::from_bytes(self.secret.as_ref()).unwrap(),
```
punwai, line 344: well, oops.

crypto/src/bls12377/mod.rs
```diff
@@ -42,7 +42,7 @@ pub struct BLS12377PublicKey {
     pub bytes: OnceCell<[u8; CELO_BLS_PUBLIC_KEY_LENGTH]>,
 }
 
-pub type BLS12377PublicKeyBytes = PublicKeyBytes<BLS12377PublicKey, { BLS12377PublicKey::LENGTH }>;
+pub type BLS12377PublicKeyBytes = PublicKeyBytes<BLS12377PublicKey, { CELO_BLS_PUBLIC_KEY_LENGTH }>;
```
kchalkias, line 45: We should avoid including project names in fixed variable names. Let's keep it as it was before.

crypto/src/bls12381.rs
```diff
@@ -557,18 +558,18 @@ impl AggregateAuthenticator for BLS12381AggregateSignature {
...
+        BLS12381PublicKey::from_bytes(bytes.as_ref()).map_err(|_| Self::Error::new())
     }
 }
 
-impl From<BLS12381PublicKey> for BLS12381PublicKeyBytes {
-    fn from(pk: BLS12381PublicKey) -> BLS12381PublicKeyBytes {
-        BLS12381PublicKeyBytes::new(pk.pubkey.to_bytes())
+impl From<&BLS12381PublicKey> for BLS12381PublicKeyBytes {
+    fn from(s: &BLS12381PublicKey) -> BLS12381PublicKeyBytes {
+        BLS12381PublicKeyBytes::new(s.pubkey.to_bytes())
```
huitseeker, line 573: I know this is not your change but we should not use `to_bytes()` when we already go through pains making sure `BLS12381PublicKey` contains the memoized representation of the bytes.

crypto/src/bls12381.rs
```diff
@@ -557,18 +558,18 @@ impl AggregateAuthenticator for BLS12381AggregateSignature {
...
     type Error = signature::Error;
 
-    fn try_into(self) -> Result<BLS12381PublicKey, Self::Error> {
+    fn try_from(bytes: BLS12381PublicKeyBytes) -> Result<BLS12381PublicKey, Self::Error> {
         // TODO(https://github.com/MystenLabs/sui/issues/101): Do better key validation
         // to ensure the bytes represent a poin on the curve.
-        BLS12381PublicKey::from_bytes(self.as_ref()).map_err(|_| Self::Error::new())
+        BLS12381PublicKey::from_bytes(bytes.as_ref()).map_err(|_| Self::Error::new())
     }
 }
```
huitseeker, line 569: The comment should really be in the `from_bytes` implementation for `BLS12381PublicKey`

crypto/src/bls12381.rs
```diff
@@ -43,7 +43,7 @@ pub struct BLS12381PublicKey {
     pub bytes: OnceCell<[u8; BLS_PUBLIC_KEY_LENGTH]>,
 }
 
-pub type BLS12381PublicKeyBytes = PublicKeyBytes<BLS12381PublicKey, { BLS12381PublicKey::LENGTH }>;
```
huitseeker, line 46: This should be unneeded.

crypto/src/ed25519.rs
```diff
@@ -26,7 +26,8 @@ use crate::{
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Ed25519PublicKey(pub ed25519_dalek::PublicKey);
 
-pub type Ed25519PublicKeyBytes = PublicKeyBytes<Ed25519PublicKey, { Ed25519PublicKey::LENGTH }>;
+pub type Ed25519PublicKeyBytes =
```
huitseeker, line 29: Unneeded.

crypto/src/ed25519.rs
```diff
@@ -64,6 +65,7 @@ impl<'a> From<&'a Ed25519PrivateKey> for Ed25519PublicKey {
 impl VerifyingKey for Ed25519PublicKey {
     type PrivKey = Ed25519PrivateKey;
     type Sig = Ed25519Signature;
+    type Bytes = Ed25519PublicKeyBytes;
```
huitseeker, line 68: Why?

crypto/src/ed25519.rs
huitseeker, line 344: Yes ! There seems to be a test missing in there.

crypto/src/ed25519.rs
```diff
@@ -387,18 +389,19 @@ impl Signer<Ed25519Signature> for Ed25519KeyPair {
...
+impl TryFrom<Ed25519PublicKeyBytes> for Ed25519PublicKey {
     type Error = signature::Error;
 
-    fn try_into(self) -> Result<Ed25519PublicKey, Self::Error> {
+    fn try_from(bytes: Ed25519PublicKeyBytes) -> Result<Ed25519PublicKey, Self::Error> {
         // TODO(https://github.com/MystenLabs/sui/issues/101): Do better key validation
         // to ensure the bytes represent a poin on the curve.
-        Ed25519PublicKey::from_bytes(self.as_ref()).map_err(|_| Self::Error::new())
+        Ed25519PublicKey::from_bytes(bytes.as_ref()).map_err(|_| Self::Error::new())
     }
```
huitseeker, line 400: Same: the comment should probably not be here.

crypto/src/traits.rs
```diff
@@ -185,3 +191,20 @@ pub trait AggregateAuthenticator:
...
+    + Debug
+    + PartialEq
+    + Eq
+    + Hash
+    + PartialOrd
+    + Ord
+    + ToFromBytes
+    + FromStr
+{
+}
```
huitseeker, line 210: Why? Can't you just derive most, if not all of these for `PublicKeyBytes`, which is essentially the only implementation we need?

crypto/src/bls12377/mod.rs
```diff
@@ -539,20 +540,24 @@ impl AggregateAuthenticator for BLS12377AggregateSignature {
...
+impl From<&BLS12377PublicKey> for BLS12377PublicKeyBytes {
+    fn from(s: &BLS12377PublicKey) -> Self {
         let mut bytes = [0u8; CELO_BLS_PUBLIC_KEY_LENGTH];
-        pk.pubkey.serialize(&mut bytes[..]).unwrap();
+        s.pubkey
+            .as_ref()
+            .into_affine()
+            .serialize(&mut bytes[..])
+            .unwrap();
         BLS12377PublicKeyBytes::new(bytes)
```
huitseeker, line 561: We have a memoized representation of the bytes in the `BLS12377PublicKey`. Let's use it, rather than do any of these conversions.

crypto/src/ed25519.rs
punwai: I think one thing is that without defining this, we cannot map a PK <-> it's corresponding PK bytes form within the VerifyingKeyTrait. One thing I added in this PR is the trait bound `TryFrom<Self::Bytes>` on VerifyingKeyBytes - I have tried other ways but I personally don't understand how to define this without an associated type.

punwai: I personally think this trait is very useful, but I could be wrong

crypto/src/ed25519.rs
punwai: You're absolutely correct! Although Tasos has kindly added it in another PR.

crypto/src/bls12377/mod.rs
```diff
@@ -539,20 +540,29 @@ impl AggregateAuthenticator for BLS12377AggregateSignature {
...
+            .bytes
+            .get_or_try_init::<_, eyre::Report>(|| {
+                let mut bytes = [0u8; CELO_BLS_PUBLIC_KEY_LENGTH];
+                pk.pubkey
+                    .as_ref()
+                    .into_affine()
+                    .serialize(&mut bytes[..])
+                    .unwrap();
+                Ok(bytes)
+            })
```
akichidis, line 564: The initialisation here seems to be the same we do on other parts as well (ex [here](https://github.com/MystenLabs/narwhal/pull/507/files#diff-ff9b43e2e185ece3a384a642ac5ba4a67e9b37f000eb819dc82c0a752b3fa23fR174) ) . I would recommend either:
1. Extracting in a method under the `BLS12377PublicKey` 
2. Initialise the `bytes` (if not vastly costly) the same time when we do set the `pubkey`

having duplicate code might lead to diverge implementation in the future

crypto/src/pubkey_bytes.rs
```diff
@@ -67,18 +67,31 @@ impl<T, const N: usize> Default for PublicKeyBytes<T, N> {
...
 }
 
+impl<T: VerifyingKey, const N: usize> FromStr for PublicKeyBytes<T, N> {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s = s.strip_prefix("0x").unwrap_or(s);
+        let value = hex::decode(s)?;
+        Self::from_bytes(&value[..]).map_err(|_| anyhow::anyhow!("byte deserialization failed"))
+    }
```
akichidis, line 77: Is there a test already covering this ? If not then I would suggest adding one

crypto/src/pubkey_bytes.rs
```diff
@@ -67,18 +67,31 @@ impl<T, const N: usize> Default for PublicKeyBytes<T, N> {
...
+}
+
+impl<T: VerifyingKey, const N: usize> Copy for PublicKeyBytes<T, N> {}
```
akichidis, line 80: is the rationale behind this that we want to over the Copy capabilities only for the structs that extend the `VerifyingKey`? If yes, what's the reason ?

crypto/src/bls12377/mod.rs
huitseeker: You can write this as 
```rust
fn from(pk: &BLS12377PublicKey) -> Self {
            BLS12377PublicKeyBytes::new(pk.as_ref().clone())
}
```

crypto/src/bls12381.rs
```diff
@@ -557,18 +558,21 @@ impl AggregateAuthenticator for BLS12381AggregateSignature {
...
-    fn from(pk: BLS12381PublicKey) -> BLS12381PublicKeyBytes {
-        BLS12381PublicKeyBytes::new(pk.pubkey.to_bytes())
+impl From<&BLS12381PublicKey> for BLS12381PublicKeyBytes {
+    fn from(s: &BLS12381PublicKey) -> BLS12381PublicKeyBytes {
+        BLS12381PublicKeyBytes::new(
+            s.bytes
+                .get_or_try_init::<_, eyre::Report>(|| Ok(s.pubkey.to_bytes()))
+                .expect("OnceCell invariant violated")
+                .clone()
+        )
```
huitseeker, line 576:
```suggestion
    fn from(s: &BLS12381PublicKey) -> BLS12381PublicKeyBytes {
        BLS12381PublicKeyBytes::new(s.as_ref().clone())
```

crypto/src/ed25519.rs
huitseeker: > without defining this, we cannot map a PK <-> it's corresponding PK bytes form within the VerifyingKeyTrait

OK, why do you want to do this?

> we can't define functions that takes in some generic PublicKey, and also interacts with the PublicKeyBytes

Do you have an example of such a function?

huitseeker (dismissed):

punwai:
@huitseeker Right now, it is just this function:


#[derive(Default)]
pub struct VerificationObligation<S>
where
    S: AggregateAuthenticator,
{
    lookup: PubKeyLookup<S::PubKey>,
    pub messages: Vec<Vec<u8>>,
    pub signatures: Vec<S>,
    pub public_keys: Vec<Vec<S::PubKey>>,
}

impl<S: AggregateAuthenticator> VerificationObligation<S> {
    pub fn new(lookup: PubKeyLookup<S::PubKey>) -> VerificationObligation<S> {
        VerificationObligation {
            lookup,
            ..Default::default()
        }
    }

    pub fn lookup_public_key(
        &mut self,
        key_bytes: &<<S as AggregateAuthenticator>::PubKey as VerifyingKey>::Bytes,
    ) -> Result<S::PubKey, SuiError> {
        match self.lookup.get(key_bytes) {
            Some(v) => Ok(v.clone()),
            None => {
                let public_key: S::PubKey = (*key_bytes)
                    .try_into()
                    .map_err(|_| SuiError::UnknownSigner)?;
                self.lookup.insert(*key_bytes, public_key.clone());
                Ok(public_key)
            }
        }
    }

    /// Add a new message to the list of messages to be verified.
    /// Returns the index of the message.
    pub fn add_message(&mut self, message: Vec<u8>) -> usize {
        self.signatures.push(S::default());
        self.public_keys.push(Vec::new());
        self.messages.push(message);
        self.messages.len() - 1
    }

    pub fn verify_all(self) -> SuiResult<PubKeyLookup<S::PubKey>> {
        S::batch_verify(
            &self.signatures[..],
            &self.public_keys.iter().map(|x| &x[..]).collect::<Vec<_>>(),
            &self.messages.iter().map(|x| &x[..]).collect::<Vec<_>>()[..],
        )
        .map_err(|error| SuiError::InvalidSignature {
            error: format!("{error}"),
        })?;

        Ok(self.lookup)
    }
}

crypto/src/pubkey_bytes.rs
punwai: I will improve test coverage on PublicKeyBytes this PR as well.

punwai:
I will try removing this trait and reimplementing narwhal.

crypto/src/traits.rs
punwai: All of these are derived anyways, it's just that I needed the bounds for something earlier - anyways I'll remove the function that requires this for now.

crypto/src/bls12377/mod.rs
huitseeker:
```suggestion
pub type BLS12377PublicKeyBytes = PublicKeyBytes<BLS12377PublicKey, { BLS12377PublicKey::LENGTH }>;
```

crypto/src/ed25519.rs
```diff
@@ -26,7 +26,8 @@ use crate::{
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Ed25519PublicKey(pub ed25519_dalek::PublicKey);
 
-pub type Ed25519PublicKeyBytes = PublicKeyBytes<Ed25519PublicKey, { Ed25519PublicKey::LENGTH }>;
+pub type Ed25519PublicKeyBytes =
+    PublicKeyBytes<Ed25519PublicKey, { ed25519_dalek::PUBLIC_KEY_LENGTH }>;
```
huitseeker, line 30:
```suggestion
pub type Ed25519PublicKeyBytes =
    PublicKeyBytes<Ed25519PublicKey, { Ed25519PublicKey::LENGTH }>;
```

crypto/src/bls12381.rs
```diff
@@ -43,7 +43,7 @@ pub struct BLS12381PublicKey {
     pub bytes: OnceCell<[u8; BLS_PUBLIC_KEY_LENGTH]>,
 }
 
-pub type BLS12381PublicKeyBytes = PublicKeyBytes<BLS12381PublicKey, { BLS12381PublicKey::LENGTH }>;
+pub type BLS12381PublicKeyBytes = PublicKeyBytes<BLS12381PublicKey, BLS_PUBLIC_KEY_LENGTH>;
```
huitseeker, line 46:
```suggestion
pub type BLS12381PublicKeyBytes = PublicKeyBytes<BLS12381PublicKey, BLS12381PublicKey::LENGTH>;
```

crypto/src/pubkey_bytes.rs
```diff
@@ -67,18 +67,27 @@ impl<T, const N: usize> Default for PublicKeyBytes<T, N> {
...
+        let s = s.strip_prefix("0x").unwrap_or(s);
+        let value = hex::decode(s)?;
+        Self::from_bytes(&value[..]).map_err(|_| anyhow::anyhow!("byte deserialization failed"))
+    }
+}
+
 // This guarantees the security of the constructor of a `PublicKeyBytes` instance
 // TODO: replace this clunky sealed marker trait once feature(associated_const_equality) stabilizes
-mod sealed {
+pub mod sealed {
```
huitseeker, line 82: The whole point is that this module must be private. 
https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed

huitseeker (commented):
This is getting really close to done!

punwai:
@huitseeker Changes made

crypto/src/pubkey_bytes.rs
huitseeker: This still needs to be turned back to private.

kchalkias (commented):
The PR is in a great state and ready to be landed after we rebase and resolve conflicts. Ping me when done and will approve

kchalkias (approved):
Awesome, LGTM!

crypto/src/pubkey_bytes.rs
punwai: Sorry for responding to this late, but a problem was that type T could possibly not have the Copy trait, which will make the compiler throw an error.

PR #526, opened 2022-07-19
feat: update the mysten-infra pointer & switch to telemetry-subscribers

huitseeker:
Advantage: we get the dynamically resettable log level + the same infra as everyone else + once we integrate [tracing_loki](https://github.com/hrxi/tracing-loki) we get that in Narwhal staging deployments as well.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -22,7 +22,10 @@ use node::{
 };
 use std::sync::Arc;
 use tokio::sync::mpsc::{channel, Receiver};
-use tracing::{info, subscriber::set_global_default};
+use tracing::info;
+#[cfg(feature = "benchmark")]
+use tracing::subscriber::set_global_default;
+#[cfg(feature = "benchmark")]
 use tracing_subscriber::filter::{EnvFilter, LevelFilter};
 
 #[tokio::main]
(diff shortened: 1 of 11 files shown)


node/src/main.rs
```diff
@@ -74,34 +78,36 @@ async fn main() -> Result<()> {
...
+            let env_filter = EnvFilter::try_from_default_env().unwrap_or(filter);
 
-    let env_filter = EnvFilter::try_from_default_env().unwrap_or(filter);
-    cfg_if::cfg_if! {
-        if #[cfg(feature = "benchmark")] {
             let timer = tracing_subscriber::fmt::time::UtcTime::rfc_3339();
             let subscriber_builder = tracing_subscriber::fmt::Subscriber::builder()
-                .with_env_filter(env_filter)
-                .with_timer(timer).with_ansi(false);
+            .with_env_filter(env_filter)
```
velvia, line 95: I think you want the indent here right?

velvia, line 95: I feel like we should be able to easily adapt TelemetrySubscribers to accommodate this benchmarking use case as well.  Maybe just add a flag to turn off ansi or whatever else is needed.

lanvidr (approved):
So we change the environment variable to update the log level? How long does it take for the changes to be in effect?

velvia (approved):
Awesome!  Was about to do this.

huitseeker:
@laura-makdah We need a bit of infra to add dynamically-resettable log levels on top of this, see https://github.com/MystenLabs/sui/pull/3269

akichidis (commented):
Thanks a lot @huitseeker for doing this!

Issue #529, opened 2022-07-19
[proto] Batch exposed in `CollectionRetrievalResult`


huitseeker:
Our `CollectionRetrievalResult` nests transactions in what appears to be a single batch. This batch is ephemeral and reflects collecting the transactions at the primary, rather than our actual notion of batch. We should either remove this wrapper, or name it in a disambiguated way.

https://github.com/MystenLabs/narwhal/blob/c312103d46189052f5f6a12ca32c233759d7ea21/types/proto/narwhal.proto#L59-L64

/cc @piersy

akichidis:
@arun-koshy @huitseeker this is what I am proposing for the `GetCollections` endpoint. I also took into account the complexity that the current compiler is introducing on the clients when nesting `Result` is used and I believe we can probably make things a bit simpler by splitting into two parts:
1. the `collections` that have only the successfully retrieved collections
2. the `errors` field that contains the errors for the collections we didn't manage to retrieve

I've also removed entirely the concept of batches as for the 99% of the cases this won't be needed here. Even if we need that information (how transactions are grouped) this will be needed only for verification processes and we can always extend the `Collection` message with an extra field providing that info.

```
message GetCollectionsResponse {
    // List of successfully retrieved collections.
    repeated Collection collections = 1;
    
    // If failed to retrieve a collection then this will be added here
    repeated CollectionError errors = 2;
}

message Collection {
    // The collection's id
    CertificateDigest id = 1;

    // The transactions that are part of the collection
    repeated Transaction transactions = 2;
}

message Transaction {
    bytes transaction = 1;
}

message CollectionError {
    // The collection for which this error refers to
    CertificateDigest id = 1;

    enum CollectionErrorType {
        COLLECTION_NOT_FOUND = 0;
        COLLECTION_TIMEOUT = 1;
        COLLECTION_ERROR = 2;
    }
    CollectionErrorType type = 2;
}
```

arun-koshy:
Everything looks good to me the only thing I am curious about is your switch to separate fields instead of a nested repeated `oneof`

> the complexity that the current compiler is introducing on the clients when nesting Result
Can you explain this further?

akichidis:
> Everything looks good to me the only thing I am curious about is your switch to separate fields instead of a nested repeated `oneof`
> 
> > the complexity that the current compiler is introducing on the clients when nesting Result
> 
> Can you explain this further?

[...]

PR #536, opened 2022-07-20
Rename CollectionRetrievalResult field 'batch' to 'transaction_list'

arun-koshy:
Fixes [Issue#529](https://github.com/MystenLabs/narwhal/issues/529)

### DIFF
--- a/examples/protos/narwhal.proto
+++ b/examples/protos/narwhal.proto
@@ -58,7 +58,7 @@ message ValidatorData {
 
 message CollectionRetrievalResult {
     oneof retrieval_result {
-        BatchMessage batch = 1;
+        BatchMessage transaction_list = 1;
         CollectionError error = 2;
     }
 }
--- a/examples/protos/narwhal.proto
+++ b/examples/protos/narwhal.proto
@@ -9,32 +9,20 @@ message CertificateDigest {
     bytes digest = 1;
 }
 
-message BatchDigest {
-    bytes digest = 1;
-}
-
-message Batch {
-    repeated Transaction transaction = 1;
-}
-
 message Transaction {
     bytes transaction = 1;
 }
 
-enum CollectionErrorType {
-    COLLECTION_NOT_FOUND = 0;
-    COLLECTION_TIMEOUT = 1;
-    COLLECTION_ERROR = 2;
-}
-
 message CollectionError {
+    // The collection for which this error refers to
     CertificateDigest id = 1;
-    CollectionErrorType error = 2;
-}
 
-message BatchMessage {
-    BatchDigest id = 1;
-    Batch transactions = 2;
+    enum CollectionErrorType {
+        COLLECTION_NOT_FOUND = 0;
+        COLLECTION_TIMEOUT = 1;
+        COLLECTION_ERROR = 2;
+    }
+    CollectionErrorType error = 2;
 }
 
 message PrimaryAddresses {
@@ -56,15 +44,23 @@ message ValidatorData {
     PrimaryAddresses primary_addresses = 3;
 }
 
+message Collection {
+    // The collection's id
+    CertificateDigest id = 1;
+
+    // The transactions that are part of the collection
+    repeated Transaction transactions = 2;
+}
+
 message CollectionRetrievalResult {
     oneof retrieval_result {
-        BatchMessage transaction_list = 1;
+        Collection collection = 1;
         CollectionError error = 2;
     }
 }
 
 message GetCollectionsRequest {
-    // List of collections to be retreived.
+    // List of collections to be retrieved.
     repeated CertificateDigest collection_ids = 1;
 }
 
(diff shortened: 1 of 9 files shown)


akichidis (commented):
I would recommend to change the response structure on the `CollectionRetrievalResult` so we align it with the inner concept of batches & transactions. I think mixing the concept of Batches with transactions might be a bit more confusing. To recap our Block / Collection is comprised of multiple batches and each batch of multiple transactions. So a Collection is something like:

```
Collection
|---> Batch 
|         |---> transaction
|         |---> transaction
|         |...........
|----> Batch
|         |----> transaction
|         |----> transaction
......
```

So this is what I would suggest to forward with our API as well. Since in our response we are returning multiple collections then we probably want to return a vector of collections. So a `Vec<Collection>` . @arun-koshy thoughts?

examples/protos/narwhal.proto
```diff
@@ -58,7 +58,7 @@ message ValidatorData {
 
 message CollectionRetrievalResult {
     oneof retrieval_result {
-        BatchMessage batch = 1;
+        BatchMessage transaction_list = 1;
```
huitseeker, line 61: The problem here is not so much the name 'batch', it's the reuse of "BatchMessage". I.e. this still makes folks believe we are reusing the Batch concept to convey this transaction list. I think it's better to use a different type.

primary/tests/integration_tests_validator_api.rs
```diff
@@ -925,15 +928,18 @@ async fn test_get_collections_with_missing_certificates() {
...
     );
 
     for result in actual_result {
         match result.retrieval_result.unwrap() {
-            RetrievalResult::Batch(batch) => {
-                let id: BatchDigest = batch.id.unwrap().into();
-                let result_batch: Batch = batch.transactions.unwrap().into();
+            RetrievalResult::TransactionList(transaction_list) => {
+                let id: BatchDigest = transaction_list.id.unwrap().into();
+                let result_batch: Batch = transaction_list.transactions.unwrap().into();
```
huitseeker, line 942: This use of `BatchDigest` and `Batch` is the one that is illegitimate: the digest in here is a `CertificateDigest`, and the transaction list itself should have its own type.

huitseeker (commented):
I think there are two topics:
1. conceptually, do we expose the internal batching structure when we retrieve a collection? That's certainly something we can revisit, but if so, perhaps not in PR comments.
2. the actual types we expose at the moment in a `CollectionRetrievalResult` should have nothing to do with a `BatchDigest` (they are not comparable to any `BatchDigest` in the system) or with a `BatchMessage` (the are the concatenation of several batches, and thus not a batch).

I've left comments in this direction in the PR.

arun-koshy:
Just to clarify `BlockWaiter` returns 

```
pub struct GetBlocksResponse {
    pub blocks: Vec<BlockResult<GetBlockResponse>>,
}
```

```
pub struct GetBlockResponse {
    pub id: CertificateDigest,
    pub batches: Vec<BatchMessage>,
}
```

```
pub struct BatchMessage {
    // TODO: revisit including the id here [see #188]
    pub id: BatchDigest,
    pub transactions: Batch,
}
```

```
pub struct Batch(pub Vec<Transaction>);
```

```
pub type Transaction = Vec<u8>;
```

Currently we are just placing the `batches` in `GetBlockResponse` into a proto as follows

```
message CollectionRetrievalResult {
    oneof retrieval_result {
        BatchMessage batch = 1;
        CollectionError error = 2;
    }
}
```

```
message BatchMessage {
    BatchDigest id = 1;
    Batch transactions = 2;
}
```

```
message Batch {
    repeated Transaction transaction = 1;
}
```

```
message Transaction {
    bytes transaction = 1;
}
```

But if I am understanding @huitseeker and @akichidis correctly we want to wrap the batches in the original `Collection` and remove BatchMessage to simplify the type tree i.e.

```
message CollectionRetrievalResult {
    oneof retrieval_result {
        Collection collection = 1;
        CollectionError error = 2;
    }
}
```

```
message Collection {
    CertificateDigest id = 1;
    repeated Batch batches = 2;
}
```

```
message Batch {
    BatchDigest id = 1;
    repeated Transaction transactions = 2;
}
```

```
message Transaction {
    bytes transaction = 1;
}
```

primary/tests/integration_tests_validator_api.rs
arun-koshy: I actually threw away the `CertificateDigest` in the original implementation and only returned the `BatchDigest` and `Batch`. Explained further below.

akichidis:
> I think there are two topics:
> 
> 1. conceptually, do we expose the internal batching structure when we retrieve a collection? That's certainly something we can revisit, but if so, perhaps not in PR comments.
> 2. the actual types we expose at the moment in a `CollectionRetrievalResult` should have nothing to do with a `BatchDigest` (they are not comparable to any `BatchDigest` in the system) or with a `BatchMessage` (the are the concatenation of several batches, and thus not a batch).
> 
> I've left comments in this direction in the PR.

@huitseeker regarding topic (1) yes I believe we should revisit this endpoint. More specifically in the aspects of:
* returned items / structure
* errors
* naming

Thus I don't recommend doing any renaming/changes and redesign it. I agree, let's not use the PR, we can setup a discussion for this.

akichidis (approved):
That looks great @arun-koshy 💯  thanks so much!

PR #541, opened 2022-07-20
crypto: update k256 and signature

joyqvq:


crypto/src/secp256k1.rs
```diff
@@ -279,8 +280,8 @@ impl KeyPair for Secp256k1KeyPair {
         self.secret
     }
 
-    fn generate<R: rand::CryptoRng + rand::RngCore>(rng: &mut R) -> Self {
-        let privkey = k256::ecdsa::SigningKey::random(rng);
+    fn generate<R: rand::CryptoRng + rand::RngCore>(_rng: &mut R) -> Self {
+        let privkey = k256::ecdsa::SigningKey::random(&mut OsRng);
```
joyqvq, line 284: workaround here: instead of using what R is passed in (the version for rand is different from what k256 library likes to take in), generate a fresh OsRng that complies to the k256 interface

crypto/src/secp256k1.rs
```diff
@@ -78,7 +79,7 @@ impl Verifier<Secp256k1Signature> for Secp256k1PublicKey {
 impl AsRef<[u8]> for Secp256k1PublicKey {
     fn as_ref(&self) -> &[u8] {
         self.bytes
-            .get_or_try_init::<_, eyre::Report>(|| Ok(self.pubkey.to_bytes()))
```
kchalkias, line 81: qq: why not use .bytes directly (that we have from OnceCell)?

joyqvq: this `bytes` OnceCell may not be populated, `get_or_try_init` tries to retrieve it by getting `bytes` directly, otherwise it will try populate it 

get_or_try_init:
```
Gets the contents of the cell, initializing it with f if the cell was empty. If the cell was empty and f failed, an error is returned.
```

crypto/src/secp256k1.rs
huitseeker: Yep. Please add a TODO to:
1. clean this hack up once we upgrade our eddsa implementation,
2. NOT use this in tests: we have several fixtures that generate / copy the same keys in tests repeatedly by passing in carefully-seeded RNGs (see `keys()` in `test_utils`).

crypto/src/secp256k1.rs
```diff
@@ -327,7 +335,13 @@ impl TryInto<Secp256k1PublicKey> for Secp256k1PublicKeyBytes {
 
 impl From<Secp256k1PublicKey> for Secp256k1PublicKeyBytes {
     fn from(pk: Secp256k1PublicKey) -> Secp256k1PublicKeyBytes {
-        Secp256k1PublicKeyBytes::new(pk.pubkey.to_bytes())
+        Secp256k1PublicKeyBytes::new(
```
kchalkias, line 338: sorry was referring to this line re oncecell, to avoid duplicated code with my previous comment

huitseeker (approved):

crypto/src/secp256k1.rs
huitseeker: Yep, on this one, we should use `as_ref()`

kchalkias (approved):
Awesome work!

PR #559, opened 2022-07-22
[network] Don't back off forever on the Semaphore 

huitseeker:
## Context
#463 introduced a `BoundedExecutor` to limit the concurrent tasks created for sending messages.

However, as a protocol requirement, some of those tasks are on an unbounded exponential backoff, by which they are retried indefinitely (until cancelled).

## The issue
The {primary, worker} × { `broadcast`, `send` } functions enqueue instances of `{Primary, Worker}::send_message` on the `BoundedExecutor`, and `send_message` is an exponential backoff with infinite retries. These tasks will hence hold a semaphore ticket potentially forever.

## This change
The present change makes those tasks hold a semaphore permit only when they are actively sending a message, re-queuing on the semaphore only once they are done with their backoff wait (out of the semaphore0. This frees up tickets for other network tasks to help make the network progress.

## Future work
The cancellation of our "reliable" network sends is not always as prompt as it could be. Fixing this is the next PR.

### DIFF
--- a/network/src/bounded_executor.rs
+++ b/network/src/bounded_executor.rs
@@ -62,12 +62,12 @@ impl BoundedExecutor {
 
     // Acquires a permit with the semaphore, first gracefully,
     // then queuing after logging that we're out of capacity.
-    async fn acquire_permit(&self) -> OwnedSemaphorePermit {
-        match self.semaphore.clone().try_acquire_owned() {
+    async fn acquire_permit(semaphore: Arc<Semaphore>) -> OwnedSemaphorePermit {
+        match semaphore.clone().try_acquire_owned() {
             Ok(p) => p,
             Err(_) => {
                 info!("concurrent task limit reached, waiting...");
-                self.semaphore.clone().acquire_owned().await.unwrap()
+                semaphore.acquire_owned().await.unwrap()
             }
         }
     }
@@ -81,7 +81,7 @@ impl BoundedExecutor {
         F: Future + Send + 'static,
         F::Output: Send + 'static,
     {
-        let permit = self.acquire_permit().await;
+        let permit = Self::acquire_permit(self.semaphore.clone()).await;
 
         self.spawn_with_permit(f, permit)
     }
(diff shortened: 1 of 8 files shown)


network/src/primary.rs
```diff
@@ -72,21 +71,25 @@ impl PrimaryNetwork {
         &mut self,
         address: Multiaddr,
         message: BincodeEncodedPayload,
-    ) -> CancelHandler<()> {
+    ) -> CancelHandler<MessageResult> {
         let client = self.client(address);
+
+        let message_send = move || {
+            let mut client = client.clone();
+            let message = message.clone();
```
mystenmark, line 79: surely this is not deep-copying the entire payload every time... right?

bmwill: As of today the message type is essentially a Bytes struct which has a very efficient clone operation (its essentially just an atomic increment)

huitseeker: The PR does not change the cloning behavior, but FWIW the client also has a cheap cloning operation.

network/src/bounded_executor.rs
```diff
@@ -81,24 +112,107 @@ impl BoundedExecutor {
...
+            let semaphore = self.semaphore.clone();
+
+            let executor = move || {
+                let semaphore = semaphore.clone();
+                BoundedExecutor::run_on_semaphore(semaphore, f())
+            };
+
+            retry_config.retry(executor)
+        };
+        self.executor.spawn(retrier())
```
bmwill, line 159: This function does not do what the documentation states it does. In particular this function, while marked as async, does not contain any awaits and will unconditionally spawn a future onto the runtime, even if we're at capacity.

huitseeker: It will indeed unconditionally spawn a task on the runtime that queues a sequence of sends, one after the other, each of which are semaphore-bound.

What else would you like to have happen?

bmwill: I would like the doc comment on the function to correctly describe what is going on in this function. Also given the current structure it could be made a non-async function

gdanezis: I also have trouble understanding this function.

My understanding is that `self.executor.spawn(retrier())` will do the following:
- it will first execute the closure above it. So this does not pass the closure as an argument but its value after execution.
- then passes the return value of the closure to the `self.executor.spawn` function (which I have to say I am not entirely familiar with).

network/src/worker.rs
```diff
@@ -69,21 +68,25 @@ impl WorkerNetwork {
...
-                    .retry(move || {
-                        let mut client = client.clone();
-                        let message = message.clone();
-                        async move { client.send_message(message).await.map_err(Into::into) }
-                    })
-                    .map(|response| {
-                        response.expect("we retry forever so this shouldn't fail");
-                    }),
-            )
+            .spawn_with_retries(self.retry_config, message_send)
```
gdanezis, line 89: This does create more tasks right?

network/src/bounded_executor.rs
```diff
@@ -81,24 +112,107 @@ impl BoundedExecutor {
...
+    }
+
+    // Equips a future with a final step that drops the held semaphore permit
```
akichidis, line 162: Since we have the `.await` in the end, it equips and executes the future, right?

network/src/bounded_executor.rs
mystenmark: it executes the closure, which returns a future to be executed - you can't restart a future from the beginning so this is necessary

network/src/worker.rs
huitseeker: Overall, yes. But it does not create more concurrent network messages.

network/src/bounded_executor.rs
huitseeker: I've removed the useless eta-expansion (which was just useful for me to figure out lifetimes & scoping)

network/src/bounded_executor.rs
huitseeker: This is an async function, so no execution here.

bmwill (approved):

Issue #534, opened 2022-07-20
Add proptest to secp256k1_test.rs


joyqvq:
Add a proptest to compare result against the signature generated by https://docs.rs/secp256k1/latest/secp256k1/

PR #562, opened 2022-07-22
crypto: add proptest against k256 with ffi wrapper impl

joyqvq:
https://github.com/MystenLabs/narwhal/issues/534

### DIFF
--- a/crypto/src/tests/secp256k1_tests.rs
+++ b/crypto/src/tests/secp256k1_tests.rs
@@ -145,3 +145,36 @@ async fn signature_service() {
     // Verify the signature we received.
     assert!(pk.verify(digest.as_ref(), &signature).is_ok());
 }
+
+use proptest::arbitrary::Arbitrary;
+
+proptest::proptest! {
+    #[test]
+    fn test_k256_against_secp256k1_lib(
+        r in <[u8; 32]>::arbitrary()
+    ) {
+        let message: &[u8] = b"Hello, world!";
+
+        // contruct private key with bytes and signs message
+        let priv_key = <Secp256k1PrivateKey as ToFromBytes>::from_bytes(&r).unwrap();
+        let key_pair = Secp256k1KeyPair::from(priv_key);
+        let key_pair_copied = key_pair.copy();
+        let signature = key_pair.sign(message);
+        assert!(key_pair.public().verify(message, &signature).is_ok());
+
+        // use k256 to construct private key with the same bytes and signs the same message
+        let priv_key_1 = k256::ecdsa::SigningKey::from_bytes(&r).map_err(|_e| signature::Error::new())?;
+        let pub_key_1 = priv_key_1.verifying_key();
+        let signature_1: k256::ecdsa::Signature = priv_key_1.sign(message);
+        assert!(pub_key_1.verify(message, &signature_1).is_ok());
+
+        // same byte string produces same private key
+        assert_eq!(key_pair.private().as_bytes(), priv_key_1.to_bytes().as_slice());
+
+        // same public key is created
+        assert_eq!(key_pair_copied.public().as_bytes(), pub_key_1.to_bytes().as_slice());
+
+        // same signatures produced from both implementations
+        assert_eq!(signature.as_ref(), ToFromBytes::as_bytes(&signature_1));
+   }
+}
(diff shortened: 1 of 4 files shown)


crypto/src/tests/secp256k1_tests.rs
```diff
@@ -145,3 +145,29 @@ async fn signature_service() {
...
+        let key_pair = Secp256k1KeyPair::from(priv_key);
+        let signature = key_pair.sign(&message);
+        assert!(key_pair.public().verify(&message, &signature).is_ok());
+
+        // use k256 to construct private key with the same bytes and signs the same message
+        let priv_key_1 = k256::ecdsa::SigningKey::from_bytes(&r).map_err(|_e| signature::Error::new())?;
+        let pub_key_1 = priv_key_1.verifying_key();
+        let signature_1: k256::ecdsa::Signature = priv_key_1.sign(message);
+        assert!(pub_key_1.verify(message, &signature_1).is_ok());
+
```
huitseeker, line 169: We might be able to check for the equality of the byte representation of the public keys here as well.

huitseeker (commented):
This is great! Thanks @joyqvq !

Here are some of the tests I think could be both quick to write and instructive:
- given a byte string of the correct length, secp256k1 and k256 agree on whether you can create a private key from it (essentially testing the scalar representation checks on deserializaiton),
- given one of the three assets (privkey, pubkey, signature) generated using one of the libraries, you can "serialize" it (to a byte representation) and get it interpreted correctly by the other library (essentially a test of serialization)

crypto/src/tests/secp256k1_tests.rs
```diff
@@ -145,3 +145,36 @@ async fn signature_service() {
...
+        // same byte string produces same private key
+        assert_eq!(key_pair.private().as_bytes(), priv_key_1.to_bytes().as_slice());
+
+        // same public key is created
+        assert_eq!(key_pair_copied.public().as_bytes(), pub_key_1.to_bytes().as_slice());
+
+        // same signatures produced from both implementations
+        assert_eq!(signature.as_ref(), ToFromBytes::as_bytes(&signature_1));
```
kchalkias, line 178: Ooc,ECDSA is a randomised algorithm, how come the signature bytes produced for the same message are the same?
I would expect that the test should be a signature generated by implA should be verified by implB and vice versa.

joyqvq: there is a function for sign with an extra nonce that produces randomized signature. do we want that? in that case the signature would be indeed different. 

```
    pub fn sign_ecdsa(&self, msg: &Message, sk: &SecretKey) -> Signature {
        self.sign_ecdsa_with_noncedata_pointer(msg, sk, ptr::null())
    }

    /// Constructs a signature for `msg` using the secret key `sk` and RFC6979 nonce
    /// and includes 32 bytes of noncedata in the nonce generation via inclusion in
    /// one of the hash operations during nonce generation. This is useful when multiple
    /// signatures are needed for the same Message and SecretKey while still using RFC6979.
    /// Requires a signing-capable context.
    pub fn sign_ecdsa_with_noncedata(
        &self,
        msg: &Message,
        sk: &SecretKey,
        noncedata: &[u8; 32],
    ) -> Signature {
        let noncedata_ptr = noncedata.as_ptr() as *const ffi::types::c_void;
        self.sign_ecdsa_with_noncedata_pointer(msg, sk, noncedata_ptr)
    }
```

similar function in k256:
```
    fn try_sign_with_rng(&self, rng: impl CryptoRng + RngCore, msg: &[u8]) -> Result<S, Error> {
        self.try_sign_digest_with_rng(rng, S::Digest::new_with_prefix(msg))
    }
```

joyqvq: re: ecrecovery, there is a code change to use recoverable signature, i stacked the PR here: https://github.com/MystenLabs/narwhal/pull/564 (still tuning, WIP)

huitseeker: Yay, folks are by default using RFC6979 🎉

kchalkias (approved):
Approving this, and we need a few extra upcoming PRs 
- clear api in our codebase for signing using a) randomized signatures b) deterministic RFC6979, default should be the randomized (just for consistency with most implementations out there and the main ECDSA standard).
- an extra test for randomized sigs (sign with implA, verify with implB and vice versa)
- comparison benchmark and test for ecrecover as well.
- ECDSA-NM (which is already in progress)

PR #564, opened 2022-07-22
crypto: use recoverable signature for secp256k1

joyqvq:
https://github.com/MystenLabs/sui/issues/3270

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -20,7 +20,9 @@ tokio = { version = "1.20.0", features = ["sync", "rt", "macros"] }
 ark-bls12-377 = { version = "0.3.0", features = ["std"], optional = true }
 hkdf = { version = "0.12.3", features = ["std"] }
 serde_with = "1.14.0"
-rust_secp256k1 = { version = "0.24.0", package = "secp256k1", features = ["rand", "bitcoin_hashes"] }
+serde_json = "1.0.80"
+schemars ="0.8.10"
+rust_secp256k1 = { version = "0.24.0", package = "secp256k1", features = ["recovery", "rand-std", "bitcoin_hashes", "global-context"] }
 
 # TODO: switch to https://github.com/celo-org/celo-bls-snark-rs 
 # when https://github.com/celo-org/celo-bls-snark-rs/issues/228 is solved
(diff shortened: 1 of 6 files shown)


network/src/primary.rs
```diff
@@ -72,6 +72,12 @@ impl PrimaryNetwork {
         self.send_message(address, message).await
     }
 
+    // Safety
```
kchalkias, line 75: We don't need in this PR

crypto/src/secp256k1.rs
```diff
@@ -29,11 +29,14 @@ pub struct Secp256k1PrivateKey {
     pub bytes: OnceCell<[u8; constants::SECRET_KEY_SIZE]>,
 }
 
+// Compact signature followed by one extra byte for recover id, used to recover public key from signature.
+pub const RECOVERABLE_SIGNATURE_SIZE: usize = constants::COMPACT_SIGNATURE_SIZE + 1;
```
kchalkias, line 33: thus, 65 bytes! great

kchalkias (commented):
The title of the PR does not match functionality (expecting non-malleable sigs (with reduced s value - but see ecrecover. 
Can we update the title to reflect that?

kchalkias (dismissed):
LGTM

kchalkias (commented):
We need to rebase and fix any failing test

kchalkias (approved):
Thanks @joyqvq. This PR is in a mergeable state now. We might revisit the default hash function (and only use keccak in our tests).

crypto/Cargo.toml
```diff
@@ -20,7 +20,9 @@ tokio = { version = "1.20.0", features = ["sync", "rt", "macros"] }
 ark-bls12-377 = { version = "0.3.0", features = ["std"], optional = true }
 hkdf = { version = "0.12.3", features = ["std"] }
 serde_with = "1.14.0"
-rust_secp256k1 = { version = "0.24.0", package = "secp256k1", features = ["rand", "bitcoin_hashes"] }
+serde_json = "1.0.80"
+schemars ="0.8.10"
```
huitseeker, line 24: Small merge error on this one

PR #583, opened 2022-07-25
fix: fix trait usage in restarter,

### LINKED ISSUE #579: Mop up KeyPair trait usage in restarter
### Description

From @huitseeker 

Mop up the trait required to use `KeyPair`. We should be able to grab the most-constrained trait, i.e. KeyPair, and obtain the rest through projection, i.e. it's easier to do
```rust
foo<Keys: KeyPair>(
    ...
    Bar<Keys::PubKey>
    ...
)
```

rather than
```rust
foo<Keys: KeyPair<PubKey = PublicKey> + Signer<PublicKey::Sig> + Send + 'static, PublicKey: VerifyingKey> (
    ...
   Bar<PublicKey>
    ...
)
```

huitseeker:
Adds a few missing type constraints in the `KeyPair` trait, closes #579.

### DIFF
--- a/crypto/src/traits.rs
+++ b/crypto/src/traits.rs
@@ -145,9 +145,9 @@ pub trait Authenticator:
 pub trait KeyPair:
     Sized + From<Self::PrivKey> + Signer<Self::Sig> + EncodeDecodeBase64 + FromStr
 {
-    type PubKey: VerifyingKey<PrivKey = Self::PrivKey>;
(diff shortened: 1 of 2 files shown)


kchalkias (approved):
Thanks

akichidis (approved):
Awesome! There is probably other places where we can apply this and reduce the amount of type parameters (ex in primary.rs)  - or at least use as practice for new code we add.

PR #584, opened 2022-07-25
feat: add an action for auto-merging dependabot updates w/in a minor range

huitseeker:
Uses https://github.com/ahmadnassri/action-dependabot-auto-merge

(Merging, after review, will need setting up a token)

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,23 @@
+name: auto-merge
+
+on:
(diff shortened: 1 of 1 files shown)


akichidis (approved):

kchalkias (approved):

huitseeker:
Actually realized @laura-makdah may have an idea of whether the current PR is the best approach or if https://github.com/fastify/github-action-merge-dependabot is better.

laura-makdah:
They look pretty similar to me

PR #590, opened 2022-07-25
Add endpoint to retrieve multiaddr of nw primary

arun-koshy:
Add endpoint to expose primary to primary interface. This endpoint will be used to publish multiaddr of the new primary to an announce protocol. 

Endpoint usage as explained by @huitseeker 
> Lifecycle of a NW MultiAddr
> 
> 1. (NW process : own MultiAddr) —- RPC endpoint ——> (Consensus process: own MultiAddr)
> 2. (Consensus process: own MultiAddr). ——- DHT / Announce —-> (DHT : registered MultiAddr for node PK)
> 3.  (DHT : registered MultiAddr for node PK) ——- DHT / Annouce —→ (x size of deployment) (DHT : registered MultiAddr for all nodes in NW)
> 4. (Consensus : registered MultiAddr for all nodes in NW) —— RPC Server ——> (NW process: registered MultAddr for all nodes in NW)
> 
> Step 1. is local RPC, 2. is DHT protocol write 3. is DHT listening (and subsumes a gossip boundary) 4. is local RPC.

### DIFF
--- a/primary/src/grpc_server/configuration.rs
+++ b/primary/src/grpc_server/configuration.rs
@@ -2,18 +2,26 @@
 // SPDX-License-Identifier: Apache-2.0
 use config::{PrimaryAddresses, SharedCommittee};
 use crypto::{traits::ToFromBytes, PublicKey};
+use multiaddr::Multiaddr;
 use std::collections::BTreeMap;
 use tonic::{Request, Response, Status};
-use types::{Configuration, Empty, NewEpochRequest, NewNetworkInfoRequest, PublicKeyProto};
+use types::{
+    Configuration, Empty, GetPrimaryAddressResponse, MultiAddrProto, NewEpochRequest,
+    NewNetworkInfoRequest, PublicKeyProto,
+};
 
 pub struct NarwhalConfiguration {
+    primary_addr: Multiaddr,
     /// The committee
     committee: SharedCommittee,
 }
 
 impl NarwhalConfiguration {
-    pub fn new(committee: SharedCommittee) -> Self {
-        Self { committee }
+    pub fn new(primary_addr: Multiaddr, committee: SharedCommittee) -> Self {
+        Self {
+            primary_addr,
+            committee,
+        }
     }
 
     /// Extracts and verifies the public key provided from the RoundsRequest.
@@ -145,5 +153,16 @@ impl Configuration for NarwhalConfiguration {
         res.map_err(|err| Status::internal(format!("Could not update network info: {:?}", err)))?;
 
         Ok(Response::new(Empty {}))
+    }
+
+    async fn get_primary_address(
+        &self,
+        _request: Request<Empty>,
+    ) -> Result<Response<GetPrimaryAddressResponse>, Status> {
+        Ok(Response::new(GetPrimaryAddressResponse {
+            primary_addr: Some(MultiAddrProto {
+                address: self.primary_addr.to_string(),
+            }),
+        }))
     }
 }
--- a/primary/src/grpc_server/configuration.rs
+++ b/primary/src/grpc_server/configuration.rs
@@ -160,7 +160,7 @@ impl Configuration for NarwhalConfiguration {
         _request: Request<Empty>,
     ) -> Result<Response<GetPrimaryAddressResponse>, Status> {
         Ok(Response::new(GetPrimaryAddressResponse {
-            primary_addr: Some(MultiAddrProto {
+            primary_address: Some(MultiAddrProto {
                 address: self.primary_addr.to_string(),
             }),
         }))
--- a/primary/src/grpc_server/configuration.rs
+++ b/primary/src/grpc_server/configuration.rs
@@ -11,15 +11,15 @@ use types::{
 };
 
 pub struct NarwhalConfiguration {
-    primary_addr: Multiaddr,
+    primary_address: Multiaddr,
     /// The committee
     committee: SharedCommittee,
 }
 
 impl NarwhalConfiguration {
-    pub fn new(primary_addr: Multiaddr, committee: SharedCommittee) -> Self {
+    pub fn new(primary_address: Multiaddr, committee: SharedCommittee) -> Self {
         Self {
-            primary_addr,
+            primary_address,
             committee,
         }
     }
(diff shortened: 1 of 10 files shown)


primary/tests/integration_tests_configuration_api.rs
```diff
@@ -181,6 +182,61 @@ async fn test_new_network_info() {
     assert_eq!(Empty {}, actual_result);
 }
 
+#[tokio::test]
+async fn test_get_primary_address() {
```
akichidis, line 186: I believe you can benefit in this test by [Cluster](https://github.com/MystenLabs/narwhal/blob/main/test_utils/src/cluster.rs). It will allow you to easily spin up - even just one node - and perform this test without having to do this setup . A small addition might be needed so the cluster can run without consensus - I'll see whether I can have something quickly so you can use it here.

primary/src/grpc_server/configuration.rs
```diff
@@ -2,18 +2,26 @@
...
 use std::collections::BTreeMap;
 use tonic::{Request, Response, Status};
-use types::{Configuration, Empty, NewEpochRequest, NewNetworkInfoRequest, PublicKeyProto};
+use types::{
+    Configuration, Empty, GetPrimaryAddressResponse, MultiAddrProto, NewEpochRequest,
+    NewNetworkInfoRequest, PublicKeyProto,
+};
 
 pub struct NarwhalConfiguration<PublicKey: VerifyingKey> {
+    primary_addr: Multiaddr,
```
akichidis, line 14: Nit: I would just go with `primary_address` , we don't gain much by truncating

primary/tests/integration_tests_configuration_api.rs
```diff
@@ -181,6 +182,61 @@ async fn test_new_network_info() {
...
+
+    // Test gRPC server with client call
+    let mut client = connect_to_configuration_client(parameters.clone());
+
+    let request = tonic::Request::new(Empty {});
+
+    let response = client.get_primary_address(request).await.unwrap();
+    let actual_result = response.into_inner();
+
+    let re = Regex::new(r"ip4\x2f127\.0\.0\.1\x2ftcp\x2f[0-9]*\x2fhttp").unwrap();
```
akichidis, line 236: The need to have a regex seems to me that the exact address is not known due to some dynamic assignment, but can't we derive the exact address from the `committee` and compare against an exact match?

primary/src/grpc_server/configuration.rs
```diff
@@ -2,18 +2,26 @@
...
     /// The committee
     committee: SharedCommittee<PublicKey>,
 }
 
 impl<PublicKey: VerifyingKey> NarwhalConfiguration<PublicKey> {
-    pub fn new(committee: SharedCommittee<PublicKey>) -> Self {
-        Self { committee }
+    pub fn new(primary_addr: Multiaddr, committee: SharedCommittee<PublicKey>) -> Self {
```
akichidis, line 20: Since the `primary_addr` is derived from the `comittee` can't we just derive it from them when we initialise the struct, so we don't have the need to pass it separately?

akichidis (approved):
Looks good to me 👍, thanks @arun-koshy!  A few comments but approving to unblock

primary/tests/integration_tests_configuration_api.rs
akichidis: @arun-koshy you can find the changes on [this PR](https://github.com/MystenLabs/narwhal/pull/599) . An example of how you could use the `Cluster` in this test can be found [here](https://github.com/MystenLabs/narwhal/pull/599/files#diff-23e4050e4099fea2c66148e23cfb98113529fa70502a78764b98494536b4c19dR39)

primary/src/grpc_server/configuration.rs
arun-koshy: We would need the name (public key) of the primary to be able to derive it from the committee so I could either pass the name or the primary address directly in. Maybe the name is better if we ever want to derive other information from the committee?

primary/tests/integration_tests_configuration_api.rs
arun-koshy: This is awesome! Thanks Tasos 🤩

arun-koshy: Will wait for your PR to merge, to properly utilize your changes.

primary/src/grpc_server/configuration.rs
```diff
@@ -146,4 +154,15 @@ impl Configuration for NarwhalConfiguration {
 
         Ok(Response::new(Empty {}))
     }
+
+    async fn get_primary_address(
+        &self,
+        _request: Request<Empty>,
+    ) -> Result<Response<GetPrimaryAddressResponse>, Status> {
```
huitseeker, line 161: Since this always succeed, does it need to return a `Result`?

test_utils/src/cluster.rs
```diff
@@ -643,6 +643,24 @@ impl AuthorityDetails {
         ProposerClient::new(channel)
     }
 
+    /// Creates a new configuration client that connects to the corresponding client.
+    /// This should be available only if the internal consensus is disabled. If
+    /// the internal consensus is enabled then a panic will be thrown instead.
```
huitseeker, line 648: I don't think I see any risk in letting this be accessible with the internal consensus enabled. Do you?

huitseeker (commented):
This looks like it's getting in pretty good shape!

primary/src/grpc_server/configuration.rs
arun-koshy: tonic requires `Result` here

```
 note: expected fn pointer `fn(&'life0 NarwhalConfiguration, tonic::Request<_>) -> Pin<Box<(dyn futures::Future<Output = Result<tonic::Response<GetPrimaryAddressResponse>, Status>> + std::marker::Send + 'async_trait)>>`
 ```

test_utils/src/cluster.rs
arun-koshy: It won't be able to connect as the grpc server only starts when internal consensus is disabled [here](https://github.com/MystenLabs/narwhal/blob/main/primary/src/primary.rs#L331-L346).

primary/src/grpc_server/configuration.rs
arun-koshy: @akichidis I changed this to pass in the name instead of primary address, if you had something else in mind let me know and I will change it in a future PR.

PR #597, opened 2022-07-26
[testing] second node restart

akichidis:
This PR is introducing the following:

### A new test for the second node restart scenario. 

When we restart 2 nodes back to back we expect them to recover as expected. The scenario is the following:
1. **Spin up a cluster of 4 nodes** and let them advance a bit
2. **Restart first node** and let **it catch up**
3. **Restart second node** and let **it catch up**
4. **Now all nodes** should be in **same commit round** and none should be behind

### A few additions on the cluster struct
* some helper methods to restart a node and retrieve the workers.


huitseeker (approved):
This is great, and runs at far higher values for the `restart_delay` / `advance_delay`. Thanks so much @akichidis !

PR #598, opened 2022-07-26
Added a control system to limit the emission of headers under low volume

gdanezis:
This PR is a quick one as I am re-learning the N+T codebase. What is does:
- It records the number of batches (own + other) seen by the primary.
- It records the number of batches between the last commit (although not sure they are committed) and the current round (not sure they are included in the round). 
- It derives a multiplier applied to the inter header timer to try to ensure some minimum of batches are included per header.

Note that the estimation of progress is very rough, and the control logic is also very simple. However, in my tests this successfully slows rounds down when no traffic is sequenced, and picks up again when there is some volume. Slowing down emission when nearly no traffic exists is really the objective here.

### DIFF
--- a/primary/src/metrics.rs
+++ b/primary/src/metrics.rs
@@ -6,7 +6,13 @@ use prometheus::{
     default_registry, register_histogram_vec_with_registry, register_int_counter_vec_with_registry,
     register_int_gauge_vec_with_registry, HistogramVec, IntCounterVec, IntGaugeVec, Registry,
 };
-use std::time::Duration;
+use std::{
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
 use tonic::Code;
 
 #[derive(Clone)]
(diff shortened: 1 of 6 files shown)


gdanezis:
This is absolutely not a priority to review or land. Its just me getting familiar with the code and tests.

primary/src/metrics.rs
```diff
@@ -66,6 +108,9 @@ pub struct PrimaryMetrics {
     pub parent_requests_header_waiter: IntGaugeVec,
     /// Number of elements in pending list of certificate_waiter
     pub pending_elements_certificate_waiter: IntGaugeVec,
+
+    // Core metrics sneaking in
+    pub core_metrics: Arc<CoreCounters>,
```
akichidis, line 113: haha really sneaking in! Yeah I would suggest keeping those out of the `PrimaryMetrics` and generally outside of the concept of metrics that we use for monitoring purposes. I would suggest to initialise and inject them independently of the above struct.

primary/src/metrics.rs
```diff
@@ -36,6 +42,42 @@ pub(crate) fn initialise_metrics(metrics_registry: &Registry) -> Metrics {
...
+impl CoreCounters {
+    pub fn new() -> Arc<Self> {
+        Arc::new(Self::default())
+    }
+
+    pub fn set_commit_round(&self, round: u64) {
+        self.last_commit_round.store(round, Ordering::Relaxed);
+        let batches_now = self.current_batches.load(Ordering::Relaxed);
+        self.last_commit_batches
+            .store(batches_now, Ordering::Relaxed);
```
akichidis, line 61: > * It records the number of batches between the last commit (although not sure they are committed) and the current round (not sure they are included in the round).

ah I see now why. Agreed, I believe though we can come up with something more precise

primary/src/proposer.rs
```diff
@@ -227,6 +236,40 @@ impl<PublicKey: VerifyingKey> Proposer<PublicKey> {
...
+
+                // Implement a control system to ensure that some min amount of work
+                // happen in every round, and extend the rounds if it does not. Here we set a target
+                // number of batches committes per round `target_min_batch_per_round`. We then
+                // observe a weighted average of how many batches we commit per round and compute
+                // a multiplier of the max header delay to try to hit this number of rounds
+                // between 1 and `max_multiplier`.
+
+                // The number of rounds since we last committed.
+                let round_diff = self.round - self.metrics.core_metrics.last_commit_round();
```
akichidis, line 248: Given that we now support reconfiguration - which resets `round = 0` - let's take that into account , as in this case `self.round` can end up being lower than the `last_commit_round`.

primary/src/proposer.rs
```diff
@@ -227,6 +236,40 @@ impl<PublicKey: VerifyingKey> Proposer<PublicKey> {
...
+
+                // The control system sets log(multiplier) = log(multiplier) + log(target) - log(weighted_ave)
+                // with multiplier clipped between 1 and max_multiplier
+                multiplier = f64::max(
+                    1.0,
+                    f64::min(
+                        max_multiplier,
+                        multiplier * target_min_batch_per_round / (ave_batch_per_round + 0.1),
+                    ),
+                );
```
akichidis, line 270: Ah I would like to catch up with you offline @gdanezis about this and go through together. I would also like to better understand how this reacts with spikes in traffic.

akichidis (commented):
Just a first pass @gdanezis with a few comments, but I would be really interested to catch up offline about this :-)

asonnino:
See also https://github.com/MystenLabs/narwhal/pull/663, if we prefer this PR we may want to drop https://github.com/MystenLabs/narwhal/pull/663

akichidis:
Closing this as we have now moved Narwhal to https://github.com/MystenLabs/sui . @gdanezis please re-open it there if it's still relevant

PR #599, opened 2022-07-26
[testing] support external consensus for Cluster

akichidis:
This PR is enhancing the Cluster testing util to support running the cluster with internal consensus disabled - so we can manage the node via the gRPC interface that we have introduced. Some necessary changes have been made to ensure that JoinHandles are returned and allow us to shutdown the nodes properly. Also a helper method has been introduced in Cluster to allow us create a new proposer client seamlessly.


arun-koshy (approved):
Thanks Tasos!

PR #601, opened 2022-07-26
[logging] Tracing span verbosity reduction

huitseeker:
Some of our tracing spans print out lists of requests in-extenso, recording the `Debug` impl of a list of arbitrary length passed as a argument. This leads to 50KB messages.

This converts them to printing just the length of the list.

Also switches the verbosity of spans in consensus::dag to tracing.


mystenmark:
You beat me to it, thanks!

mystenmark (approved):

velvia:
Thanks @huitseeker .... anything which dumps out all the details of an object should probably be trace in most cases.

PR #602, opened 2022-07-26
fix: Correct the generics in ExecutionState

huitseeker:
This is a follow-up of #583. The `ExecutionState` trait incorrectly required one of its methods (`handle_consensus_transaction`) to be generic in a `PublicKey: VerifyingKey` parameter.
The Sui code base has no such genericity, making the implementation difficult. Moreover, the parameter is not needed in the first place (the `ExecutionState` implementations work fine with one single instantiated type for public keys).

### DIFF
--- a/executor/src/lib.rs
+++ b/executor/src/lib.rs
@@ -50,6 +50,8 @@ pub type SerializedTransactionDigest = u64;
 
 #[async_trait]
 pub trait ExecutionState {
+    type PubKey: VerifyingKey;
+
     /// The type of the transaction to process.
     type Transaction: DeserializeOwned + Send + Debug;
 
@@ -62,12 +64,12 @@ pub trait ExecutionState {
     /// Execute the transaction and atomically persist the consensus index. This function
     /// returns an execution outcome that will be output by the executor channel. It may
     /// also return a new committee to reconfigure the system.
-    async fn handle_consensus_transaction<PublicKey: VerifyingKey>(
+    async fn handle_consensus_transaction(
         &self,
-        consensus_output: &ConsensusOutput<PublicKey>,
+        consensus_output: &ConsensusOutput<Self::PubKey>,
         execution_indices: ExecutionIndices,
         transaction: Self::Transaction,
-    ) -> Result<(Self::Outcome, Option<Committee<PublicKey>>), Self::Error>;
+    ) -> Result<(Self::Outcome, Option<Committee<Self::PubKey>>), Self::Error>;
 
     /// Simple guardrail ensuring there is a single instance using the state
     /// to call `handle_consensus_transaction`. Many instances may read the state,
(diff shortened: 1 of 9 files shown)


node/tests/reconfigure.rs
```diff
@@ -48,16 +48,17 @@ impl SimpleExecutionState {
...
 
-    async fn handle_consensus_transaction<PublicKey: VerifyingKey>(
+    async fn handle_consensus_transaction(
         &self,
-        _consensus_output: &ConsensusOutput<PublicKey>,
+        _consensus_output: &ConsensusOutput<Ed25519PublicKey>,
         execution_indices: ExecutionIndices,
         transaction: Self::Transaction,
-    ) -> Result<(Self::Outcome, Option<Committee<PublicKey>>), Self::Error> {
+    ) -> Result<(Self::Outcome, Option<Committee<Ed25519PublicKey>>), Self::Error> {
```
huitseeker, line 61: This demo implementation demos exactly what the problem was.

executor/src/lib.rs
```diff
@@ -62,12 +64,12 @@ pub trait ExecutionState {
     /// Execute the transaction and atomically persist the consensus index. This function
     /// returns an execution outcome that will be output by the executor channel. It may
     /// also return a new committee to reconfigure the system.
-    async fn handle_consensus_transaction<PublicKey: VerifyingKey>(
+    async fn handle_consensus_transaction(
         &self,
-        consensus_output: &ConsensusOutput<PublicKey>,
+        consensus_output: &ConsensusOutput<Self::PubKey>,
```
joyqvq, line 69: this `ConsensusOutput` struct means something different in sui vs narwhal:

```
pub struct ConsensusOutput {
    #[serde(with = "serde_bytes")]
    pub message: Vec<u8>,
    pub sequence_number: SequenceNumber,
}
```

in narwhal
```
pub struct ConsensusOutput<PublicKey: VerifyingKey> {
    /// The sequenced certificate.
    pub certificate: Certificate<PublicKey>,
    /// The (global) index associated with this certificate.
    pub consensus_index: SequenceNumber,
}
```

huitseeker: We will indeed have no other choice, but that argument can simply be ignored until it is useful.

huitseeker:
@joyqvq I have posted a demo of the update this PR will allow me to make:
https://github.com/MystenLabs/sui/compare/main...huitseeker:update_narwhal_pointers_2?expand=1

gdanezis (approved):
Thanks this looks good -- I presume we can just `_` / ignore the generic type for the key on the Sui side for the moment right?

huitseeker:
@gdanezis After this PR, there is no generic type for the key required anywhere visible by Sui. See https://github.com/MystenLabs/sui/compare/main...huitseeker:update_narwhal_pointers_2?expand=1 for how this looks on the Sui side.

PR #603, opened 2022-07-27
Replace consensus number notification Atomic with async watch

gdanezis:
Refactor the cleanup code to use a watch that notifies other components of the consensus number change, instead of actively polling a shared Atomic u64 at every iteration of each select loop.


primary/src/header_waiter.rs
```diff
@@ -329,39 +326,42 @@ impl<PublicKey: VerifyingKey> HeaderWaiter<PublicKey> {
                         },
                         ReconfigureNotification::Shutdown => return
                     }
-                }
-            }
+                },
+
+                // Check for a new consensus round number
+                Ok(()) = self.rx_new_consensus_round.changed() => {
```
akichidis, line 332: I believe we might still need to have this outside of the `select` statement so GC doesn't run only when the round changes as requests as this will make GC run only once per round - while requests for earlier rounds might still come in.

That being said what we can probably do is keep the way you refactored it, and maybe immediately deliver any `header` either when parents or batches is requested in case the header's round is passed gc round logic so we don't necessarily need to wait until some GC run.

primary/src/primary.rs
```diff
@@ -189,7 +183,7 @@ impl Primary {
             certificate_store.clone(),
             synchronizer,
             signature_service.clone(),
-            consensus_round.clone(),
+            rx_new_consensus_round.clone(),
```
akichidis, line 186: you can also do `tx_newconsensus_round.subscribe()` to create new Receiver, in this case don't think there is much of a difference but just to raise awareness in case you haven't seen it.

primary/src/certificate_waiter.rs
```diff
@@ -44,8 +41,8 @@ pub struct CertificateWaiter<PublicKey: VerifyingKey> {
     committee: Committee<PublicKey>,
     /// The persistent storage.
     store: Store<CertificateDigest, Certificate<PublicKey>>,
-    /// The current consensus round (used for cleanup).
-    consensus_round: Arc<AtomicU64>,
+    /// Receiver for signal of round change
+    rx_new_consensus_round: watch::Receiver<u64>,
```
akichidis, line 45: Nit: to be honest I would probably stick to `rx_consensus_round` or `rx_current_consensus_round` . Whether the value communicated through is new or not is another matter and something that we detect via the `changed` method anyways. But again, a nit :)

akichidis (commented):
Thanks @gdanezis for this, I believe the watch is the better solution here! I've left a couple of comments, let me know your thoughts

asonnino:
I’ll be back fully back tomorrow and will be able to give a complete review. In the meantime, have you considered re-using the “reconfiguration” watch channel to propagate the consensus round? We may simply add another enum variant to this channel? The sender is held by the state handler and virtually every other task has a receiver.

primary/src/certificate_waiter.rs
huitseeker: I would go for `rx_consensus_round_updates`. A comment on this to indicate what the senders would look like would help.

huitseeker (approved):
This is great @gdanezis ! Thanks so much! There's a nasty `Arc<ArcSwap<>>` (the one on the `SharedCommittee`) that's going to need removal as well,  but you're right in general, the watch is the simpler approach.

primary/src/header_waiter.rs
gdanezis: Hey @akichidis -- I have moved it outside protected by a flag, but besides when the commit round advances, I am not sure under what other condition to trigger the GC. Any ideas?

primary/src/primary.rs
gdanezis: Yeah, that is nice, and used it in the main code (not tests).

akichidis (approved):
Great work @gdanezis ! Much better with the `watch` channel!

akichidis:
> I’ll be back fully back tomorrow and will be able to give a complete review. In the meantime, have you considered re-using the “reconfiguration” watch channel to propagate the consensus round? We may simply add another enum variant to this channel? The sender is held by the state handler and virtually every other task has a receiver.

@asonnino really good idea. The only thing that holds me a bit back are two things:
1. Giving too much responsibility to the reconfiguration channel and effectively couple our selfs - now too many messages will go via the same topic
2. The nature of the `watch` channel is to only keep [the last sent](https://docs.rs/tokio/0.2.22/tokio/sync/watch/index.html) value. That being said I am not sure of any edge cases we could observe here as all those are important messages.

But I agree it probably saves us monitoring another channel. Let's think that again separately?

PR #604, opened 2022-07-27
Make network interface info optional

### LINKED ISSUE #5309: [consensus API] Make the new network info optional

arun-koshy:
Fixes [Issue#539](https://github.com/MystenLabs/sui/issues/5309)

### DIFF
--- a/primary/src/block_synchronizer/mod.rs
+++ b/primary/src/block_synchronizer/mod.rs
@@ -634,10 +634,9 @@ impl BlockSynchronizer {
             .committee
             .others_primaries(&self.name)
             .into_iter()
-            .filter(|(_, address)| address.is_some())
-            .map(|(name, address)| (name, address.unwrap().primary_to_primary))
-            .filter(|(_, primary_to_primary)| primary_to_primary.is_some())
-            .map(|(name, primary_to_primary)| (name, primary_to_primary.unwrap()))
+            .filter_map(|(name, primary_address)| {
+                primary_address.map(|address| (name, address.primary_to_primary))
+            })
             .unzip();
 
         self.primary_network
--- a/primary/src/block_synchronizer/mod.rs
+++ b/primary/src/block_synchronizer/mod.rs
@@ -628,13 +628,16 @@ impl BlockSynchronizer {
     // It returns back the primary names to which we have sent the requests.
     #[instrument(level = "trace", skip_all)]
     async fn broadcast_batch_request(&mut self, message: PrimaryMessage) -> Vec<PublicKey> {
-        // Naively now just broadcast the request to all the primaries
-
+        // Naively now just broadcast the request to all the primaries we have
+        // addresses for.
         let (primaries_names, primaries_addresses) = self
             .committee
             .others_primaries(&self.name)
             .into_iter()
-            .map(|(name, address)| (name, address.primary_to_primary))
+            .filter(|(_, address)| address.is_some())
+            .map(|(name, address)| (name, address.unwrap().primary_to_primary))
+            .filter(|(_, primary_to_primary)| primary_to_primary.is_some())
+            .map(|(name, primary_to_primary)| (name, primary_to_primary.unwrap()))
             .unzip();
 
         self.primary_network
--- a/primary/src/core.rs
+++ b/primary/src/core.rs
@@ -157,12 +157,14 @@ impl Core {
         self.current_header = header.clone();
         self.votes_aggregator = VotesAggregator::new();
 
-        // Broadcast the new header in a reliable manner.
+        // Broadcast the new header in a reliable manner to primaries we have
+        // addresses for.
         let addresses = self
             .committee
             .others_primaries(&self.name)
             .into_iter()
-            .map(|(_, x)| x.primary_to_primary)
+            .filter(|(_, x)| x.is_some())
+            .filter_map(|(_, x)| x.unwrap().primary_to_primary)
             .collect();
 
         let message = PrimaryMessage::Header(header.clone());
@@ -289,7 +291,9 @@ impl Core {
                     .committee
                     .primary(&header.author)
                     .expect("Author of valid header is not in the committee")
-                    .primary_to_primary;
+                    .expect("Primary addresses of the author of valid header are not in the committee")
+                    .primary_to_primary
+                    .expect("Primary to primary address of the author of valid header is not in the committee");
                 let handler = self
                     .network
                     .send(address, &PrimaryMessage::Vote(vote))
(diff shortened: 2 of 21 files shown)


huitseeker (changes requested):
1. I would expect a change to the .proto, to make it explicit to callers they do not have to provide this information, is that possible?
2. how reasonable is it to push the option in the backend? At the moment, we panick if the address we're being passed does not parse. [Example](https://github.com/MystenLabs/narwhal/blob/7e49fa42e32437c4039f4ae40957e241dc0c2ec1/network/src/primary.rs#L60-L62) If instead of an empty address (or a parsing, but unreachable one), we get an explicit `Option`, we could instead hold off on trying to contact this validator.
Is there a reasonable cutoff point where we can push an `Option` futher into the backend (ideally until the `Committee`) ?

Requesting changes until we fix (2.) to the extend we don't run headlong into a panic.

arun-koshy:
> 1. I would expect a change to the .proto, to make it explicit to callers they do not have to provide this information, is that possible?

Proto3 fields are optional by default (i.e. generated code for existing [PrimaryAddresses](https://github.com/MystenLabs/narwhal/blob/main/types/src/generated/narwhal.rs#L37-L43) proto message). They recently added back the `optional` keyword but its used for [field presence](https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md) for basic types. I think convention is to not put the explicit `optional` keyword in the proto, but I don't mind adding it or a comment if thats best for our clients.


>    Is there a reasonable cutoff point where we can push an `Option` futher into the backend (ideally until the `Committee`) ?

Refactored the backend with optional primary addresses.

bmwill:
So I do think we want to make things optional, but I think we actually want to make the worker addresses optional over the primary's. We should always know concretely who the validators are and at least 1 address for them but we should be completely removing all the worker information out of the committee structure.

bmwill:
Ok I did read the linked issue and I see the intent....maybe we instead want to completely remove all network info from the committee structure?

primary/src/block_synchronizer/mod.rs
```diff
@@ -605,13 +605,16 @@ impl BlockSynchronizer {
...
+        // addresses for.
         let (primaries_names, primaries_addresses) = self
             .committee
             .others_primaries(&self.name)
             .into_iter()
-            .map(|(name, address)| (name, address.primary_to_primary))
+            .filter(|(_, address)| address.is_some())
+            .map(|(name, address)| (name, address.unwrap().primary_to_primary))
+            .filter(|(_, primary_to_primary)| primary_to_primary.is_some())
+            .map(|(name, primary_to_primary)| (name, primary_to_primary.unwrap()))
```
huitseeker, line 617: This is more concisely:
```suggestion
            .filter_map(|(name, address)| address.map(|addr| addr.primary_to_primary.map(|p2p| (name, p2p))))
```
It's fine to decompose this into more matching for clarity, rather than using combinators, but I think it's both possible and worthwhile to avoid unwraps.

primary/src/core.rs
```diff
@@ -158,12 +158,14 @@ impl Core {
...
-        // Broadcast the new header in a reliable manner.
+        // Broadcast the new header in a reliable manner to primaries we have
+        // addresses for.
         let addresses = self
             .committee
             .others_primaries(&self.name)
             .into_iter()
-            .map(|(_, x)| x.primary_to_primary)
+            .filter(|(_, x)| x.is_some())
+            .filter_map(|(_, x)| x.unwrap().primary_to_primary)
```
huitseeker, line 168:
```suggestion
            .filter_map(|(_, opt)| opt.map(|x| x.primary_to_primary))
```

primary/src/core.rs
```diff
@@ -316,12 +320,13 @@ impl Core {
...
 
-            // Broadcast the certificate.
+            // Broadcast the certificate to all primaries we have addresses for.
             let addresses = self
                 .committee
                 .others_primaries(&self.name)
                 .into_iter()
-                .map(|(_, x)| x.primary_to_primary)
+                .filter(|(_, x)| x.is_some())
+                .filter_map(|(_, x)| x.unwrap().primary_to_primary)
```
huitseeker, line 329:
```suggestion
                .filter_map(|(_, opt)| opt.map(|x| x.primary_to_primary))
```

primary/src/header_waiter.rs
```diff
@@ -312,7 +314,8 @@ impl HeaderWaiter {
                         let addresses = self.committee
                             .others_primaries(&self.name)
                             .into_iter()
-                            .map(|(_, x)| x.primary_to_primary)
+                            .filter(|(_, x)| x.is_some())
+                            .filter_map(|(_, x)| x.unwrap().primary_to_primary)
```
huitseeker, line 318:
```suggestion
                            .filter_map(|(_, opt)| opt.map(|x| x.primary_to_primary))
```

huitseeker (changes requested):
So, w.r.t @bmwill 's comment, removing all network information from the committee seems excessive, but having an option for each of the interfaces of a primary / worker seems excessive as well: the networking information for an endpoint should be entirely provisioned or none at all.

But the major issue I have here is one of protocol: In several places, notably in all the places where we use a `(unreliable_)?broadcast` function, we iterate over the addresses in the committee, and send a message to all of them. From a protocol standpoint, at least for the reliable variants of those, we cannot make progress unless those messages are sent to all their destinaries. The current approach, which amounts to sending to those primaries we have addresses for, will lead to a loss of liveness, I think.

How do we fix this? The simplest way I can think of is to modify the `ReliableNetwork` trait (at least), and modify the `broadcast` function there so that we wait, before the actual sending, until we have at least (2f+1) addresses in the `Committee`.

We should also implement this wait for all the functions in this same trait.

arun-koshy:
> How do we fix this? The simplest way I can think of is to modify the ReliableNetwork trait (at least), and modify the broadcast function there so that we wait, before the actual sending, until we have at least (2f+1) addresses in the Committee.

@huitseeker this is what I plan on doing. First thing I thought about doing was passing the `Committee` into the `ReliableNetwork` broadcast method instead of a list of addresses so that we could calculate if the appropriate total stake has been reached but realized that this would have to be special cased for both the `WorkerNetwork` and the `PrimaryNetwork`. So I think it makes sense to remove the default trait impl for reliable broadcast and then implement the trait in both the worker & primary case separately but with a shared trait interface that looks like this

```
async fn broadcast(
        &mut self,
        committee: Committee,
        worker_cache: SharedWorkerCache,
        message: &Self::Message,
    ) -> Vec<CancelOnDropHandler<MessageResult>>
```

The difference would be the worker impl would retrieve the worker addresses and the primary impl would retrieve the primary addresses but they would both retry the same way until we have atleast (2f+1) addresses. 

The worker broadcast seems like it will work fine however the  issue I am seeing for the primary broadcast is that without a `SharedCommittee` I am not sure how we can properly retry until we have (2f+1) addresses as the committee that is passed in will never change. Do we want to switch to using a SharedCommittee or do we have to wait for a [NewEpoch](https://github.com/MystenLabs/narwhal/blob/6b92f8eb38f906e8602dc1f2cd69d108371c1575/types/src/primary.rs#L554) or a [UpdateCommittee](https://github.com/MystenLabs/narwhal/blob/6b92f8eb38f906e8602dc1f2cd69d108371c1575/types/src/primary.rs#L556) signal?

akichidis:
Let's re-open this in https://github.com/MystenLabs/sui if it's still relevant

PR #605, opened 2022-07-27
[fix] linter issue

akichidis:
Linter complains in main (for good reason!) as a parameter is missing from the `Cluster` in the `test_second_node_restart` test


huitseeker (approved):
Another instance where we need Bors.

PR #608, opened 2022-07-27
crypto: expose recover pubkey feature for Secp256k1Signature

joyqvq:


crypto/src/tests/secp256k1_tests.rs
```diff
@@ -67,6 +67,17 @@ fn test_public_key_bytes_conversion() {
     assert_eq!(kp.public().as_bytes(), rebuilded_pk.as_bytes());
 }
 
+#[test]
+fn test_public_key_recovery() {
```
akichidis, line 71: Could you please test the error scenario as well?

crypto/src/secp256k1.rs
```diff
@@ -393,3 +393,14 @@ impl From<Secp256k1PrivateKey> for Secp256k1KeyPair {
         Secp256k1KeyPair { name, secret }
     }
 }
+
+impl Secp256k1Signature {
+    /// Recover public key from signature
+    pub fn recover(&self, hashed_msg: &[u8]) -> Result<Secp256k1PublicKey, signature::Error> {
+        let hashed_msg = rust_secp256k1::Message::from_slice(hashed_msg).unwrap();
```
akichidis, line 400: By doing directly the `unwrap` a panic will be raised if the message is invalid - which can happen. I would suggest to refactor this and return an/the error instead when the  result is not an Ok.

akichidis (commented):
Thanks for this @joyqvq ! A couple of comments from my side

kchalkias (approved):

PR #612, opened 2022-07-27
[network] refactor: Network renamings

huitseeker:
These renamings will help clarify and regularize the semantics of our network-related functions.

`CancelHandler` -> `CancelOnDropHandler`
unreliable `broadcast` -> `unreliable_broadcast`
Extracted from a larger, still-upcoming PR.

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -23,17 +23,18 @@ pub use crate::{
 // the result of our network messages
 pub type MessageResult = Result<tonic::Response<types::Empty>, anyhow::Error>;
 
+/// This adapter will make a [`tokio::task::JoinHandle`] abort its handled task when the handle is dropped.
 #[derive(Debug)]
 #[must_use]
-pub struct CancelHandler<T>(pub tokio::task::JoinHandle<T>);
+pub struct CancelOnDropHandler<T>(tokio::task::JoinHandle<T>);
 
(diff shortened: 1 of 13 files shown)


network/src/lib.rs
```diff
@@ -25,15 +25,15 @@ pub type MessageResult = Result<tonic::Response<types::Empty>, anyhow::Error>;
 
 #[derive(Debug)]
 #[must_use]
-pub struct CancelHandler<T>(pub tokio::task::JoinHandle<T>);
+pub struct CancelOnDropHandler<T>(tokio::task::JoinHandle<T>);
```
akichidis, line 28: ❤️

akichidis: Since you are touching this, could you also please add a quick comment about the behaviour as well?

network/src/primary.rs
```diff
@@ -245,7 +247,7 @@ impl PrimaryToWorkerNetwork {
             .await
     }
 
-    pub async fn broadcast<T: VerifyingKey>(
+    pub async fn unreliable_broadcast<T: VerifyingKey>(
```
akichidis, line 250: Similar, since you are doing the renaming , a short 1-2 lines of what I should expect from this method would be great!

akichidis (approved):
Thanks @huitseeker  for this! A couple of optional comments.

PR #614, opened 2022-07-28
[refactor] The Great Genericity Simplification

huitseeker:
## Context 
MystenLabs/narwhal#35 introduced generic traits for public keys (`VerifyingKey`), private keys (`SigningKey`), signatures (`Authenticator`) and key pairs (`KeyPair`). We then added support for more schemes implementing this, notably BLS12-381 (#37), BLS12-377 (#36) and secp256k1 (#484). 

## The issue

However, the code base was left in a state where most structures are defined generically, which is unwieldy (see e.g. the definitions of `ExecutionState` and `ConsensusProtocol` in resp. MystenLabs/narwhal#489 and MystenLabs/narwhal#385) and limits us in using genericity elsewhere (e.g. MystenLabs/narwhal#171 could have conveyed the same functionality in ~ 1/2 the SLoC). It also led to some repair PRs (#583 MystenLabs/narwhal#602).

## The solution

We notice, following the discipline in Sui (https://github.com/MystenLabs/sui/pull/2994) and Diem before it, that we do not have to keep the code base generic since we expect at most one signing scheme to be used for authority signing at any given time. We expect to always be able to pick that single scheme at compile time (this is only used for validator signing, not user signing). We therefore confine the genericity to the crypto crate, and use aliases to pick one specific scheme at the top of that crate (`crypto::{PublicKey, Signature, PrivateKey, KeyPair}`). Those aliases are then what is used in the rest of the code base. 

Provided we do not add nasty introspection in the specifics of one scheme between now and then, we will only have to switch the definition of the aliases in order to switch the signing scheme we use throughout the code base (including tests).

This should pave the way for further refactorings, among which MystenLabs/sui#5292.

### DIFF
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -54,8 +54,6 @@ pub mod secp256k1;
 
 pub mod traits;
 
-pub type CryptoError = ed25519_dalek::ed25519::Error;
-
 pub const DIGEST_LEN: usize = 32;
 
 pub fn blake2b_256<F: Fn(&mut blake2::VarBlake2b)>(closure: F) -> [u8; DIGEST_LEN] {
--- a/crypto/src/lib.rs
+++ b/crypto/src/lib.rs
@@ -14,13 +14,12 @@ use blake2::{digest::VariableOutput, VarBlake2b};
 use rand::{rngs::OsRng, CryptoRng, RngCore};
 use serde::{Deserialize, Serialize};
 
-pub use signature::{Signature, Verifier};
+pub use signature::{Signature as _, Verifier};
 use std::fmt;
 use tokio::sync::{
     mpsc::{channel, Sender},
     oneshot,
 };
-use traits::{Authenticator, KeyPair};
 
 #[cfg(test)]
 #[path = "tests/pubkey_bytes_tests.rs"]
@@ -44,15 +43,26 @@ pub mod bls12377;
 #[path = "tests/bls12381_tests.rs"]
 pub mod bls12381_tests;
 
+// Signing traits
+pub mod traits;
+// Key scheme implementations
 pub mod bls12381;
 pub mod ed25519;
+pub mod secp256k1;
+// Other tooling
 pub mod hkdf;
 pub mod pubkey_bytes;
 pub mod serde_helpers;
 
-pub mod secp256k1;
-
-pub mod traits;
+// Here we select the types that are used by default in the code base.
+// The whole code base should only:
+// - refer to those aliases and not use the individual scheme implementations
+// - not use the schemes in a way that break genericity (e.g. using their Struct impl functions)
+// - swap one of those aliases to point to another type if necessary
+pub type PublicKey = ed25519::Ed25519PublicKey;
+pub type Signature = ed25519::Ed25519Signature;
+pub type PrivateKey = ed25519::Ed25519PrivateKey;
+pub type KeyPair = ed25519::Ed25519KeyPair;
 
 pub const DIGEST_LEN: usize = 32;
 
@@ -110,11 +120,11 @@ pub trait Hash {
 // Generic Keypair
 ////////////////////////////////////////////////////////////////
 
-pub fn generate_production_keypair<K: KeyPair>() -> K {
+pub fn generate_production_keypair<K: traits::KeyPair>() -> K {
     generate_keypair::<K, _>(&mut OsRng)
 }
 
-pub fn generate_keypair<K: KeyPair, R>(csprng: &mut R) -> K
+pub fn generate_keypair<K: traits::KeyPair, R>(csprng: &mut R) -> K
 where
     R: CryptoRng + RngCore,
 {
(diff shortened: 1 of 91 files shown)


crypto/src/lib.rs
```diff
@@ -44,17 +43,26 @@ pub mod bls12377;
...
-pub mod secp256k1;
-
-pub mod traits;
-
-pub type CryptoError = ed25519_dalek::ed25519::Error;
+// Here we select the types that are used by default in the code base.
+// The whole code base should only:
+// - refer to those aliases and not use the individual scheme implementations
+// - not use the schemes in a way that break genericity (e.g. using their Struct impl functions)
+// - swap one of those aliases to point to another type if necessary
```
kchalkias, line 61: I would add an example here how someone would update all of them for BLS for instance (we either update all four types here or not when we switch schemes).

crypto/src/lib.rs
```diff
@@ -44,17 +43,26 @@ pub mod bls12377;
...
+pub type PublicKey = ed25519::Ed25519PublicKey;
```
kchalkias, line 62: Does this mean that projects which use a different scheme will have to fork our crate and update these types with their preferred algorithm?

huitseeker: That depends. Projects that want to use our crypto library won't have to (in fact it should soon live under mysten-infra). Projects that want to use Narwhal with a different scheme will have to, yes.
We can play a conditional compilation game here (reminiscent of Dalek backend selection) but it has severe limitations (as you know features are additive).

types/src/primary.rs
```diff
@@ -560,23 +558,23 @@ pub enum PrimaryMessage<PublicKey: VerifyingKey> {
 /// Message to reconfigure worker tasks.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(bound(deserialize = "PublicKey: VerifyingKey"))]
```
akichidis, line 560: `#[serde(bound(deserialize = "PublicKey: VerifyingKey"))]` this is not needed any more right?

akichidis: Same comment applies for all the other similar instances

akichidis (approved):
This makes total sense, what an awesome simplification! 💯 Thanks a lot @huitseeker . Left a couple of comments.

One more thing from my side, is there a test we can introduce which checks the expected chosen types (ex `ed25519::Ed25519PublicKey`) for the codebase? I mostly would like to make sure that we have something that breaks when someone tries to make a significant change like this. Not a blocker for the PR and only if you think as well that gives value.

kchalkias (commented):
Awesome work, two minor comments:

- I think we should point to a clear test that verifies if the 4 types are aligned (they refer to the same scheme).
- due to your recent BCS change in the yaml file, I would add a comment somewhere that explains exactly what someone should do to replace our ed with a new scheme, ie BLS. It's not only the types we change, it's the yaml file as well, which might be missed by devs in the future.

huitseeker:
@kchalkias I would invite you to download this PR, edit the crypto file to change one of the type aliases (but not the others), and witness the dozens of compilation errors. We do not compromise one bit on type safety in making this change: due to the way each trait implementation refers to its other associated (pun intended) types, it's impossible to change one type without the others, and expect *any* of the codebase to function. I'll add a comment, but the probability someone makes a blunder on this is zero.

@akichidis regarding making sure that Ed25519PublicKey is indeed the type used as the concrete instantiation of the alias, we already have a few tests that cover it:
- the BCS manifest already records our concrete type (by construction, the serde-reflection machinery dereferences type aliases)
- the snapshot tests for our configuration files contain explicit ser/de of actual keys, which will not deserialize under any other scheme.

huitseeker:
@kchalkias 
> It's not only the types we change, it's the yaml file as well, which might be missed by devs in the future.

The BCS file already has its own test protecting it, so that a change in type aliases will break that test. The test already has documentation on how to address the failure.

kchalkias:
@huitseeker IMO we shouldn't always expect from Devs to run tests for a change like this it would be nice to have a well written doc. It's gonna be part of our docathon anyway, so it's fine as it is now.

akichidis:
> @akichidis regarding making sure that Ed25519PublicKey is indeed the type used as the concrete instantiation of the alias, we already have a few tests that cover it:
> 
> the BCS manifest already records our concrete type (by construction, the serde-reflection machinery dereferences type aliases)
> the snapshot tests for our configuration files contain explicit ser/de of actual keys, which will not deserialize under any other scheme.

Ah awesome, thanks @huitseeker !

huitseeker:
> IMO we shouldn't always expect from Devs to run tests for a change like this, it would be nice to have a well written doc on what's required to change the scheme.

That is ok, we are not expecting them to run tests here. It's not the tests that break, it's the type checker. In concrete terms (and you can try it yourself), they will try to syntactically replace the type aliases partially, and their entire IDE will light up with dozens of problems. A partial replace does not type check.

You don't need to run any tests to see this.

PR #616, opened 2022-07-28
fix: make apt update quiet

allan-bailey:
Adding `-q` to the apt commands to make them quiet.  No need to see that output.


huitseeker (approved):
💯

akichidis (approved):

PR #617, opened 2022-07-28
fix: redirect apt output to /dev/null

allan-bailey:
So the `-qq` option doesn't work, so adding redirect to /dev/null

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -21,7 +21,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get -qq update \
     libssl-dev \
     pkg-config \
     clang \
-    cmake  2> /dev/null > /dev/null \
+    cmake > /dev/null 2>&1 \
     && rm -rf /var/lib/apt/lists/*
 # Install the fmt
 RUN rustup component add rustfmt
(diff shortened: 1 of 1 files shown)


Docker/Dockerfile
```diff
@@ -11,7 +11,7 @@ WORKDIR "$WORKDIR/narwhal"
 ARG BUILD_MODE
 
 # Install basic dependencies
-RUN apt-get -qq update \
+RUN DEBIAN_FRONTEND=noninteractive apt-get -qq update \
```
akichidis, line 14: If the `-qq` doesn't work do we still want to leave it there?

allan-bailey: For `apt update` it does work, for `install` it doesn't.

Docker/Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN apt-get -qq update \
     libssl-dev \
     pkg-config \
     clang \
-    cmake \
+    cmake  2> /dev/null > /dev/null \
```
huitseeker, line 24: Would `> /dev/null 2>&1` not work in a more readable fashion?

huitseeker (approved):
Left a comment for an example.

PR #618, opened 2022-07-28
crypto: use result instead of unwrap when parsing

joyqvq:


kchalkias (approved):
Approving, we might spend some time in the future on returning meaningful errors per what failed. Not in this PR but during the clean up and docathon week

crypto/src/secp256k1.rs
```diff
@@ -231,12 +231,20 @@ impl<'de> Deserialize<'de> for Secp256k1Signature {
...
+            Ok(rec_id) => match rust_secp256k1::ecdsa::RecoverableSignature::from_compact(
+                &bytes[..64],
+                rec_id,
+            ) {
+                Ok(sig) => Ok(Secp256k1Signature {
+                    sig,
+                    bytes: OnceCell::new(),
+                }),
+                Err(_) => Err(signature::Error::new()),
+            },
```
huitseeker, line 247:
```suggestion
RecoveryId::from_i32(bytes[64] as i32)
            .and_then(|rec_id| {
                RecoverableSignature::from_compact(&bytes[..64], rec_id).map(|sig| {
                    Secp256k1Signature {
                        sig,
                        bytes: OnceCell::new(),
                    }
                })
            })
            .map_err(signature::Error::from_source)
```

huitseeker (commented):
1. I think we can use imports rather than fully qualified names,
2. I suggested a reformulation,
3. `signature::Error` has `from_source` for cases when it's not safety-critical to leak the error. I suggest it here, but it may or may not be appropriate.

crypto/src/secp256k1.rs
joyqvq: (thanks for bearing with my rusty rust, looks so much better 💯 )

```
**NOTE:** The "source" should **NOT** be used to propagate cryptographic errors e.g. signature parsing or verification errors. 
```
as we are parsing signature here, this may not be suitable

PR #556, opened 2022-07-21
crypto: use secp256k1 instead of k256

joyqvq:

PR #619, opened 2022-07-28
crypto: use ecrecover to verify signature to ensure non-malleability

joyqvq:


crypto/src/secp256k1.rs
```diff
@@ -85,9 +85,18 @@ impl Verifier<Secp256k1Signature> for Secp256k1PublicKey {
         #[cfg(not(test))]
         let message = Message::from_hashed_data::<rust_secp256k1::hashes::sha256::Hash>(msg);
 
-        let vrfy = Secp256k1::verification_only();
-        vrfy.verify_ecdsa(&message, &signature.sig.to_standard(), &self.pubkey)
```
joyqvq, line 89: this library is not safe to use since it will verify valid for both signature and its flipped version. 

i was able to reproduce against the newly added unit tests. 

we need to ensure the recovered pubkey to match the original one to be sure it is not malleated. the flipped signature would fail here.

joyqvq:
good news is it makes verify even faster:

```
Secp256k1 signature verification
                        time:   [34.152 µs 34.189 µs 34.231 µs]
                        change: [-52.080% -51.965% -51.846%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) low mild
  3 (3.00%) high mild
  3 (3.00%) high severe
```
compared to https://github.com/MystenLabs/narwhal/pull/556#issuecomment-1191969201

crypto/src/secp256k1.rs
huitseeker: > i was able to reproduce against the newly added prop tests.

If you did get a proptest failure, the error message allows you to record the seed of the rng, so that this specific generated case gets ran through on further test runs.

The seed is printed with the test failure, as are instruction on which file to record it in.

crypto/src/secp256k1.rs
```diff
@@ -85,9 +85,18 @@ impl Verifier<Secp256k1Signature> for Secp256k1PublicKey {
...
+        match signature.sig.recover(&message) {
+            Ok(recovered_key) => {
+                if self.as_bytes() == recovered_key.serialize().as_slice() {
+                    Ok(())
+                } else {
+                    Err(signature::Error::new())
+                }
+            }
+            Err(_) => Err(signature::Error::new()),
+        }
```
huitseeker, line 99:
```suggestion
        match signature.sig.recover(&message) {
            Ok(recovered_key) if self.as_bytes() == recovered_key.serialize().as_slice() => Ok((),
            _ => Err(signature::Error::new()),
        }
```

huitseeker (approved):
Just a couple of comments, thanks!

PR #622, opened 2022-07-29
refactor: make functions returning join handles `#[must_use]`

huitseeker:
This should make the scoping and termination of tasks (which will be key to MystenLabs/sui#5313) clearer to analyze.

(Along with MystenLabs/narwhal#427 and MystenLabs/narwhal#428, ) Closes MystenLabs/narwhal#77.

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -210,6 +210,7 @@ impl<Protocol> Consensus<Protocol>
 where
     Protocol: ConsensusProtocol + Send + 'static,
 {
+    #[must_use]
     pub fn spawn(
         committee: Committee,
         store: Arc<ConsensusStore>,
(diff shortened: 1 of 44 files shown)


akichidis (approved):
Thanks so much for this @huitseeker , it will keep us honest!

Feel free to completely ignore this proposal, but since this appears to be quite important to us (using the returned JoinHandles) I would propose to be a bit more verbose and add the reason for why we want this to be a `must_use` so we avoid someone accidentally removing those in the future.

Ex `#[must_use = "should use to avoid creating detached tasks losing control over them"]`

again it might be a bit far fetched...

huitseeker:
@akichidis good suggestion, but I have another, even more structured refactoring on the way, which should supersede this comment.

Issue #544, opened 2022-07-20
Use input Rng across all KeyPair impl


joyqvq:
1. clean this hack up once we upgrade our eddsa implementation,
2. NOT use this in tests: we have several fixtures that generate / copy the same keys in tests repeatedly by passing in carefully-seeded RNGs (see `keys()` in `test_utils`).

_Originally posted by @huitseeker in https://github.com/MystenLabs/narwhal/pull/541#discussion_r925925119_

Issue #535, opened 2022-07-20
Replace ed25519-dalek with ed25519-consensus


joyqvq:
Less likely to be a task in narwhal, once we are done extracting the cryptos, replace ed25519-dalek so that we can bump the version for `rand` and `signature` as well.

PR #624, opened 2022-07-29
crypto: replace ed25519-dalek with ed25519-consensus

joyqvq:
https://github.com/MystenLabs/narwhal/issues/535


crypto/src/ed25519.rs
```diff
@@ -40,12 +46,17 @@ pub struct Ed25519KeyPair {
 
 #[serde_as]
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct Ed25519Signature(#[serde_as(as = "Ed25519Sig")] pub ed25519_dalek::Signature);
+pub struct Ed25519Signature {
+    #[serde_as(as = "Ed25519Sig")]
+    pub sig: ed25519_consensus::Signature,
+    #[serde(skip)]
+    pub bytes: OnceCell<[u8; ED25519_SIGNATURE_LENGTH]>,
+}
```
joyqvq, line 54: unfortunately the consensus library does not support as_ref out of the box and had to use this workaround. 

@huitseeker @punwai do you know of any trick to use serder_as to flatten a field to not include the tag? so the serialization would be 
"\"39+HHD6kbDZoFVkDiob1cfHIoq3FE8tMGckMIs58za/lys/omQtvxouGLvJ1SPiLBf7Kge2a4L7uYY1L4iAwCA==\""
instead of
"{\"sig\":\"39+HHD6kbDZoFVkDiob1cfHIoq3FE8tMGckMIs58za/lys/omQtvxouGLvJ1SPiLBf7Kge2a4L7uYY1L4iAwCA==\"}"

otherwise, this would have downstream impact on config files again

crypto/src/tests/ed25519_tests.rs
```diff
@@ -66,8 +66,9 @@ fn test_serde_signatures_human_readable() {
     println!("{:?}", serialized);
     assert_eq!(
         format!(
-            "\"{}\"",
-            base64ct::Base64::encode_string(&signature.0.to_bytes())
+            "{{\"sig\":\"{}\"}}",
+            // "\"{}\"",
+            base64ct::Base64::encode_string(&signature.sig.to_bytes())
```
joyqvq, line 71: where it changes

crypto/src/ed25519.rs
```diff
@@ -354,11 +372,11 @@ impl KeyPair for Ed25519KeyPair {
         }
     }
 
-    fn generate<R: rand::CryptoRng + rand::RngCore>(rng: &mut R) -> Self {
-        let kp = ed25519_dalek::Keypair::generate(rng);
+    fn generate<R: rand::CryptoRng + rand::RngCore>(_rng: &mut R) -> Self {
+        let kp = ed25519_consensus::SigningKey::new(OsRng);
```
joyqvq, line 376: i plan to upgrade rand and replace all OsRng in one PR with https://github.com/MystenLabs/narwhal/issues/544 to keep this one focused

crypto/src/ed25519.rs
```diff
@@ -18,17 +19,22 @@ use crate::{
...
+pub const ED25519_PUBLIC_KEY_LENGTH: usize = 32;
+pub const ED25519_SIGNATURE_LENGTH: usize = 64;
+
 ///
 /// Define Structs
 ///
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Ed25519PublicKey(pub ed25519_dalek::PublicKey);
+pub struct Ed25519PublicKey(pub ed25519_consensus::VerificationKey);
```
punwai, line 32: You can use serde_with here.

crypto/src/ed25519.rs
punwai: Yeah, I think you should be using serde_with for this.

punwai: Not that I've seen - I've just been writing it without serde_as in situations like these.

config/tests/snapshots/config_tests__committee.snap
```diff
@@ -4,7 +4,7 @@ expression: committee
 ---
 {
   "authorities": {
-    "IP26ybELdYe7p7W8FjvOaeeW1x5O1EwQ/LRIhon3oUQ=": {
+    "ildi4hrBzbOHBELHe0w69Yx87bh3nQJw5tTx4vc2fXQ=": {
```
joyqvq, line 7: this is expected as rand 0.8.5 uses ChaChaRng for 20 rounds whereas rand 0.7.3 uses 12 rounds with reference to https://github.com/rust-random/rand/issues/932

from 0.7.3: 
```
/// The current algorithm used is the ChaCha block cipher with 20 rounds.
/// This may change as new evidence of cipher security and performance
/// becomes available.
```

from 0.8.5:
```
/// The current algorithm used is the ChaCha block cipher with 12 rounds. Please
/// see this relevant [rand issue] for the discussion. This may change as new 
/// evidence of cipher security and performance becomes available.

/// [rand issue]: https://github.com/rust-random/rand/issues/932
```

crypto/src/ed25519.rs
```diff
@@ -223,8 +242,36 @@ impl Display for Ed25519Signature {
 
 impl Default for Ed25519Signature {
     fn default() -> Self {
-        let sig = <ed25519_dalek::Signature as Signature>::from_bytes(&[0u8; 64]).unwrap();
-        Ed25519Signature(sig)
+        <Ed25519Signature as Signature>::from_bytes(&[1u8; ED25519_SIGNATURE_LENGTH]).unwrap()
+    }
+}
+
+impl Serialize for Ed25519Signature {
```
joyqvq, line 249: as discussed serde_as would produce `{"sig": base64_encoded_string}` instead of a flattened `base64_encoded_string`, removed related functions in serder_helper

crypto/src/ed25519.rs
```diff
@@ -287,25 +338,15 @@ impl AggregateAuthenticator for Ed25519AggregateSignature {
...
             return Err(signature::Error::new());
         }
-        let mut inner_messages: Vec<&[u8]> = Vec::new();
-        for i in 0..messages.len() {
-            for _ in 0..pks[i].len() {
-                inner_messages.push(messages[i]);
+        let mut batch = batch::Verifier::new();
+
+        for i in 0..pks.len() {
+            for j in 0..pks[i].len() {
```
joyqvq, line 344: rewrote this a bit to adapt to the library, since the length is the same, a double for loop should work? cc @punwai

crypto/src/tests/ed25519_tests.rs
```diff
@@ -41,7 +41,7 @@ fn serialize_deserialize() {
 
     let private_key = kpref.private();
     let bytes = bincode::serialize(&private_key).unwrap();
-    let privkey = bincode::deserialize::<Ed25519PublicKey>(&bytes).unwrap();
+    let privkey = bincode::deserialize::<Ed25519PrivateKey>(&bytes).unwrap();
```
joyqvq, line 44: i dont know why this test passed on main branch.

node/tests/staged/narwhal.yaml
```diff
@@ -43,8 +43,6 @@ Committee:
     - epoch: U64
 Ed25519PublicKey:
   NEWTYPESTRUCT: STR
-Ed25519Signature:
-  NEWTYPESTRUCT: BYTES
```
joyqvq, line 47: is there any downstream impact on this file? it seems like just a documentation

crypto/src/ed25519.rs
punwai: lgtm

huitseeker (commented):
Thanks @joyqvq ! This looks good, I just have 3 questions:
1. Is this change sufficient to fix #544 as well?
2. Can we write one negative test for invalid EdDSA serialization? The full issue is #500, no need to solve it, I just want to see a non-regression on the unit test you fix: you fix the code, I would like to see the non-fixed code fail.
3. We should open an issue to inspect our now pervasive use of `OnceCell` in key/signature material and make sure it complies with serde best practices.

config/tests/snapshots/config_tests__committee.snap
joyqvq: yeah, only for the generate function in KeyPair trait that takes in the Rng

joyqvq:
> Thanks @joyqvq ! This looks good, I just have 3 questions:
> 
> 1. Is this change sufficient to fix #544 as well?
> 2. Can we write one negative test for invalid EdDSA serialization? The full issue is #500, no need to solve it, I just want to see a non-regression on the unit test you fix: you fix the code, I would like to see the non-fixed code fail.
> 3. We should open an issue to inspect our now pervasive use of `OnceCell` in key/signature material and make sure it complies with serde best practices.

1. yes - although i only touched ed25519 files to keep the scope for review, will follow up a diff to upgrade rand across the board and eliminate the workaround
2. 👍 good point, test added, will pick up full issue next
3. opened

huitseeker (approved):
Tip top! thanks a bunch, @joyqvq !

PR #625, opened 2022-07-29
Switch to use `Multiaddr` to configure prometheus server for consistency

### LINKED ISSUE #575: Prometheus metrics configuration should take a MultiAddr
The Prometheus metrics are exposed on an endpoint that, unlike the rest of the infrastructure, is not specified with a `Multiaddr` but with a `Socket`, parsed from an IPV4 in its default instance:
https://github.com/MystenLabs/narwhal/blob/28b2c89ded51f79386eed2e23c2a3325cfe3e8d6/config/src/lib.rs#L137-L150

It should be converted to use a `Multiaddr` like the rest of our config.

mwtian:
For consistency (in json / yaml config files), we would like to use the text format `Multiaddr` to specify the address and port for all services, including prometheus.

Closes #575

### DIFF
--- a/Docker/gen.validators.sh
+++ b/Docker/gen.validators.sh
@@ -120,7 +120,7 @@ cat > ${target}/parameters.json <<EOF
     "sync_retry_delay": "10_000ms",
     "sync_retry_nodes": 3,
     "prometheus_metrics": {
-        "socket_addr": "0.0.0.0:8010"
+        "socket_addr": "/ip4/0.0.0.0/tcp/8010"
     }
 }
 EOF
--- a/Docker/gen.validators.sh
+++ b/Docker/gen.validators.sh
@@ -120,7 +120,7 @@ cat > ${target}/parameters.json <<EOF
     "sync_retry_delay": "10_000ms",
     "sync_retry_nodes": 3,
     "prometheus_metrics": {
-        "socket_addr": "/ip4/0.0.0.0/tcp/8010"
+        "socket_addr": "/ip4/0.0.0.0/tcp/8010/http"
     }
 }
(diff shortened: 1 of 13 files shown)


Docker/gen.validators.sh
```diff
@@ -120,7 +120,7 @@ cat > ${target}/parameters.json <<EOF
     "sync_retry_delay": "10_000ms",
     "sync_retry_nodes": 3,
     "prometheus_metrics": {
-        "socket_addr": "0.0.0.0:8010"
```
huitseeker, line 123: 👍 Nice catch!

huitseeker (approved):
Super! This perhaps hints at mentioning the following three places on top of the `parameter_XX_matches` tests as places to update at when updating a snapshot:
- the Docker file,
- the Fab files for the benchmark,
- the Sui configurations & snapshot tests (which as you probably expect, will break as well when we import this Narwhal commit to Sui)

Docker/gen.validators.sh
mwtian: Thanks. In fact this json looks the same as `Docker/validators/parameters.json` so I changed to copy the file instead.

mwtian:
> Super! This perhaps hints at mentioning the following three places on top of the `parameter_XX_matches` tests as places to update at when updating a snapshot:
> 
> * the Docker file,
> * the Fab files for the benchmark,
> * the Sui configurations & snapshot tests (which as you probably expect, will break as well when we import this Narwhal commit to Sui)

Good point. I hope we can reduce the list a bit but `parameters.json` for benchmarking is indeed different. Added a comment.

PR #627, opened 2022-07-30
[testing-framework] Zero crypto

punwai:
#### Summary
This file contains an implementation of a negligible-cost (apart from some ocassional pk copying) trivial cryptographic scheme. The purpose of this library is to allow seamless benchmarking of systems without taking into account the cost of cryptographic primitives - and hence providing a theoretical maximal throughput that a system could achieve if the cost of crypto is optimized away.

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -37,6 +37,7 @@ readonly = "0.2.1"
 serde_bytes = "0.11.6"
 digest = "0.10.3"
 workspace-hack = { version = "0.1", path = "../workspace-hack" }
+sha3 = "0.10.1"
 
 [[bench]]
 name = "crypto"
@@ -49,11 +50,11 @@ harness = false
 default = []
 celo = ["ark-ec", "ark-ff", "ark-serialize", "ark-bls12-377", "ark-std", "celo-bls"]
 copy_key = []
+zero_crypto = []
 
 [dev-dependencies]
 bincode = "1.3.3"
 criterion = "0.3.6"
-sha3 = "0.10.1"
 hex-literal = "0.3.4"
 proptest = "1.0.0"
 proptest-derive = "0.3.0"
--- a/crypto/src/ed25519.rs
+++ b/crypto/src/ed25519.rs
@@ -6,7 +6,7 @@ use serde::{de, Deserialize, Serialize};
 use serde_with::serde_as;
 use signature::{Signature, Signer, Verifier};
 use std::{
-    fmt::{self, Display},
+    fmt::{self, Debug, Display},
     str::FromStr,
 };
 
@@ -22,7 +22,7 @@ use crate::{
 /// Define Structs
 ///
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct Ed25519PublicKey(pub ed25519_dalek::PublicKey);
 
 pub type Ed25519PublicKeyBytes = PublicKeyBytes<Ed25519PublicKey, { Ed25519PublicKey::LENGTH }>;
@@ -102,6 +102,12 @@ impl Default for Ed25519PublicKey {
 impl Display for Ed25519PublicKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         write!(f, "{}", Base64::encode_string(self.0.as_bytes()))
+    }
+}
+
+impl Debug for Ed25519PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        <Self as Display>::fmt(self, f)
     }
 }
 
--- a/crypto/src/test_utils/zero_crypto.rs
+++ b/crypto/src/test_utils/zero_crypto.rs
@@ -5,7 +5,7 @@
  * This file contains an implementation of a negligible-cost (apart from some ocassional pk copying) trivial
  * cryptographic scheme. The purpose of this library is to allow seamless benchmarking of systems without
  * taking into account the cost of cryptographic primitives - and hence providing a theoretical maximal
- * throughput that a system could achieve if the cost of crypto is optimized away. 
+ * throughput that a system could achieve if the cost of crypto is optimized away.
  */
 use base64ct::{Base64, Encoding};
 use digest::Digest;
--- a/crypto/src/test_utils/zero_crypto.rs
+++ b/crypto/src/test_utils/zero_crypto.rs
@@ -48,11 +48,11 @@ pub struct ZeroKeyPair {
 }
 
 #[serde_as]
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
 pub struct ZeroSignature {}
 
 #[serde_as]
-#[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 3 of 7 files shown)


crypto/src/ed25519.rs
```diff
@@ -105,6 +105,12 @@ impl Display for Ed25519PublicKey {
     }
 }
 
+impl Debug for Ed25519PublicKey {
```
kchalkias, line 108: I wouldn't change that in this PR (better keep it focused on ZeroCrypto). We can post it in another PR (even if tiny)

crypto/Cargo.toml
```diff
@@ -49,11 +50,11 @@ harness = false
 default = []
 celo = ["ark-ec", "ark-ff", "ark-serialize", "ark-bls12-377", "ark-std", "celo-bls"]
 copy_key = []
+zero_crypto = []
```
kchalkias, line 53: I'd change the name to something else, cause zero_crypto might be confused with zero knowledge proofs.

crypto/src/test_utils/zero_crypto.rs
```diff
@@ -0,0 +1,400 @@
...
+    AggregateAuthenticator, Authenticator, EncodeDecodeBase64, KeyPair, SigningKey, ToFromBytes,
+    VerifyingKey,
+};
+use sha3::Sha3_256;
+
+///
+/// Define Structs
+///
+
+const PRIVATE_KEY_LENGTH: usize = 32;
```
kchalkias, line 37: We can use smaller keys sizes here to reduce the cost impact even further. I think the minimum we could go and ensure no collisions is 10-16 bytes for both keys. I'd pick 16.

crypto/src/test_utils/zero_crypto.rs
```diff
@@ -0,0 +1,400 @@
...
+impl Verifier<ZeroSignature> for ZeroPublicKey {
+    fn verify(&self, _msg: &[u8], _signature: &ZeroSignature) -> Result<(), signature::Error> {
+        Ok(())
+    }
+}
+
+impl<'a> From<&'a ZeroPrivateKey> for ZeroPublicKey {
+    fn from(secret: &'a ZeroPrivateKey) -> Self {
+        let mut hasher = Sha3_256::new();
+        hasher.update(secret.0);
```
kchalkias, line 144: I like the approach of using the hash of priv key to derive the pub. Although we could in theory just copy the value to the pubKey, this function is not affecting our performance, so I agree with your logic.

crypto/src/test_utils/zero_crypto.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+}
+
+impl VerifyingKey for ZeroPublicKey {
+    type PrivKey = ZeroPrivateKey;
+    type Sig = ZeroSignature;
+
+    const LENGTH: usize = PUBLIC_KEY_LENGTH;
+
+    fn verify_batch(
```
kchalkias, line 160: Please add comments to all of these empty functions, to explain in one liner why they always successfully return.

crypto/src/test_utils/zero_crypto.rs
```diff
@@ -0,0 +1,400 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+ * This file contains an implementation of a negligible-cost (apart from some ocassional pk copying) trivial
+ * cryptographic scheme. The purpose of this library is to allow seamless benchmarking of systems without
```
kchalkias, line 6: Add a WARNING note, that this scheme is 100% unsafe and should never be used in production.

crypto/src/test_utils/zero_crypto.rs
```diff
@@ -0,0 +1,400 @@
...
+impl Display for ZeroSignature {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(f, "{}", Base64::encode_string(self.as_ref()))
+    }
+}
+
+impl Authenticator for ZeroSignature {
+    type PubKey = ZeroPublicKey;
+    type PrivKey = ZeroPrivateKey;
+    const LENGTH: usize = 0;
```
kchalkias, line 208: I'd use at least one byte size vs completely empty sig. I.e. byte zero is a valid sig, and anything else is false or you completely omit checking the byte. I think it would make sense to also have a failed sig to pass some tests in the future.

crypto/src/test_utils/zero_crypto.rs
```diff
@@ -0,0 +1,400 @@
...
+
+impl KeyPair for ZeroKeyPair {
+    type PubKey = ZeroPublicKey;
+    type PrivKey = ZeroPrivateKey;
+    type Sig = ZeroSignature;
+
+    #[cfg(feature = "copy_key")]
+    fn copy(&self) -> Self {
+        ZeroKeyPair {
+            name: ZeroPublicKey(self.name.0),
```
kchalkias, line 295: One day, not in this PR, we should replace `.name` with a meaningful value (i.e., `public_key`, `private_key`)

kchalkias (commented):
Awesome work, added a few hints to improve experience, but almost ready to approve.

huitseeker (changes requested):
Features in Rust are additive: https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification

It's too easy to have this be included inadvertently. To even consider doing this I would want to 1/consider if we need this in the code base (as opposed to in a branch for running benchmarks), and 2/a panic bomb inserted at runtime that is only bypassed if a very explicit "I_KNOW_WHAT_IM_DOING" environment variable is set.

crypto/src/ed25519.rs
punwai: Oops

crypto/src/test_utils/zero_crypto.rs
punwai: Will open an issue on this

crypto/src/test_utils/zero_crypto.rs
punwai: I can add this, but just wondering why we ever care about failed sigs?

crypto/src/test_utils/zero_crypto.rs
punwai: I've kept private key to 32 bytes, as this is just helpful for deriving it from seed.

punwai:
@huitseeker Would it also work if we just had a very long feature name?

punwai:
@kchalkias ping

kchalkias (commented):
That would be an awesome feature for benchmarks and tests; probably we should port it to `fastcrypto`, then enable it for testing/benchmarking purposes only.

kchalkias (commented):
@jonas-lj I'd suggest we move the functionality of a zero cost scheme implementation to `fastcrypto` under a specific `feature` (ie `unsafe-schemes`). The purpose of such a scheme is to be able to benchmark Sui with a scheme that introduces zero computational penalty, so we could track the impact of signature operations in the overall TPS.

This would also help @joyqvq to test modular crypto's _network_ impact on a validator signing/verification design using delegation to multiple workers. TL;DR it is possible that a worker model introduces network delays (due to the cost of transferring and receiving messages to/from workers) that we might realize is more significant than the CPU resources.

huitseeker:
Moved to fastcrypto https://github.com/MystenLabs/narwhal/pull/627#pullrequestreview-1115106962

PR #637, opened 2022-08-01
[observability] Log channel capacity every 5 sec

gdanezis:
Added a periodic (5 sec) log message with the capacity of the primary core channels. In case one of them is at zero, we emit an error as this might lead to deadlock.

### DIFF
--- a/primary/src/primary.rs
+++ b/primary/src/primary.rs
@@ -30,6 +30,7 @@ use prometheus::Registry;
 use std::{
     net::Ipv4Addr,
     sync::{atomic::AtomicU64, Arc},
+    time::Duration,
 };
 use store::Store;
 use tokio::{
@@ -40,7 +41,7 @@ use tokio::{
     task::JoinHandle,
 };
 use tonic::{Request, Response, Status};
-use tracing::info;
+use tracing::{info, log::error};
 use types::{
     error::DagError, BatchDigest, BatchMessage, BincodeEncodedPayload, Certificate,
     CertificateDigest, Empty, Header, HeaderDigest, PrimaryToPrimary, PrimaryToPrimaryServer,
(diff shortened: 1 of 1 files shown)


huitseeker (dismissed):
YES!

gdanezis:
Yep feel free to refactor to use something more legit indeed

gdanezis:
Ok, it seems that the approach by which a task keeps hold of channel senders does not play well with some reconfig work -- so some tests do not terminate @asonnino

primary/src/primary.rs
```diff
@@ -103,6 +104,59 @@ impl Primary {
...
+        let mon_tx_get_block_commands = tx_get_block_commands.clone();
+        let mon_tx_batches = tx_batches.clone();
+        let mon_tx_block_removal_commands = tx_block_removal_commands.clone();
+        let mon_tx_batch_removal = tx_batch_removal.clone();
+        let mon_tx_block_synchronizer_commands = tx_block_synchronizer_commands.clone();
+        let mon_tx_certificate_responses = tx_certificate_responses.clone();
+        let mon_tx_payload_availability_responses = tx_payload_availability_responses.clone();
+        let mon_tx_state_handler = tx_state_handler.clone();
+
+        tokio::task::spawn(async move {
```
huitseeker, line 127: This doesn't work well: you want to collect the task created in this line, and put that handle (like all the task handles created in this file) in the `handles` vector at the end of this function.

gdanezis: And also need too shutdown if told so. New patch incoming.

huitseeker (approved):
Opened https://github.com/MystenLabs/narwhal/issues/655 for follow-up, thanks @gdanezis !

PR #645, opened 2022-08-01
[fix] Docker gen.validators script

akichidis:
This PR is:
* fixing the `gen.committee.py` file to assign the correct id to the workers
* changing the `gen.validators.sh` file to make prometheus parse workers as well for metrics.

### DIFF
--- a/Docker/gen.validators.sh
+++ b/Docker/gen.validators.sh
@@ -108,12 +108,18 @@ cp -r templates/{grafana,prometheus} ${target}/
 t=$(($num - 1))
 for i in $(seq -f %02g 0 ${t})
 do
-    scrape="primary_${i}:8010"
+    scrape_primary="primary_${i}:8010"
+    scrape_worker="worker_${i}:8010"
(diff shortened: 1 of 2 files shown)


huitseeker (approved):
Is there something we could have captured in the snapshot tests updated in https://github.com/MystenLabs/narwhal/pull/625 that would have helped the next developer how to make those updates in advance?
(e.g. : adding to the test : "you should run the following Docker script and check everything passes")

akichidis:
> Is there something we could have captured in the snapshot tests updated in #625 that would have helped the next developer how to make those updates in advance? (e.g. : adding to the test : "you should run the following Docker script and check everything passes")

I see what you mean @huitseeker . Off the top of my head, I believe the most straightforward way would be to have a dedicated test that runs the [python code](https://github.com/MystenLabs/narwhal/blob/main/Docker/scripts/gen.committee.py) that generates the committee.json file and compares the output like we do in the snapshot testing. Might be other more elegant ways but this will give us strong guarantees that we remain consistent

PR #649, opened 2022-08-01
fix(network): set a better initial retry delay for network

huitseeker:
Addresses https://github.com/MystenLabs/sui/issues/1529

### DIFF
--- a/network/src/retry.rs
+++ b/network/src/retry.rs
@@ -44,10 +44,10 @@ impl RetryConfig {
     // default gives 5-6 retries in ~30 s total retry time.
 
     /// Default for [`RetryConfig::max_retry_interval`] (500 ms).
-    pub const DEFAULT_INITIAL_RETRY_INTERVAL: Duration = Duration::from_millis(100);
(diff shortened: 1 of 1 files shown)


network/src/retry.rs
```diff
@@ -44,7 +44,7 @@ impl RetryConfig {
     // default gives 5-6 retries in ~30 s total retry time.
 
     /// Default for [`RetryConfig::max_retry_interval`] (500 ms).
-    pub const DEFAULT_INITIAL_RETRY_INTERVAL: Duration = Duration::from_millis(100);
+    pub const DEFAULT_INITIAL_RETRY_INTERVAL: Duration = Duration::from_millis(500);
 
     /// Default for [`RetryConfig::max_retry_interval`] (15 s).
     pub const DEFAULT_MAX_RETRY_INTERVAL: Duration = Duration::from_secs(1);
```
bmwill, line 50: While we're here do we want to bump this up too? Maybe to match the comment?

huitseeker: @bmwill done!

bmwill (approved):

PR #651, opened 2022-08-01
add rpc endpoint for querying a primary's worker information

bmwill:
In addition it changes how the grpc code is generated and moves it to be generated via a build script because otherwise you can end up in an interesting situation where you break compilation due to the proto definition changing and being unable to regenerate the generated code via the bootstrap.rs test when you have a dirty working directory.


types/proto/narwhal.proto
```diff
@@ -184,6 +184,8 @@ service WorkerToWorker {
 service WorkerToPrimary {
     // Sends a message
     rpc SendMessage(BincodeEncodedPayload) returns (Empty) {}
+
+    rpc WorkerInfo(Empty) returns (BincodeEncodedPayload) {}
```
arun-koshy, line 188: I was thinking the request would include the worker ID they are looking for and that worker address would be returned instead of the entire map but this works for now.

arun-koshy (approved):
Looks like a good start to me and thanks for moving us back to build.rs :)

types/proto/narwhal.proto
arun-koshy: Actually I take that back even on a failed request to a specific worker it would be pretty cheap to just request all the worker info and update the entire map in cache.

bmwill: Yeah for now it seems simplest to just request everything, its a small amount of data for now.

bmwill:
> thanks for moving us back to build.rs

Well its my fault for trying to be smart and go the bootstrap.rs test path, we can look at it as a semi-failed experiment :D

PR #652, opened 2022-08-01
[test] loss of liveness end to end test

akichidis:
This PR is introducing two tests for the loss of liveness and recovery. More about the tests can be found in comments. Also a small refactoring in Cluster to allow us assert nodes progress in common way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3818,6 +3818,7 @@ dependencies = [
  "executor",
  "futures",
  "indexmap",
+ "itertools",
  "multiaddr",
  "mysten-network 0.1.0 (git+https://github.com/mystenlabs/mysten-infra.git?rev=123c9e40b529315e1c1d91a54fb717111c3e349c)",
  "node",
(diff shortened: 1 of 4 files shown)


asonnino (approved):

huitseeker (commented):
This looks good to me.
Is there some guidance we could put in as to what the tests would look like when we solve #664 ? I'm thinking in particular that the expectations at the top of `test_loss_of_liveness_without_recovery` are bound to surprise a few folks on the expected behaviour. Could we help explain the difference between where we are now and where we want to be?

akichidis:
> This looks good to me. Is there some guidance we could put in as to what the tests would look like when we solve MystenLabs/sui#5315 ? I'm thinking in particular that the expectations at the top of `test_loss_of_liveness_without_recovery` are bound to surprise a few folks on the expected behaviour. Could we help explain the difference between where we are now and where we want to be?

@huitseeker I've added some extra comments in the test and referenced the issue.

PR #660, opened 2022-08-01
Preserve order of batches

### LINKED ISSUE #398: Maintain order of transactions as much as possible
See Section "(C) Sequencing.": https://www.notion.so/mystenlabs/Narwhal-Progress-d954e29f6e0d4f7ebfa3c1dd6bbcfe48

asonnino:
Fix #398


node/tests/staged/narwhal.yaml
```diff
@@ -52,10 +52,10 @@ Header:
...
     - epoch: U64
     - payload:
-        MAP:
-          KEY:
-            TYPENAME: BatchDigest
-          VALUE: U32
+        SEQ:
+          TUPLE:
+            - TYPENAME: BatchDigest
+            - U32
```
akichidis, line 58: Ah so it serialises as a sequence of tuples!

akichidis (commented):
Changes look good! Could you please rebase @asonnino before I approve?

Also have we thought in converting the `payload` to a vector?  To be honest I am not sure how much value I see this being a map. Especially if we expand this further with extra info (ex the size of the batch - this will probably be needed for Celo). Not to address in this PR, but something to think about.

akichidis (approved):

node/tests/staged/narwhal.yaml
asonnino: Yes looks like it but I agree that we can change it into a vector

PR #661, opened 2022-08-01
Clarify max_header_delay

asonnino:
People keep asking about the DAG timeout in partual-synchrony. This PR just clarifies it with a comment.


akichidis (approved):
Thanks @asonnino for putting in this comment! It also sounds that we might want to add a metric in this case so we can monitor (and alert probably) if that happens to frequently - potentially byzantine nodes and/or misconfigurations both sound pretty important

PR #662, opened 2022-08-01
[crypto] accept references in batch verification

punwai:
**Problem:** In Narwhal crypto, we used to:
Copy a Public Key every time we want to verify a signature (once from copying out of the 'committee' struct, the second time copying into the narwhal references function, which does not take in references)

This was probably fine with ed25519, where the public keys were already serialized bytes,  but BLS12381 stores public keys in some other form - which adds extra cost very time we deep-copy the public keys. Note that this happened 30 times for each certificate.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -397,7 +397,7 @@ impl Committee {
         self.authorities
             .iter()
             .filter(|(name, _)| *name != myself)
-            .map(|(name, authority)| (name.deref().clone(), authority.primary.clone()))
+            .map(|(name, authority)| (name.clone(), authority.primary.clone()))
             .collect()
     }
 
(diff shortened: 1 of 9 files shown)


config/src/lib.rs
```diff
@@ -397,7 +397,7 @@ impl Committee {
         self.authorities
             .iter()
             .filter(|(name, _)| *name != myself)
-            .map(|(name, authority)| (name.deref().clone(), authority.primary.clone()))
+            .map(|(name, authority)| (name.clone(), authority.primary.clone()))
```
punwai, line 400: Need this change, otherwise it doesn't work if you change it to BLS.

akichidis:
@punwai is this still something you want to go forward? If yes could you please the linter & tests as they seem to fail?

punwai:
@akichidis I think we'll definitely try to merge this, but after another PR goes through :)

huitseeker (changes requested):
Please add a topic or an issue this fixes.

punwai:
@huitseeker Added comments! Will ping again when this is rebased and beautified.

kchalkias:
@punwai What's the state of this one?

huitseeker:
Made obsolete by https://github.com/MystenLabs/fastcrypto/pull/90

PR #663, opened 2022-08-01
Limit empty certificates

### LINKED ISSUE #5301: Limit empty-headers
See section "(F) Validity": https://www.notion.so/mystenlabs/Narwhal-Progress-d954e29f6e0d4f7ebfa3c1dd6bbcfe48

asonnino:
Fix MystenLabs/sui#5301. This PR is delicate, please pay careful attention at the proposer's conditions to advance round. This PR is not urgent.

### DIFF
--- a/primary/src/proposer.rs
+++ b/primary/src/proposer.rs
@@ -29,7 +29,10 @@ pub mod proposer_tests;
 #[derive(Debug)]
 #[cfg_attr(test, derive(PartialEq))]
 pub enum ProposerMessage {
+    /// Sent by the core to notify the proposer that it received a quorum of certificates with
+    /// the same round number.
     NewParents(Vec<Certificate>, Round, Epoch),
+    /// Sent the by core to notify the proposer it received a non-empty certificate.
     MeaningfulRound(Round),
 }
 
@@ -43,7 +46,8 @@ pub struct Proposer {
     signature_service: SignatureService<Signature>,
     /// The size of the headers' payload.
     header_size: usize,
-    /// The maximum message delay after GST (only meaningful in partial-synchrony).
+    /// The maximum message delay after Global Stabilization Time (GST). This is only meaningful in
+    /// partial-synchrony.
     delta: Duration,
     /// The network model in which the node operates.
     network_model: NetworkModel,
@@ -72,7 +76,7 @@ pub struct Proposer {
     /// running the DAG in asynchrony).
     common_case: bool,
     /// The highest round for which we received a non-empty certificate from other peers.
-    highest_useful_round: Round,
+    highest_peers_round: Option<Round>,
 
     /// Metrics handler
     metrics: Arc<PrimaryMetrics>,
@@ -113,7 +117,7 @@ impl Proposer {
                 digests: Vec::with_capacity(2 * header_size),
                 payload_size: 0,
                 common_case: true,
-                highest_useful_round: 1,
+                highest_peers_round: None,
                 metrics,
             }
             .run()
@@ -256,19 +260,21 @@ impl Proposer {
             // Check if we have enough parents; if we have enough digests; if other primaries already
             // made a non-certificate for this round (and need our certificate to make progress); and
             // if the timer expired (only meaningful in partial-synchrony).
-            let enough_parents = !self.last_parents.is_empty();
-            let we_have_enough_digests = self.payload_size >= self.header_size;
-            let others_have_digests = self.round == self.highest_useful_round;
             let timer_expired = timer.is_elapsed();
+            let enough_parents = !self.last_parents.is_empty();
+            let we_have_enough_batches = self.payload_size >= self.header_size;
+            let others_have_batches = self
+                .highest_peers_round
+                .is_some_and(|r| self.round == r - 1);
 
             // Check whether advancing round allows the system to make meaningful progress. The system
             // makes meaningful progress if it advances with non-empty certificates.
-            let meaningful_progress = we_have_enough_digests || others_have_digests;
+            let meaningful_progress = we_have_enough_batches || others_have_batches;
 
             // Print a few debug logs.
             debug!("Do we have enough parents: {enough_parents}");
-            debug!("Do we have enough digests: {we_have_enough_digests}");
-            debug!("Have we learned about a non-empty certificate for this round: {others_have_digests}");
+            debug!("Do we have enough digests: {we_have_enough_batches}");
+            debug!("Have we learned about a non-empty certificate for this round: {others_have_batches}");
             debug!("Are we in the common case: {}", self.common_case);
             debug!("Did the timer expire: {timer_expired}");
 
@@ -338,7 +344,8 @@ impl Proposer {
 
                     },
                     ProposerMessage::MeaningfulRound(round) => {
-                        self.highest_useful_round = max(self.highest_useful_round, round);
+                        self.highest_peers_round = self.highest_peers_round
+                            .map_or_else(|| Some(round), |r| Some(max(r, round)));
                     },
                 },
 
--- a/primary/src/proposer.rs
+++ b/primary/src/proposer.rs
@@ -112,7 +112,7 @@ impl Proposer {
                 digests: Vec::with_capacity(2 * header_size),
                 payload_size: 0,
                 common_case: true,
-                highest_useful_round: 0,
+                highest_useful_round: 1,
                 metrics,
             }
             .run()
@@ -263,6 +263,13 @@ impl Proposer {
             // Check whether advancing round allows the system to make meaningful progress. The system
             // makes meaningful progress if it advances with non-empty certificates.
             let meaningful_progress = we_have_enough_digests || others_have_digests;
+
+            // Print a few debug logs.
+            debug!("Do we have enough parents: {enough_parents}");
+            debug!("Do we have enough digests: {we_have_enough_digests}");
+            debug!("Have we learned about a non-empty certificate for this round: {others_have_digests}");
+            debug!("Are we in the common case: {}", self.common_case);
+            debug!("Did the timer expire: {timer_expired}");
 
             // Check whether we can move to the next round and proposer a new header.
             if (self.common_case || timer_expired) && enough_parents && meaningful_progress {
--- a/primary/src/proposer.rs
+++ b/primary/src/proposer.rs
@@ -23,6 +23,11 @@ use types::{
 #[path = "tests/proposer_tests.rs"]
 pub mod proposer_tests;
 
+pub enum ProposerMessage {
+    NewParents(Vec<Certificate>, Round, Epoch),
+    MeaningfulRound(Round),
+}
+
 /// The proposer creates new headers and send them to the core for broadcasting and further processing.
 pub struct Proposer {
     /// The public key of this primary.
@@ -41,7 +46,7 @@ pub struct Proposer {
     /// Watch channel to reconfigure the committee.
     rx_reconfigure: watch::Receiver<ReconfigureNotification>,
     /// Receives the parents to include in the next header (along with their round number).
-    rx_core: Receiver<(Vec<Certificate>, Round, Epoch)>,
+    rx_core: Receiver<ProposerMessage>,
     /// Receives the batches' digests from our workers.
     rx_workers: Receiver<(BatchDigest, WorkerId)>,
     /// Sends newly created headers to the `Core`.
@@ -57,6 +62,10 @@ pub struct Proposer {
     digests: Vec<(BatchDigest, WorkerId)>,
     /// Keeps track of the size (in bytes) of batches' digests that we received so far.
     payload_size: usize,
+    /// Whether the dag can advance to the next round. Note that if we timeout, we ignore this
+    /// check and advance anyway.
+    advance: bool,
+
     /// Metrics handler
     metrics: Arc<PrimaryMetrics>,
 }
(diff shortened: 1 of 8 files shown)


asonnino:
See also https://github.com/MystenLabs/narwhal/pull/598

primary/src/proposer.rs
```diff
@@ -201,26 +220,60 @@ impl Proposer {
...
-            // (ii) we have enough digests (minimum header size) and we are on the happy path (we can vote for
-            // the leader or the leader has enough votes to enable a commit). The latter condition only matters
-            // in partially synchrony.
+            // Check if we have enough parents; if we have enough digests; if other primaries already
+            // made a non-certificate for this round (and need our certificate to make progress); and
+            // if the timer expired (only meaningful in partial-synchrony).
             let enough_parents = !self.last_parents.is_empty();
-            let enough_digests = self.payload_size >= self.header_size;
+            let we_have_enough_digests = self.payload_size >= self.header_size;
+            let others_have_digests = self.round == self.highest_useful_round;
```
akichidis, line 261: So let's take for example the following situation in a cluster of 4 validators (with ids 1,2,3,4). They all start from `round 0`. 
* Let's assume that `validator 1` is having enough batches and proposes a `new header at round 1`. The header is sent to everyone, they vote and a certificate is being created & broadcasted. 
* Now all the nodes (validators 2,3,4) receive the certificate and update their `highest_useful_round` in their proposers to `1`. 
* Let's assume that none of those validators have enough batches so they won't create any new headers by their own.  
* The validators (2,3,4) proposers  are still in `self.round == 0` and the `self.highest_useful_round == 1`
* Because of that the `meaningful_progress` will be `false` (as the check is `others_have_digests = self.round == self.highest_useful_round` 
* None of the validators 2,3,4 will create a new header to help validator 1 progress, unless they start receiving some batches their own and eventually create a header.

@asonnino assuming that I got the above right it seems to me that this is not desired behaviour, right?

akichidis (commented):
This is just a first pass, I want to look again on this. @asonnino left a few comments, let me know your thoughts

primary/src/proposer.rs
```diff
@@ -33,15 +43,15 @@ pub struct Proposer {
     signature_service: SignatureService<Signature>,
     /// The size of the headers' payload.
     header_size: usize,
-    /// The maximum delay to wait for batches' digests.
-    max_header_delay: Duration,
+    /// The maximum message delay after GST (only meaningful in partial-synchrony).
```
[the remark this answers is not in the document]
asonnino: Fixed

primary/src/proposer.rs
```diff
@@ -57,6 +67,13 @@ pub struct Proposer {
     digests: Vec<(BatchDigest, WorkerId)>,
     /// Keeps track of the size (in bytes) of batches' digests that we received so far.
     payload_size: usize,
+    /// Whether we are on the happy path: i.e., we are after GST and the leader is not Byzantine.
+    /// Note that `common_case` is only used in partial synchrony (it is alway set to `true` when
+    /// running the DAG in asynchrony).
+    common_case: bool,
+    /// The highest round for which we received a non-empty certificate from other peers.
+    highest_useful_round: Round,
```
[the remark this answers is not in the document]
asonnino: fixed

primary/src/proposer.rs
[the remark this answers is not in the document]
asonnino: fixed

primary/src/proposer.rs
```diff
@@ -23,6 +26,13 @@ use types::{
 #[path = "tests/proposer_tests.rs"]
 pub mod proposer_tests;
 
+#[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+pub enum ProposerMessage {
+    NewParents(Vec<Certificate>, Round, Epoch),
+    MeaningfulRound(Round),
```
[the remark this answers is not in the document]
asonnino: fixed

primary/src/proposer.rs
asonnino: mmh, it seems I didn't push the latest commit, checking now

asonnino: yes this was not the latest it seems, I pushed fixes to your comments along with the latest version

primary/src/proposer.rs
```diff
@@ -201,27 +224,63 @@ impl Proposer {
...
+
+            // Print a few debug logs.
+            debug!("Do we have enough parents: {enough_parents}");
+            debug!("Do we have enough digests: {we_have_enough_batches}");
+            debug!("Have we learned about a non-empty certificate for this round: {others_have_batches}");
+            debug!("Are we in the common case: {}", self.common_case);
+            debug!("Did the timer expire: {timer_expired}");
+
+            // Check whether we can move to the next round and proposer a new header.
+            if (self.common_case || timer_expired) && enough_parents && meaningful_progress {
```
huitseeker, line 283: So we go from 1.
```rust
(timer_expired || (we_have_enough_digests && common_case)) && enough_parents
```
to 2.
```rust
(common_case || timer_expired) && enough_parents && (we_have_enough_digests || others_have_batches)
```

1. is equivalent to
```rust
(common_case || timer_expired) && enough_parents && (we_have_enough_digests || timer_expired)
```
Which means you're just replacing `timer_expired` by `others_have_batches` in the third clause.

Fair enough.

Now here's my problem: assume everybody else, for some reason, proposes a valid but empty header, and I as a validator do not have batches. Nobody proposes. Then the others start receiving transactions.

Don't we lose liveness?

Can we accuse a number of other nodes of something specific? Which nodes?

I think we might need another timer.

types/src/primary.rs
```diff
@@ -433,6 +433,10 @@ impl Certificate {
     pub fn origin(&self) -> PublicKey {
         self.header.author.clone()
     }
+
+    pub fn is_empty(&self) -> bool {
+        self.header.payload.is_empty()
+    }
```
huitseeker, line 439: Can we add 
1. a scary warning on the definition of `is_empty` saying it is now load-bearing for the liveness of the protocol, and its semantics should check there are >0 transactions in the header?
2. can we add a test for this function?
3. can we open an issue noting that it would be enough to create empty batches, to fool folks into thinking headers are not? we need to length-prefix our hashes anyway, this makes me think we want a `tx_count` in the header to resolve this.

primary/src/tests/proposer_tests.rs
```diff
@@ -73,15 +81,66 @@ async fn propose_payload() {
...
-    // Ensure the proposer makes a correct header from the provided payload.
+    // Ensure the proposer makes a valid proposal
     let header = rx_headers.recv().await.unwrap();
     assert_eq!(header.round, 1);
     assert_eq!(header.payload.get(&digest), Some(&worker_id));
     assert!(header.verify(&committee(None)).is_ok());
 }
+
+#[tokio::test]
+async fn advance_to_help() {
```
huitseeker, line 104: I would like a test to explore what happens when most nodes submit empty headers as well.

huitseeker (commented):
This is interesting, and I like the PR, especially for its constrained behavior over #598. But I have questions over whether we can consider the behavior of proposing an empty block as clearly Byzantine.

primary/src/proposer.rs
asonnino: You are right, if this buggy behaviour happens I don't see how we can create a fraud proof. However I don't think we have fraud proofs for liveness attacks anyway (e.g. what do we do if buggy nodes simply stop creating headers?).

Let's keep thinking about this (I would not recommend landing this PR until we are sure DevNet works anyway).

Proposing as soon as f+1 nodes propose empty or non-empty header may not work. Upon booting, there are 2f+1 empty certificates (genesis); this means all honest nodes will propose their first (possibly empty) header; from there it seems they will always have 2f+1 (possibly empty) headers triggering nodes to move forward.

primary/src/proposer.rs
```diff
@@ -57,6 +71,13 @@ pub struct Proposer {
     digests: Vec<(BatchDigest, WorkerId)>,
     /// Keeps track of the size (in bytes) of batches' digests that we received so far.
     payload_size: usize,
+    /// Whether we are on the happy path: i.e., we are after GST and the leader is not Byzantine.
+    /// Note that `common_case` is only used in partial synchrony (it is alway set to `true` when
+    /// running the DAG in asynchrony).
+    common_case: bool,
+    /// The highest round for which we received a non-empty certificate from other peers.
+    highest_peers_round: Option<Round>,
```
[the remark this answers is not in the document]
asonnino: fixed

primary/src/proposer.rs
```diff
@@ -201,27 +224,63 @@ impl Proposer {
...
+            // Check whether advancing round allows the system to make meaningful progress. The system
+            // makes meaningful progress if it advances with non-empty certificates.
+            let meaningful_progress = we_have_enough_batches || others_have_batches;
+
+            // Print a few debug logs.
+            debug!("Do we have enough parents: {enough_parents}");
+            debug!("Do we have enough digests: {we_have_enough_batches}");
+            debug!("Have we learned about a non-empty certificate for this round: {others_have_batches}");
+            debug!("Are we in the common case: {}", self.common_case);
+            debug!("Did the timer expire: {timer_expired}");
```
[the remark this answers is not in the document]
asonnino: fixed

primary/src/proposer.rs
```diff
@@ -33,15 +46,16 @@ pub struct Proposer {
     signature_service: SignatureService<Signature>,
     /// The size of the headers' payload.
     header_size: usize,
-    /// The maximum delay to wait for batches' digests.
-    max_header_delay: Duration,
+    /// The maximum message delay after Global Stabilization Time (GST). This is only meaningful in
+    /// partial-synchrony.
+    delta: Duration,
```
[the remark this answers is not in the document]
asonnino: fixed

primary/src/proposer.rs
```diff
@@ -57,6 +71,13 @@ pub struct Proposer {
     digests: Vec<(BatchDigest, WorkerId)>,
     /// Keeps track of the size (in bytes) of batches' digests that we received so far.
     payload_size: usize,
+    /// Whether we are on the happy path: i.e., we are after GST and the leader is not Byzantine.
+    /// Note that `common_case` is only used in partial synchrony (it is alway set to `true` when
+    /// running the DAG in asynchrony).
+    common_case: bool,
```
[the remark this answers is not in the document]
asonnino: fixed

primary/src/proposer.rs
huitseeker: > Proposing as soon as f+1 nodes propose empty or non-empty header may not work. Upon booting, there are 2f+1 empty certificates (genesis); this means all honest nodes will propose their first (possibly empty) header; from there it seems they will always have 2f+1 (possibly empty) headers triggering nodes to move forward.

You seem to be reasoning on empty rounds in the round that precedes the proposal (the parents). I'm suggesting reasoning on counting the proposals of the current round.

Let me put it this way, you're proposing this for gating the proposal condition:
```
(common_case || timer_expired) && enough_parents && (we_have_enough_digests || others_have_batches)
```

Could we not recover from the case where all "committed to empty" with the following condition:
```
(common_case || timer_expired) && enough_parents && (we_have_enough_digests || others_have_batches || (timer_expired && f_plus_1_proposed_this_round))
```
?

types/src/primary.rs
huitseeker: See also https://github.com/MystenLabs/narwhal/issues/517

primary/src/proposer.rs
asonnino: Ha I see. Yes, I believe this would fix the scenario you described

types/src/primary.rs
asonnino: I added the ping in the proposer directly, where you thinking to put it in the function you linked?

primary/src/proposer.rs
huitseeker: I believe this 6-variable clause will need really, really didactic commenting around it :D

types/src/primary.rs
huitseeker: I'm not sure I understand the question?

asonnino: ha, that is because it makes no sense :)

I mean `ping -> print`. I added a tracing/log print in the proposer to warn about empty headers

primary/src/proposer.rs
asonnino: So this will fix the issue you described, but it also mean that if there are no transactions we keep making empty certificates? (Although at a lower pace, dictated by 'timer_expired').

Let's say there are no client transactions in the system and that all authorities are honest.
At round 0:
-  Once the first timer expires: (common_case || timer_1_expired) = true
- Since all nodes have the genesis: enough_parents = true
- when the second timer expires, since all nodes have f+1 genesis blocks (round 0): (timer_2_expired && f_plus_1_proposed_this_round)) = true

huitseeker: > since all nodes have f+1 genesis blocks
How is that leading to `f_plus_1_proposed_this_round == true`?

asonnino: The current round is round 0; do you mean the "next_round"?

huitseeker: Yes.

akichidis:
@asonnino closing this as we have moved repo under SUI https://github.com/MystenLabs/sui. Please re-open PR there if it's still relevant

PR #668, opened 2022-08-02
network: migrate the primary-to-primary interface to a quic p2p networking stack (anemo)

bmwill:
Brief summary of the motivation for this change:
- Http (and gRPC) are built with one side of a connection being a client and the other being the server. Only clients are able to initial requests. This leads to requiring two nodes to maintain two connections between each other in order to communicate between each other.
- Due to the above, we can run into network reachability issues where a connection can be established from Node A -> B but not B -> A. Due to the way that narwhals communication is designed this leads to buffering up messages that cannot be delivered for some time.
- This new stack enables both sides of a connection to initiate Requests, enabling us to use 1 connection for all communication.
- Today we communicate in the clear with no authentication, QUIC requires TLS and as such we'll now how authentication and encryption by default (this would be possible with grpc but the libs we use make this more difficult since we don't rely on a central CA)
- Anemo is a p2p networking stack and as such has better utilities around answering questions like "how many connections do i have open", "who am i connected with", coupled with the ability to actually drop a connection from a peer (which today is very difficult or impossible with our current stack)
- Better connection pooling, today we actually make lots of connections between two nodes, with the new stack we'll better be able to share connections.
- QUIC has the potential to enable higher throughput and latency due to solving the head-of-line blocking problem that TCP can suffer from when multiplexing multiple streams of requests over the same connection

### DIFF
--- a/network/Cargo.toml
+++ b/network/Cargo.toml
@@ -21,11 +21,15 @@ tokio-util = { version = "0.7.3", features = ["codec"] }
 tonic = { version = "0.7.2", features = ["tls"] }
 tracing = "0.1.36"
 types = { path = "../types" }
+crypto = { path = "../crypto" }
 
 serde = "1.0.144"
 mysten-network = { git = "https://github.com/mystenlabs/mysten-infra.git", rev = "9deac015fbf66e24b6da9699630e06750eaa094a" }
 workspace-hack = { version = "0.1", path = "../workspace-hack" }
 eyre = "0.6.8"
+
+anemo = { git = "https://github.com/mystenlabs/anemo.git", rev = "6278d0fa78147a49ff2cb9dd2e45e763886be0a0" }
+anyhow = "1.0.62"
 
 [dev-dependencies]
 bincode = "1.3.3"
--- a/network/src/primary.rs
+++ b/network/src/primary.rs
@@ -3,182 +3,26 @@
 
 use crate::{
     metrics::{Metrics, PrimaryNetworkMetrics},
-    traits::{BaseNetwork, LuckyNetwork, ReliableNetwork, UnreliableNetwork},
-    BoundedExecutor, CancelOnDropHandler, MessageResult, RetryConfig, MAX_TASK_CONCURRENCY,
+    traits::{BaseNetwork, Lucky, ReliableNetwork2, UnreliableNetwork, UnreliableNetwork2},
+    BoundedExecutor, CancelOnDropHandler, RetryConfig, MAX_TASK_CONCURRENCY,
 };
+use anemo::PeerId;
 use async_trait::async_trait;
+use crypto::PublicKey;
 use multiaddr::Multiaddr;
 use rand::{rngs::SmallRng, SeedableRng as _};
 use std::collections::HashMap;
 use tokio::{runtime::Handle, task::JoinHandle};
-use tonic::{transport::Channel, Code};
-use tracing::error;
+use tonic::transport::Channel;
 use types::{
     BincodeEncodedPayload, PrimaryMessage, PrimaryToPrimaryClient, PrimaryToWorkerClient,
     PrimaryWorkerMessage,
 };
 
-pub struct PrimaryNetwork {
-    clients: HashMap<Multiaddr, PrimaryToPrimaryClient<Channel>>,
-    config: mysten_network::config::Config,
-    retry_config: RetryConfig,
-    /// Small RNG just used to shuffle nodes and randomize connections (not crypto related).
-    rng: SmallRng,
-    // One bounded executor per address
-    executors: HashMap<Multiaddr, BoundedExecutor>,
-    metrics: Option<Metrics<PrimaryNetworkMetrics>>,
-}
-
 fn default_executor() -> BoundedExecutor {
     BoundedExecutor::new(MAX_TASK_CONCURRENCY, Handle::current())
 }
 
-impl Default for PrimaryNetwork {
-    fn default() -> Self {
-        let retry_config = RetryConfig {
-            // Retry for forever
-            retrying_max_elapsed_time: None,
-            ..Default::default()
-        };
-
-        Self {
-            clients: Default::default(),
-            config: Default::default(),
-            retry_config,
-            rng: SmallRng::from_entropy(),
-            executors: HashMap::new(),
-            metrics: None,
-        }
-    }
-}
-
-impl PrimaryNetwork {
-    pub fn new(metrics: Metrics<PrimaryNetworkMetrics>) -> Self {
-        Self {
-            metrics: Some(Metrics::from(metrics, "primary".to_string())),
-            ..Default::default()
-        }
-    }
-
-    fn update_metrics(&self) {
-        if let Some(m) = &self.metrics {
-            for (addr, executor) in &self.executors {
-                m.set_network_available_tasks(
-                    executor.available_capacity() as i64,
-                    Some(addr.to_string()),
-                );
-            }
-        }
-    }
-
-    pub fn cleanup<'a, I>(&mut self, to_remove: I)
-    where
-        I: IntoIterator<Item = &'a Multiaddr>,
-    {
-        for address in to_remove {
-            self.clients.remove(address);
-        }
-    }
-}
-
-impl BaseNetwork for PrimaryNetwork {
-    type Client = PrimaryToPrimaryClient<Channel>;
-
-    type Message = PrimaryMessage;
-
-    fn client(&mut self, address: Multiaddr) -> Self::Client {
-        self.clients
-            .entry(address.clone())
-            .or_insert_with(|| Self::create_client(&self.config, address))
-            .clone()
-    }
-
-    fn create_client(config: &mysten_network::config::Config, address: Multiaddr) -> Self::Client {
-        //TODO don't panic here if address isn't supported
-        let channel = config.connect_lazy(&address).unwrap();
-        PrimaryToPrimaryClient::new(channel)
-    }
-}
-
-#[async_trait]
-impl UnreliableNetwork for PrimaryNetwork {
-    async fn unreliable_send_message(
-        &mut self,
-        address: Multiaddr,
-        message: BincodeEncodedPayload,
-    ) -> JoinHandle<()> {
-        let mut client = self.client(address.clone());
-        let handle = self
-            .executors
-            .entry(address)
-            .or_insert_with(default_executor)
-            .spawn(async move {
-                let _ = client.send_message(message).await;
-            })
-            .await;
-
-        self.update_metrics();
-
-        handle
-    }
-}
-
-#[async_trait]
-impl LuckyNetwork for PrimaryNetwork {
-    fn rng(&mut self) -> &mut SmallRng {
-        &mut self.rng
-    }
-}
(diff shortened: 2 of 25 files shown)


andll:
I understand that there probably was a lot of discussion offline, but I think it could still be worth to put some summary into this PR given how large change this is.

Particularly those things would be interesting(at lest to me):
- Problems that we see with current stack and how new stack fixes them
- Are there any metrics that are likely to improve in devnet after this diff is deployed?

bmwill:
> I understand that there probably was a lot of discussion offline, but I think it could still be worth to put some summary into this PR given how large change this is.

I'm happy to chat more offline if you need more context. The plan is to eventually replace the stack in Sui as well.

> Particularly those things would be interesting(at lest to me):
> 
> * Problems that we see with current stack and how new stack fixes them

updated the PR summary

> * Are there any metrics that are likely to improve in devnet after this diff is deployed?

This is only part 1 of a long migration but i would expect that the number of buffered messages we see internally to drop and have a lower steady state during times of network instability (starting a network or restarting a node). It may require completion of all the interfaces before things are much better though.

network/Cargo.toml
```diff
@@ -21,12 +21,16 @@ tokio-util = { version = "0.7.3", features = ["codec"] }
...
 tracing = "0.1.36"
 types = { path = "../types" }
+crypto = { path = "../crypto" }
 
 serde = "1.0.144"
 mysten-network = { git = "https://github.com/mystenlabs/mysten-infra.git", rev = "696fa25982acdc87ab64ec6a33cb111edc63a331" }
 workspace-hack = { version = "0.1", path = "../workspace-hack" }
 eyre = "0.6.8"
 
+anemo = { git = "https://github.com/bmwill/anemo.git", rev = "677af751e4c9b5f9822c7d2d5f05aa5f925d871c" }
```
dmitri-perelman, line 31: Why did you choose to push this part to a separate repo? (I might be missing context here)

bmwill: I've been working on this on the side for a while now, it was easier to have it be somewhere simple for me to work on. I expect it may make sense to move this elsewhere (ie to the mysten org) at some point

bmwill: I do think that it makes sense to be in its own repo, similar to the crypto stuff being separated, though

dmitri-perelman: Why not to keep it in `mysten-infra`? No strong opinions, up to you guys.

We'd definitely want to move it to Mysten org to have proper code reviews / CI etc.

network/src/primary.rs
```diff
@@ -263,3 +107,119 @@ impl UnreliableNetwork for PrimaryToWorkerNetwork {
...
+            rng: SmallRng::from_entropy(),
+            executors: HashMap::new(),
+        }
+    }
+
+    pub fn cleanup<'a, I>(&mut self, _to_remove: I)
+    where
+        I: IntoIterator<Item = &'a Multiaddr>,
+    {
+        // do nothing atm
```
aschran, line 140: Confused what this is? Should it maybe be turned into a TODO or removed?

network/src/traits.rs
```diff
@@ -104,3 +105,80 @@ pub trait ReliableNetwork: BaseNetwork {
...
+
+    /// Broadcasts a message to all `peers` passed as an argument.
+    /// The attempts to send individual messages are best effort and will not be retried.
+    async fn unreliable_broadcast(
+        &mut self,
+        peers: Vec<PublicKey>,
+        message: &Message,
+    ) -> Vec<JoinHandle<()>> {
+        let mut handlers = Vec::new();
+        for peer in peers {
```
aschran, line 121: Would it be more idiomatically rust-y to do something functional here, like make this whole function body just

peers.into_iter().map(<call to unreliable_send>).collect()

Or is that too opaque or not possible for some reason? (asking mostly for my own rust edification). The nice aspect of it would be that it's just one line, and maybe faster if it enables pre-sizing the new vector correctly, etc.

network/src/traits.rs
```diff
@@ -104,3 +105,80 @@ pub trait ReliableNetwork: BaseNetwork {
...
+
+#[async_trait]
+pub trait LuckyNetwork2<Message> {
+    /// Pick a few addresses at random (specified by `nodes`) and try (best-effort) to send the
+    /// message only to them. This is useful to pick nodes with whom to sync.
+    async fn lucky_broadcast(
+        &mut self,
+        mut peers: Vec<PublicKey>,
+        message: &Message,
+        nodes: usize,
```
aschran, line 141: I think a slightly more verbose name like `num_nodes` would be clearer here (unless this is already following an established convention?)

Or maybe even `num_peers` or `num_targets`? Curious why the word "nodes" - does that have a specific meaning in this context?

network/src/traits.rs
```diff
@@ -104,3 +105,80 @@ pub trait ReliableNetwork: BaseNetwork {
...
+    T: UnreliableNetwork2<M> + Send,
+    T: Lucky,
+{
+    async fn lucky_broadcast(
+        &mut self,
+        mut peers: Vec<PublicKey>,
+        message: &M,
+        nodes: usize,
+    ) -> Vec<JoinHandle<()>> {
+        peers.shuffle(self.rng());
```
aschran, line 158: Would `partial_shuffle` be better here to avoid shuffling stuff you're just gonna truncate?

(Or `choose_multiple`, which seems more natural, but would require reworking all these interfaces to use iterators instead of vecs? Curious if that's worth considering, or if not then how did you come to decide to pass around vecs?)

network/src/traits.rs
```diff
@@ -104,3 +105,80 @@ pub trait ReliableNetwork: BaseNetwork {
...
+        peer: PublicKey,
+        message: &Message,
+    ) -> CancelOnDropHandler<anyhow::Result<anemo::Response<()>>>;
+
+    async fn broadcast(
+        &mut self,
+        peers: Vec<PublicKey>,
+        message: &Message,
+    ) -> Vec<CancelOnDropHandler<anyhow::Result<anemo::Response<()>>>> {
+        let mut handlers = Vec::new();
```
aschran, line 177: Same question about a functional version of this

primary/src/tests/helper_tests.rs
```diff
@@ -147,23 +209,38 @@ async fn test_process_certificates_batch_mode() {
...
-    let address = committee.primary(&name).unwrap();
-    let mut handler = PrimaryToPrimaryMockServer::spawn(address.primary_to_primary);
+    let address = committee
+        .primary(&requestor_name)
+        .unwrap()
+        .primary_to_primary;
+    let requestor_key = primary_keys.pop().unwrap();
+    let (mut handler, _network) = PrimaryToPrimaryMockServer::spawn(requestor_key, address);
+
+    // Wait for connectivity
```
aschran, line 219: Wondering if this is worth factoring out to a common test util? It seems to appear in several places here & other files

aschran (changes requested):

network/src/primary.rs
```diff
@@ -263,3 +107,119 @@ impl UnreliableNetwork for PrimaryToWorkerNetwork {
...
+
+#[async_trait]
+impl ReliableNetwork2<PrimaryMessage> for PrimaryNetwork {
+    async fn send(
+        &mut self,
+        peer: PublicKey,
+        message: &PrimaryMessage,
+    ) -> CancelOnDropHandler<anyhow::Result<anemo::Response<()>>> {
+        // Safety
+        // Since this spawns an unbounded task, this should be called in a time-restricted fashion.
```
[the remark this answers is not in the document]
bmwill: These comments were just moved so i didn't add them :)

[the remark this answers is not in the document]
bmwill: Same here

network/src/primary.rs
bmwill: Yeah i'll add a TODO

network/src/traits.rs
bmwill: Unfortunately that doesn't work with the async call in the for loop otherwise i would prefer the method you've suggested.

network/src/traits.rs
bmwill: each entity is a "node" in the network, but yeah i'll change to num_nodes.

network/src/traits.rs
bmwill: This code was really just moved I don't know the exact motivation for how it was done here. I think this works for now but we can revisit and clean it up once the networking migration has finished.

network/src/traits.rs
bmwill: Same issue, async makes this not work :/

primary/src/tests/helper_tests.rs
bmwill: I think there's actually quite a bit of work that would be done to refactor lots of the common testing boilerplate we currently have. Given this is a large change already (and there's more to come) it might make sense to wait till the dust settles so we have a better base to refactor from.

network/Cargo.toml
bmwill: I've moved the repo to mystenlabs/anemo

primary/src/tests/helper_tests.rs
aschran: SGTM!

network/src/primary.rs
```diff
@@ -263,3 +107,117 @@ impl UnreliableNetwork for PrimaryToWorkerNetwork {
...
+    network: anemo::Network,
+    retry_config: RetryConfig,
+    /// Small RNG just used to shuffle nodes and randomize connections (not crypto related).
+    rng: SmallRng,
+    // One bounded executor per address
+    executors: HashMap<anemo::PeerId, BoundedExecutor>,
+}
+
+impl PrimaryNetwork {
+    pub fn new(network: anemo::Network) -> Self {
```
arun-koshy, line 121: What are your thoughts on handling the `anemo::Network` construction in this method and only passing the `NetworkIdentity` as arguments? i.e.

`pub fn new(myself: NetworkIdentity, peers: Vec<NetworkIdentity>) -> Self {`

```
struct NetworkIdentity {
    key_pair: KeyPair,
    address: MultiAddr
}
```

arun-koshy (approved):
Thanks Brandon! Just some minor comments, but other than that LGTM 🚀

primary/tests/epoch_change.rs
```diff
@@ -465,6 +469,8 @@ async fn test_restart_with_new_committee_change() {
 
         // Wait for the committee to shutdown.
         join_all(handles).await;
+        // Provide a small amount of time for any background tasks to shutdown
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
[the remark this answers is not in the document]
bmwill: The network shuts down when all the handles to it are dropped. After the join above we we need some delay for the background network task to unbind from the network port otherwise the test will fail when we try to startup a new network and bind on the same port.

network/src/primary.rs
bmwill: I don't think that'll work since we only want a single anemo network instance where are multiple components all have their own "PrimaryNetwork" handles. Some point soon we'll also have the other interfaces all operating on top of a single instance of an anemo network so that will need to be built prior to this so that it can be shared.

PR #682, opened 2022-08-04
[metrics] Makes all channels metered

### LINKED ISSUE #655: Track channel utilization using metrics
We're about to merge https://github.com/MystenLabs/narwhal/pull/637 which tracks channel capacity by logging it every 5s.

We'd rather take inspiration from https://github.com/diem/diem/blob/latest/common/channel/src/lib.rs, which creates `IntGauge` instances per channel for the occupancy of their butter. Note this channel wrapper is based on the std::sync::mpsc, while we rather use tokio::sync::mpsc.

In order to the integrate this in our metrics, we could:
1. Add the IntGauge metric in the [PrimaryMetrics](https://github.com/MystenLabs/narwhal/blob/d1498bac211bd875f020d6f71e9539e88217f208/primary/src/metrics.rs#L72)
2. We initialize and retrieve the metrics already [here](https://github.com/MystenLabs/narwhal/blob/d1498bac211bd875f020d6f71e9539e88217f208/primary/src/primary.rs#L113) (channels will need to be initialised after the metrics)
3. Then we do have two options (a) pass the PrimaryMetrics in the [new method](https://github.com/diem/diem/blob/37fc28b3fa5b1ef8e9463c150e29ef60078cd327/common/channel/src/lib.rs#L115) (b) we just clone and pass the gauge directly. We probably want to do the (a) for consistency

We could also use #637 but just emitting the metrics in gauge. And we’ll need a label with the channel name there so we can differentiate the values.

(thanks to @akichidis for a lot of that direction)

huitseeker:
This is a continuation of #637. Previously, we were logging the capacity of most (but not all) channels in the primary on a periodic basis.

This PR introduces a new primitive, `types::metered_channel`, which wraps a `tokio::mpsc::{Sender, Receiver}` and a `prometheus::IntGauge` in a way distantly inspired by `diem::common::channel`. 

Then it uses it in all the top-level channels involved in the primary, and exposes those metrics through the registry of the primary.

Closes #655


types/src/metered_channel.rs
```diff
@@ -0,0 +1,295 @@
...
+pub struct Receiver<T> {
+    inner: mpsc::Receiver<T>,
+    gauge: IntGauge,
+}
+
+impl<T> Receiver<T> {
+    /// Receives the next value for this receiver.
+    /// Decrements the gauge in case of a successful `recv`.
+    pub async fn recv(&mut self) -> Option<T> {
+        self.inner.recv().inspect(|_| self.gauge.dec()).await
```
akichidis, line 42: I believe the `inspect` will get executed even the result is `None` - if that's the case then we shouldn't probably decrement in this case. Can we have a test for this unhappy case?

types/src/metered_channel.rs
```diff
@@ -0,0 +1,295 @@
...
+impl<T> Unpin for Receiver<T> {}
+
+impl<T> Sender<T> {
+    /// Sends a value, waiting until there is capacity.
+    /// Increments the gauge in case of a successful `send`.
+    pub async fn send(&self, value: T) -> Result<(), SendError<T>> {
+        (*self)
+            .inner
+            .send(value)
+            .inspect(|_| self.gauge.inc())
```
akichidis, line 86: Similar with the sender, if `inspect` gets executed even in the case of an `Err` result then we shouldn't increment the value

types/src/metered_channel.rs
```diff
@@ -0,0 +1,295 @@
...
+        self.inner.closed().await
+    }
+
+    /// Attempts to immediately send a message on this `Sender`
+    /// Increments the gauge in case of a successful `try_send`.
+    pub fn try_send(&self, message: T) -> Result<(), TrySendError<T>> {
+        (*self)
+            .inner
+            .try_send(message)
+            // remove this unsightly hack once https://github.com/rust-lang/rust/issues/91345 is resolved
```
akichidis, line 101: Probably this is still not relevant as issue seems to be closed.

types/src/metered_channel.rs
```diff
@@ -0,0 +1,295 @@
...
+        self.inner.is_closed()
+    }
+
+    /// Waits for channel capacity. Once capacity to send one message is
+    /// available, it is reserved for the caller.
+    /// Increments the gauge in case of a successful `reserve`.
+    pub async fn reserve(&self) -> Result<Permit<'_, T>, SendError<()>> {
+        self.inner
+            .reserve()
+            // remove this unsightly hack once https://github.com/rust-lang/rust/issues/91345 is resolved
```
akichidis, line 124: Same comment with above

types/src/metered_channel.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    /// Waits for channel capacity, moving the `Sender` and returning an owned
+    /// permit. Once capacity to send one message is available, it is reserved
+    /// for the caller.
+    /// Increments the gauge in case of a successful `reserve_owned`.
+    pub async fn reserve_owned(self) -> Result<OwnedPermit<T>, SendError<()>> {
+        self.inner
+            .reserve_owned()
+            // remove this unsightly hack once https://github.com/rust-lang/rust/issues/91345 is resolved
```
akichidis, line 139: Same with above

types/src/metered_channel.rs
```diff
@@ -0,0 +1,295 @@
...
+        self.inner.try_reserve().map(|val| {
+            // remove this unsightly hack once https://github.com/rust-lang/rust/issues/91345 is resolved
+            self.gauge.inc();
+            val
+        })
+    }
+
+    /// Tries to acquire a slot in the channel without waiting for the slot to become
+    /// available, returning an owned permit.
+    ///      Increments the gauge in case of a successful `try_reserve`.
```
akichidis, line 160: Nit:  ...case of a successful `try_reserve` --> `try_reserve_owned`

akichidis (commented):
Great work @huitseeker !! Can't wait to see this in prod. Left a few comments, let me know your thoughts

types/src/metered_channel.rs
gdanezis: I suggest we simply read the `capacity` and stick it in the gauge rather that trying to carefully march the capacity using `inc` and `dec` which seems fragile to me.

akichidis: Good thought @gdanezis . Reporting capacity might also give us some nice queue semantics (which we could calc in grafana anyways) of "how close am I to deplete?"

types/src/metered_channel.rs
huitseeker: I'm confused: this link works and is still current in the Rust documentation.

types/src/metered_channel.rs
huitseeker: Same as well

types/src/metered_channel.rs
huitseeker: Same as well

primary/src/primary.rs
```diff
@@ -72,114 +72,94 @@ impl Primary {
...
+            &primary_channel_metrics.tx_certificate_responses,
+        );
+        let (tx_payload_availability_responses, rx_payload_availability_responses) = channel(
+            CHANNEL_CAPACITY,
+            &primary_channel_metrics.tx_payload_availability_responses,
+        );
+        let (tx_state_handler, rx_state_handler) =
+            channel(CHANNEL_CAPACITY, &primary_channel_metrics.tx_state_handler);
+        // TODO: this channel pair name is highly counterintuitive: see initialization in node and rename(?)
+        let (tx_committed_certificates, rx_consensus) = metered_channel::wrap(
```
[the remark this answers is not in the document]
huitseeker: Yep, that's correct, I'll deal with this channel from the node side.

primary/src/metrics.rs
```diff
@@ -35,11 +41,189 @@ pub(crate) fn initialise_metrics(metrics_registry: &Registry) -> Metrics {
...
         node_metrics: Some(node_metrics),
         endpoint_metrics: Some(endpoint_metrics),
+        primary_channel_metrics: Some(primary_channel_metrics),
         primary_endpoint_metrics: Some(primary_endpoint_metrics),
         network_metrics: Some(network_metrics),
     }
 }
 
+#[derive(Clone)]
+pub struct PrimaryChannelMetrics {
```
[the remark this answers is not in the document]
huitseeker: I like the many metrics to maintain in code: they actually add friction to the use of channels, which we have too many of already.
Moreover, note two important things:
1. we are not metering capacity, but occupancy,
2. at the moment, our metrics all have the same maximum capacity. I expect that like in any Tokio application, we will shortly make these capacities diverge. As a consequence, the numbers between two metrics would no longer be comparable.

node/src/lib.rs
```diff
@@ -132,7 +132,15 @@ impl Node {
...
-        let (tx_new_certificates, rx_new_certificates) = channel(Self::CHANNEL_CAPACITY);
+        // This gauge is porcelain: do not modify it without also modifying `primary::metrics::PrimaryChannelMetrics::replace_registered_new_certificates_metric`
+        // This hack avoids a cyclic dependency in the initialization of consensus and primary
+        let new_certificates_counter = IntGauge::new(
+            "tx_new_certificates",
+            "occupancy of the channel from the `primary::Core` to the `Consensus`",
+        )
+        .unwrap();
+        let (tx_new_certificates, rx_new_certificates) =
+            types::metered_channel::channel(Self::CHANNEL_CAPACITY, &new_certificates_counter);
```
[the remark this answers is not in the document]
huitseeker: Here's the issue: even if I introduce it as consensus metrics, we cannot make this part of the initialization of consensus, because we would need to:
- create the channel in the node, with its associated metric
- pass the sending end to the Consensus, and probably re-register the associated metric there,
- and then pass the receiving end of the channel to the primary.

This is the exact symmetric of what we're doing now:
- create the channel in the node initialization, with its associated metric
- pass the sending end to the consensus,
- and then pass the receiving end of the channel to the primary, re-register the associated metric there,

I'm happy to let you have a go at it in a follow-up PR, if you have an idea of how to clean this up.

types/src/metered_channel.rs
```diff
@@ -0,0 +1,295 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+#![allow(dead_code)] // TODO: complete tests - This kinda sorta facades the whole tokio::mpsc::{Sender, Receiver}: without tests, this will be fragile to maintain.
```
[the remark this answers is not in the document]
huitseeker: I'll add some tests and just remove the owned variants, because we will definitely use the `reserve` methods soon (for backpressure propagation).

akichidis (approved):
Thanks so much @huitseeker for doing this, very valuable! 💯

types/src/metered_channel.rs
huitseeker: Except this will only work if all channels have the same capacity, which they really will not have for long. A consequence of having percolating backpressure (my next PR) will be to see the need for different queue sizes.

types/src/metered_channel.rs
huitseeker: You're right, fixed.

PR #686, opened 2022-08-04
Simplify executor-sui interface

asonnino:
Now that we know more about how Narwhal is interfacing with Sui, we can simplify a bit the Executor interface.


akichidis (approved):
Changes look good to me , nice simplification! 👍

Issue #566, opened 2022-07-22
Cleanup Network


asonnino:
### Description
Most of our networks (PrimaryToWorkerNetwork, PrimaryNetwork, WorkerNetwork) have a HashMap of address -> clients. When the committee changes and the network addresses change, the networks will simply expand this hash map but it is never cleaned up.

huitseeker:
And we introduced a parallel `HashMap` of `BoundedExecutor`s in #559 indeed.

PR #687, opened 2022-08-04
Fix network memory leaks

asonnino:
Fix memory leaks  in the network stack when updating the committee (or changing epoch)

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -13,7 +13,7 @@ use crypto::{traits::EncodeDecodeBase64, PublicKey};
 use multiaddr::Multiaddr;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap, HashSet},
     fs::{self, OpenOptions},
     io::{BufWriter, Write as _},
     ops::Deref,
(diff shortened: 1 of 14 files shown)


network/src/primary.rs
```diff
@@ -73,6 +73,15 @@ impl PrimaryNetwork {
...
+
+    pub fn cleanup<'a, I>(&mut self, to_remove: I)
+    where
+        I: IntoIterator<Item = &'a Multiaddr>,
+    {
+        for address in to_remove {
+            self.clients.remove(address);
+        }
+    }
 }
```
akichidis, line 85: Could you please add tests for this behaviour?

akichidis (commented):
I would like to have a look again on this, but don't have many comments. Tried to think of how we could probably implement in a different way (seems to me that we do similar stuff in many places) but can't think of something right now. In the meanwhile could you please sync with @arun-koshy as he is looking to extra the worker addresses from committee? It seems that you touch some similar stuff.

network/src/primary.rs
asonnino: @huitseeker Do you think this function could be in a nice trait? It is used by 3 out of 4 of our network structs

bmwill (approved):

network/src/primary.rs
huitseeker: @asonnino The natural place to add it is the `BaseNetwork` trait.

config/src/lib.rs
```diff
@@ -451,6 +451,45 @@ impl Committee {
...
+                            .map(|address| &address.primary_to_worker),
+                    )
+            })
+            .flatten()
+            .collect()
+    }
+
+    /// Return the network addresses that are present in the current committee but that are absent
+    /// from the new committee (provided as argument).
+    pub fn network_diff<'a>(&'a self, new_committee: &'a Self) -> HashSet<&Multiaddr> {
```
huitseeker, line 486: Nit: `new` and `old` are concepts for the caller of this method. The typical way to present things in such a method is `self` and `other` (for the argument).

huitseeker (commented):
Same comment on `get_all_network_addresses`, otherwise LGTM. Please link to #566, which you're closing.

huitseeker:
@asonnino You'll want to rebase as we have fixed your `cargo-deny` issue on main.

network/src/primary.rs
asonnino: But not all networks can have this functions. To be precise, they all need it except `WorkerToPrimary` that only has a single client.

PR #691, opened 2022-08-05
Simplify executor-consensus interface

asonnino:
Simplify the consensus-executor interface by removing a sync that is not used. This sync was originally added when we planned to run the executor (and Sui) on a different machine than the consensus


executor/src/subscriber.rs
```diff
@@ -185,29 +84,26 @@ impl Subscriber {
             tokio::select! {
                 // Receive the ordered sequence of consensus messages from a consensus node.
                 Some(message) = self.rx_consensus.recv() => {
-                    // Process the consensus message (synchronize missing messages, download transaction data).
-                    let sequence = self.handle_consensus_message(&message).await?;
```
huitseeker, line 189: I am so happy to see this line go away.

huitseeker (approved):
- a big red diff with -700 lines,
- removing a complex completion loop,
- and a store,
- and a blocking call (see comment) on the direct line to the output of consensus (hence backpressuring it),
- and we need none of that.

... what's not to like? 🤩 🤩 🤩

LefKok (approved):

executor/src/subscriber.rs
```diff
@@ -185,29 +84,26 @@ impl Subscriber {
...
-                    self.next_consensus_index += sequence.len() as SequenceNumber;
-
-                    // Wait for the transaction data to be available in the store. We will then execute the transactions.
-                    for message in sequence {
-                        let digests = message.certificate.header.payload.keys().cloned().collect();
-                        let future = Self::waiter(digests, self.store.clone(), message);
-                        waiting.push(future);
-                    }
+                    // Send the certificate to the batch loader to download all transactions' data.
+                    self.tx_batch_loader
```
gdanezis, line 88: Is this a metered channel? Are we sure if this blocks it will not take down a tokio select that is required for liveness? Should we make it print a warn then error if the delay is too long? (I have become paranoid now).

executor/src/subscriber.rs
```diff
@@ -185,29 +84,26 @@ impl Subscriber {
...
+                        .send(message.clone())
+                        .await
+                        .expect("Failed to send message ot batch loader");
+
+                    // Wait for the transaction data to be available in the store. We will then forward these
+                    // transactions to the Executor Core for execution.
```
gdanezis, line 94: Maybe tell us a little more in this comment how this is guaranteed to happen, and therefore liveness if safe?

node/src/lib.rs
```diff
@@ -300,16 +284,14 @@ impl Node {
             store.batch_store.clone(),
             execution_state,
             tx_reconfigure,
-            /* rx_consensus */ rx_consensus_to_client,
-            /* tx_consensus */ tx_client_to_consensus,
```
akichidis, line 304: Ah much nicer now!

executor/src/subscriber.rs
```diff
@@ -185,29 +84,26 @@ impl Subscriber {
...
                 },
 
                 // Receive here consensus messages for which we have downloaded all transactions data.
-                Some(message) = waiting.next() => self
-                    .tx_executor
-                    .send(message?)
-                    .await
-                    .map_err(|_| SubscriberError::ExecutorConnectionDropped)?,
+                Some(message) = waiting.next() => {
+                    if self.tx_executor.send(message?).await.is_err() {
```
gdanezis, line 102: Same questions about sending as above.

gdanezis (approved):
Very happy to see this simplification.

akichidis (approved):
Red is good! Thanks @asonnino for simplifying this 💯

executor/src/subscriber.rs
asonnino: It is indeed not metered, but it seems that the last PR of @huitseeker fixes this. The batch loader can only block if the network blocks.

PR #705, opened 2022-08-06
Avoid blocking `tokio::select` branches on a potent. full channel

huitseeker:
Checkpointing some (incomplete) backpressure work on top of #691.
This PR is an incremental improvement.

## Context

We often have branches of a `tokio::select` that fire, and then send a message on a channel as part of their immediate processing. This message is sent with `sender.send(message).await`, which does not return immediately. 

## The issue

This will yield to the executor, but not let the rest of the `select` resume.

## The solution

We introduce a macro that helps provide semantics by which a `select` branch fires if it has received an input for its condition (typically a `waiting.next()` or `receiver.recv()`) *and* can acquire a permit to send a message for a downstream channel (typically a `sender.send(result)`). This gives the other branches in the `select` a chance to fire in case the downstream channel is full, as the macro-using branch will then not fire. The macro meshes better than an async function with [the unusual borrowing behavior of the `select` macro](https://tokio.rs/tokio/tutorial/select#borrowing).

### DIFF
--- a/executor/src/errors.rs
+++ b/executor/src/errors.rs
@@ -21,10 +21,31 @@ macro_rules! ensure {
     };
 }
 
+#[macro_export]
+macro_rules! try_fut_and_permit {
+    ($fut:expr, $sender:expr) => {
+        futures::future::TryFutureExt::unwrap_or_else(
+            futures::future::try_join(
+                $fut,
+                futures::TryFutureExt::map_err($sender.reserve(), |_e| {
+                    SubscriberError::ClosedChannel(stringify!(sender).to_owned())
+                }),
+            ),
+            |e| {
+                tracing::error!("{e}");
+                panic!("I/O failure, killing the node.");
+            },
+        )
+    };
+}
+
 pub type SubscriberResult<T> = Result<T, SubscriberError>;
 
 #[derive(Debug, Error, Clone)]
 pub enum SubscriberError {
+    #[error("channel {0} closed unexpectedly")]
+    ClosedChannel(String),
+
     #[error("Storage failure: {0}")]
     StoreError(#[from] StoreError),
 
--- a/executor/src/subscriber.rs
+++ b/executor/src/subscriber.rs
@@ -1,11 +1,11 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
-use crate::errors::{SubscriberError, SubscriberResult};
-use consensus::ConsensusOutput;
-use futures::{
-    future::try_join_all,
-    stream::{FuturesOrdered, StreamExt},
+use crate::{
+    errors::{SubscriberError, SubscriberResult},
+    try_fut_and_permit,
 };
+use consensus::ConsensusOutput;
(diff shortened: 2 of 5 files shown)


executor/src/errors.rs
```diff
@@ -21,10 +21,31 @@ macro_rules! ensure {
     };
 }
 
+#[macro_export]
+macro_rules! try_fut_and_permit {
+    ($fut:expr, $sender:expr) => {
+        futures::future::TryFutureExt::unwrap_or_else(
+            futures::future::try_join(
+                $fut,
+                futures::TryFutureExt::map_err($sender.reserve(), |_e| {
```
gdanezis, line 30: Very elegant idea to use the reserve here. I am not fully fluent with macros: this will keep and return what is provided by the reservation (permit) until the send happens right?

executor/src/subscriber.rs
```diff
@@ -83,12 +82,9 @@ impl Subscriber {
...
             tokio::select! {
                 // Receive the ordered sequence of consensus messages from a consensus node.
-                Some(message) = self.rx_consensus.recv() => {
+                (Some(message), permit) = try_fut_and_permit!(self.rx_consensus.recv().map(Ok), self.tx_batch_loader) => {
                     // Send the certificate to the batch loader to download all transactions' data.
-                    self.tx_batch_loader
-                        .send(message.clone())
-                        .await
-                        .expect("Failed to send message ot batch loader");
+                    permit.send(message.clone());
```
gdanezis, line 87: I confirmed the answer to my own question above is "yes". All is good!

executor/src/errors.rs
```diff
@@ -21,10 +21,31 @@ macro_rules! ensure {
...
+                    SubscriberError::ClosedChannel(stringify!(sender).to_owned())
+                }),
+            ),
+            |e| {
+                tracing::error!("{e}");
+                panic!("I/O failure, killing the node.");
+            },
+        )
+    };
+}
```
gdanezis, line 40: Lets stick this macro in the utils crate, and re-use it in places in Sui where we risk blocking select loops.

types/src/error.rs
```diff
@@ -23,10 +23,31 @@ macro_rules! ensure {
     };
 }
 
+#[macro_export]
+macro_rules! try_fut_and_permit {
```
gdanezis, line 27: Do we need to define twice?

gdanezis (approved):
I think this is a very elegant solution to the tokio::select loop blocking potential issue.

executor/src/subscriber.rs
```diff
@@ -83,12 +82,9 @@ impl Subscriber {
         loop {
             tokio::select! {
                 // Receive the ordered sequence of consensus messages from a consensus node.
-                Some(message) = self.rx_consensus.recv() => {
+                (Some(message), permit) = try_fut_and_permit!(self.rx_consensus.recv().map(Ok), self.tx_batch_loader) => {
```
akichidis, line 85: So @huitseeker should we expect on every iteration to try and acquire a permit for the requested channel? I am trying to wrap my head around of how the underlying join will work here on the select branch.

asonnino (approved):
NIce!

huitseeker:
This was closed by mistake, will re-open.

PR #707, opened 2022-08-07
Use `TaskGroup` to ensure all primary / worker tasks are cancelled on error and panic

mwtian:
Put all JobHandle of primary, worker and consensus into a [TaskGroup](https://crates.io/crates/task-group), so if one task returns an error or panics, the whole group will be cancelled.

If we want to stop the primary or worker, we can just drop the `TaskManager` associated with the `TaskGroup`.

For cluster tests, I'm assuming JobHandles will be finish on their own, i.e. checking where a JobHandle `is_finished()` is equivalent to checking if it has been cancelled.


node/src/lib.rs
```diff
@@ -231,7 +232,12 @@ impl Node {
             });
         }
 
-        Ok(handles)
+        let (task_group, task_manager) = TaskGroup::new();
+        for (name, handle) in handles {
+            let _ = task_group.spawn(name, handle);
```
asonnino, line 237: what would be returned here?

asonnino: what are we spawning exactly?

node/src/lib.rs
```diff
@@ -341,8 +347,12 @@ impl Node {
                 store.batch_store.clone(),
                 metrics.clone(),
             );
-            handles.extend(worker_handles);
+            // TODO: propagate worker task names if needed.
```
asonnino, line 350: should this be an issue?

node/src/lib.rs
mwtian: Added a comment. Basically a future that can be awaited as the JoinHandle is returned. But the returned future is unnecessary because we will use the `task_manager` to await for task terminations.

node/src/lib.rs
mwtian: Opened an issue and linked here. If other aspects of this PR look ok, I can make the change in this PR too.

mwtian:
Another potential refactor is to simplify shutdown handling: instead of each task implement logic to handle `ReconfigureNotification::Shutdown`, we may be able to handle it in one place to shutdown all tasks.

huitseeker (commented):
Sorry for the late review! This LGTM modulo a rebase,
>  Another potential refactor is to simplify shutdown handling: instead of each task implement logic to handle ReconfigureNotification::Shutdown, we may be able to handle it in one place to shutdown all tasks.

I think the goal of the `ReconfigureNotification::Shutdown` is to allow more ad-hoc graceful shutdown logic than killing the task.

mwtian:
> Sorry for the late review! This LGTM modulo a rebase,
> 
> > 
> 
> Another potential refactor is to simplify shutdown handling: instead of each task implement logic to handle ReconfigureNotification::Shutdown, we may be able to handle it in one place to shutdown all tasks.
> 
> I think the goal of the `ReconfigureNotification::Shutdown` is to allow more ad-hoc graceful shutdown logic than killing the task.

Makes sense.

huitseeker (approved):
TY!

PR #711, opened 2022-08-07
feat: Make all Worker channels metered

huitseeker:
This is an extension of #682 to the worker

### DIFF
--- a/worker/src/synchronizer.rs
+++ b/worker/src/synchronizer.rs
@@ -14,15 +14,13 @@ use std::{
 };
 use store::{Store, StoreError};
 use tokio::{
-    sync::{
-        mpsc::{channel, Receiver, Sender},
-        watch,
-    },
+    sync::{mpsc, watch},
     task::JoinHandle,
     time::{sleep, Duration, Instant},
 };
 use tracing::{debug, error};
 use types::{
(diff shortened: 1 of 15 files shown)


gdanezis (approved):
Awesome conversion!

worker/src/synchronizer.rs
```diff
@@ -164,7 +162,7 @@ impl Synchronizer {
 
                             // Add the digest to the waiter.
                             let deliver = digest;
-                            let (tx_cancel, rx_cancel) = channel(1);
+                            let (tx_cancel, rx_cancel) = mpsc::channel(1);
```
akichidis, line 165: Ah probably those could be oneshot channels? (not concern of this PR of course - but something to keep in mind)

akichidis (approved):
Great work @huitseeker , thanks for doing this!

worker/src/synchronizer.rs
huitseeker: They usually can't: this is a cancellation handler, and the branches of the `tokio::select` that GC and those that unpack items in waiting race each other. The GC might then try to cancel an item that's already delivered (and the delivery sinks the receiving end of the channel). With a oneshot channel, that may lead to a panic.

Issue #706, opened 2022-08-07
[executor] Route the executor requests without the `worker_to_worker` interface


huitseeker:
* the `Executor`'s [`BatchLoader`](https://github.com/MystenLabs/narwhal/blob/main/executor/src/batch_loader.rs) is using the network to retrieve batches from its own worker,
* that's too bad, because we have a [`BlockWaiter`](https://github.com/MystenLabs/narwhal/blob/main/primary/src/block_waiter.rs) that allows a primary to retrieve payload from any particular block, feeding itself both on its own and on others' workers. 
* but the *very* problematic thing with this is that the `BatchLoader` is [using the ](https://github.com/MystenLabs/narwhal/blob/main/executor/src/batch_loader.rs#L151)`WorkerToWorkerClient` to issue these batch requests, which means that primary is impersonating a worker, in order to contact its own worker, and [using the publicly network interface of its worker](https://github.com/MystenLabs/narwhal/blob/9d66ac0ef14ea4754c4147dd6ff533a55e522513/executor/src/lib.rs#L144-L161) (as opposed to the loopback `primary_to_worker` interface).
* When the primary and worker are co-located, this round-trip through the exterior could be much more expensive than accessing a loopback interface.
* More importantly it would mean that *the response packets coming out of the co-located worker are in a contest with the other requests coming from the outside*. This would explain encountering the error messages "`Failed to receive batch reply from worker `".

Conclusion: I think we should excise the `WorkerToWorker` network from the `BatchLoader`. A way to do that seems to be to mimic some of the behavior of the `BlockWaiter` to retrieve solely batches (rather than blocks). /cc @akichidis for better ideas.

akichidis:
@huitseeker totally onboard in swapping the `batch_loader` with the `block_waiter`. By using `block_waiter` we also get on top transparent synchronization capabilities for missing batches. I don't believe we need to mimic block_waiter, we can just go ahead and use it. The `executor` is using the `batch_loader` to essentially request a "block" it self - meaning, all the batches associated with a certificate - and that's what we exactly do in the `block_waiter`.

PR #715, opened 2022-08-07
fix: Mitigate issue #706

huitseeker:
## Context:
The `BatchLoader` is establishing a primary->worker communication using a public address, which is adding latency and competing with outbound traffic.
The proper fix us to use the `BlockWaiter`, see #706.

## The issue:
We would like a mitigation to be deployed and effective sooner.

## The mitigation:
We inspect the worker addresses used by the `BatchLoader`, and rewrite their hostname to localhost when it matches the local primary.

### DIFF
--- a/executor/src/lib.rs
+++ b/executor/src/lib.rs
@@ -19,6 +19,7 @@ mod execution_state;
 mod sequencer;
 
 pub use errors::{ExecutionStateError, SubscriberError, SubscriberResult};
+use multiaddr::{Multiaddr, Protocol};
 pub use state::ExecutionIndices;
 
 use crate::{batch_loader::BatchLoader, core::Core, subscriber::Subscriber};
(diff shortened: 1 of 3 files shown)


tharbert:
@huitseeker this connectivity should work as expected:

```
root@sui-validator-0:/sui# host localhost
localhost has address 127.0.0.1
localhost has IPv6 address ::1
Host localhost not found: 3(NXDOMAIN)

root@sui-validator-0:/sui# nc localhost 8084
@
^C
```

`@` is indicative of a grpc service listening as expected

we will verify again in staging

LefKok (approved):

gdanezis (approved):
This is a good hack for the moment indeed!

executor/src/lib.rs
```diff
@@ -169,3 +200,73 @@ impl Executor {
         ])
     }
 }
+
+fn replace_distant_by_localhost(target: &mut Multiaddr, hostname_pattern: &Protocol) {
```
akichidis, line 204: Maybe logging the successful replace can give us in staging/devnet a quick feedback whether we have successfully overwritten the address

akichidis (approved):
Thanks for that @huitseeker ! 👍

asonnino (approved):

Issue #712, opened 2022-08-07
[consensus, observability] Make the Consensus & Executor channels metered


huitseeker:
This is the follow-up of #682 that makes Consensus & Executor channels metered.
See also #711 . Opening for visibility of ongoing work.

PR #716, opened 2022-08-08
[metrics] channel metrics in consensus and executor

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/712

### DIFF
--- a/consensus/src/metrics.rs
+++ b/consensus/src/metrics.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 use prometheus::{
     default_registry, register_int_counter_with_registry, register_int_gauge_vec_with_registry,
-    IntCounter, IntGaugeVec, Registry,
+    register_int_gauge_with_registry, IntCounter, IntGauge, IntGaugeVec, Registry,
 };
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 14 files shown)


consensus/src/subscriber.rs
```diff
@@ -28,11 +25,11 @@ pub struct SubscriberHandler {
     /// Receive reconfiguration update.
     rx_reconfigure: watch::Receiver<ReconfigureNotification>,
     // Channel to receive the output of consensus.
-    rx_sequence: Receiver<ConsensusOutput>,
+    rx_sequence: metered_channel::Receiver<ConsensusOutput>,
```
gdanezis, line 28: Maybe you want to make the `metered_channel` `Sender` / `Receiver` the default you use in this scope, unless you also use a lot of `mpsc`?

gdanezis (approved):
Awesome to be able to get metrics from these channels.

I think this is the 3rd PR that converts channels to metered channels. Is it worth putting the mpsc Send / Rec functions into the disallow list of clippy?

consensus/src/subscriber.rs
akichidis: Good point, will do that!

akichidis:
> Awesome to be able to get metrics from these channels.
> 
> I think this is the 3rd PR that converts channels to metered channels. Is it worth putting the mpsc Send / Rec functions into the disallow list of clippy?

Good idea @gdanezis . I would like to see first whether we can go as far and eliminate the use of mpsc Send/Rec in our interface with SUI (because they'll have to comply) , but we can for sure do that

consensus/src/metrics.rs
```diff
@@ -61,3 +61,28 @@ impl Default for ConsensusMetrics {
         Self::new(default_registry())
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct ChannelMetrics {
+    /// occupancy of the channel from the `Consensus` to `SubscriberHandler`
```
huitseeker, line 67: We might want to add a comment telling the reader to go look at the `tx_committed_certificates` and `tx_new_certificates` in the primary as well, as those are also relevant.

huitseeker (approved):
Tip top!
While I'm in favor of any measures that simplify imports, I think it's impossible to convert all channels to `metered` and that we shouldn't spend too much time on it: we use a number of ephemeral channels for cancellation, that we should not meter.

PR #724, opened 2022-08-08
feat: Avoid blocking tokio::select branches on a potent. full channel

huitseeker:
This is a re-do of #705.
Checkpointing some (incomplete) backpressure work on top of #691.
This PR is an incremental improvement.

- This was previously reviewed as https://github.com/MystenLabs/narwhal/pull/705. 
- The problem that PR tried to solve was once the outgoing queue was full, you'd block the other branches of the `tokio::select`,
- The issue with that PR was that once you fixed that, our `waiting : Futures{Un,O}rdered` would fill up indefinitely,
- This new version of the PR creates `BoundedFutures{Un,O}rdered` waiting buffers, gives them (arguably) sensible defaults, and uses preconditions to gate select branches.

## Todo

- [x] bound the waiting futures in the subscriber,
- [x] bound the waiting futures in the certificate_waiter,
- [x] bound the waiting futures in the header_waiter,

### DIFF
--- a/types/src/bounded_future_queue.rs
+++ b/types/src/bounded_future_queue.rs
@@ -1,0 +1,250 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use futures::{
+    stream::{FuturesOrdered, FuturesUnordered},
+    Future, TryFutureExt, TryStreamExt,
+};
+use tokio::sync::{AcquireError, Semaphore, SemaphorePermit};
+
+pub struct UnorderedPermit<'a, T: Future> {
+    permit: SemaphorePermit<'a>,
+    futures: &'a BoundedFuturesUnordered<T>,
+}
+
+/// An async-friendly FuturesUnordered of bounded size. In contrast to a bounded channel,
+/// the queue makes it possible to modify and remove entries in it. In contrast to a FuturesUnordered,
+/// the queue makes it possible to enforce a bound on the number of items in the queue.
+pub struct BoundedFuturesUnordered<T: Future> {
+    /// The maximum number of entries allowed in the queue
+    capacity: usize,
+    /// The actual items in the queue.
+    queue: FuturesUnordered<T>,
+    /// This semaphore has as many permits as there are empty spots in the
+    /// `queue`, i.e., `capacity - queue.len()` many permits
+    push_semaphore: Semaphore,
+}
+
+impl<T: Future> std::fmt::Debug for BoundedFuturesUnordered<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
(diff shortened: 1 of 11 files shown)


types/src/bounded_future_queue.rs
```diff
@@ -0,0 +1,256 @@
...
+
+pub struct UnorderedPermit<'a, T: Future> {
+    permit: SemaphorePermit<'a>,
+    futures: &'a BoundedFuturesUnordered<T>,
+}
+
+/// An async-friendly FuturesUnordered of bounded size. In contrast to a bounded channel,
+/// the queue makes it possible to modify and remove entries in it. In contrast to a FuturesUnordered,
+/// the queue makes it possible to enforce a bound on the number of items in the queue.
+pub struct BoundedFuturesUnordered<T: Future> {
```
akichidis, line 17: I believe that probably a good next step (for another PR) on those structures would be to introduce metrics as well to monitor the available capacity as we do with channels - we might want to get notified when becoming saturated . Let me know your thoughts

types/src/bounded_future_queue.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+}
+
+impl<U, V, T: Future<Output = Result<U, V>>> BoundedFuturesUnordered<T> {
+    /// Creates a future that attempts to resolve the next item in the stream.
+    /// If an error is encountered before the next item, the error is returned instead.
+    pub fn try_next(&mut self) -> impl Future<Output = Result<Option<U>, V>> + '_ {
+        self.queue.try_next().inspect_ok(|val| {
+            if val.is_some() {
+                self.push_semaphore.add_permits(1)
```
akichidis, line 86: Why do we want to add a permit only in case of successful response?

akichidis (approved):
This is excellent work @huitseeker 💯 !! Left a couple of nit comments

types/src/bounded_future_queue.rs
huitseeker: Yep, happy to introduce a metric even in this PR.

types/src/bounded_future_queue.rs
huitseeker: Because this is a [`TryStreamExt::try_next`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_next), so the semantics of this returning an `Option::None` are the same as [`StreamExt::next`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.next), i.e. the stream is empty. In which case its available permits is already at capacity, and adding more permits would raise it above the capacity.

huitseeker: In [53a2a28](https://github.com/MystenLabs/narwhal/pull/724/commits/53a2a28af5faa6e6e75475c1616010823afba137) I've edited tests to reflect what happens with the capacity when you do `try_next()` past the end of the stream. I checked those tests do not pass if you remove the corresponding `if val.is_some()` guards.

PR #728, opened 2022-08-08
[metrics] adding metrics for batch maker

akichidis:
A repair item from the recent incident #7 , we want to have visibility on the created batches per worker. On top we are tracking the size of them (using a histogram metric).


asonnino (approved):

huitseeker (approved):
LGTM

PR #729, opened 2022-08-08
crypto: expose verify with hashed message as input in library

joyqvq:
https://github.com/MystenLabs/sui/issues/3844

exposing the verify hashed message API, so that Sui Move can use this to implement native function as part of the API


kchalkias (approved):
Thanks, yes this is needed for some dapps

PR #733, opened 2022-08-10
[crypto] zeroize bls12381 secrets

punwai:
Currently, deleted secret keys are kept in memory after they are dropped. This could be a problem as it could potentially later be recovered by an attacker.  This PR implements zeroize on all BLS12381 secrets (KP and privkey), and includes tests that ensure that the secret key is deleted once secrets are dropped in the program.


crypto/src/bls12381.rs
```diff
@@ -403,7 +404,7 @@ impl KeyPair for BLS12381KeyPair {
     }
 
     fn private(self) -> Self::PrivKey {
-        self.secret
+        BLS12381PrivateKey::from_bytes(self.secret.as_ref()).unwrap()
```
punwai, line 407: required, because we implemented drop for KeyPair.

crypto/src/tests/bls12381_tests.rs
```diff
@@ -400,3 +403,75 @@ async fn signature_service() {
...
+
+    // Check that self.bytes is zeroized
+    unsafe {
+        let mut vec = Vec::new();
+        for i in 0..BLS12381PrivateKey::LENGTH {
+            vec.push(*bytes_ptr.add(i));
+        }
+        assert_ne!(&vec[..], &sk_bytes[..]);
+    }
+}
```
huitseeker, line 443: YMMV, but something like 
`let memory: &[u8] = unsafe { ::std::slice::from_raw_parts(ptr, BLS12381PrivateKey::LENGTH) }` may just give you something easier to inspect, no?

crypto/src/tests/bls12381_tests.rs
```diff
@@ -400,3 +403,75 @@ async fn signature_service() {
     // Verify the signature we received.
     assert!(pk.verify(digest.as_ref(), &signature).is_ok());
 }
+
+// Checks if the private keys zeroed out
+#[test]
+fn test_zeroization_priv_key() {
+    let kp = keys().pop().unwrap();
+    let mut sk = kp.private();
```
huitseeker, line 411: Are we sure `keys` delivers some non-zero bytes in the first place for `sk`? Are we sure that invariant will remain the same forever if somebody modifies how `keys()` works?

crypto/src/tests/bls12381_tests.rs
```diff
@@ -400,3 +403,75 @@ async fn signature_service() {
...
+    let mut sk_bytes = Vec::new();
+
+    {
+        let kp = keys().pop().unwrap();
+        let sk = kp.private();
+        sk_bytes.extend_from_slice(sk.as_ref());
+
+        ptr = std::ptr::addr_of!(sk.privkey) as *const u8;
+        bytes_ptr = &sk.as_ref()[0] as *const u8;
+    }
```
huitseeker, line 460: I for one am OK testing just zeroize on drop, and inferring that if it works on drop, it works everywhere.
But I won't fight against the inclusion of one more test, of course.

huitseeker (commented):
Left a few indications, but this globally looks v. good.

crypto/src/tests/bls12381_tests.rs
punwai: I think that makes sense, drop calls zeroize itself so it makes no sense to test manually calling zeroize.

punwai: I've changed the PR to only zeroize on drop

crypto/src/tests/bls12381_tests.rs
punwai: I've added a deterministic key generation within the test itself.

crypto/src/tests/bls12381_tests.rs
punwai: Thanks for this! I think it is much cleaner - I haven't touched unsafe rust much so I just didn't know about this.

huitseeker (approved):
LGTM!

PR #734, opened 2022-08-10
Consensus log message improvements

huitseeker:
Improves ~two~ a log messages:
- "TooOld"
- ~"Leader has/hasn't got enough support"~ this one actually doesn't need editing: you can figure it out from the certificate display


akichidis (approved):
LGTM 👍

PR #738, opened 2022-08-10
[refactor] use block_waiter instead of batch_loader 

akichidis:
Resolves https://github.com/MystenLabs/narwhal/issues/706

This PR is:
* deleting the `batch_loader` in executor
* introducing the channels to use the `block_waiter` to fetch the blocks/payloads
* introducing the `temp_batch_store` instead of reusing the `batch_store` where the later is meant to be used from the worker nodes. Also on the `temp_batch_store` we have the ability to store the unserialised `Batch` directly leading to simplification.
* adding retry logic to fetch the blocks until successfully fetched (or permanently fail because of irrecoverable error)
* adding `end to end test` to confirm the consensus output is successfully received
* removing the code to replace the worker addresses part of the fix https://github.com/MystenLabs/narwhal/pull/715

**NOTE:** Please review this first / as well https://github.com/MystenLabs/narwhal/pull/746


executor/src/subscriber.rs
```diff
@@ -1,95 +1,127 @@
...
+            // Something really wrong is going on with the block waiter
+            tx_get_block_commands.send(command).await?;
+
+            let result = receiver.await.map_err(|err| PayloadRetrieveError(id))?;
+
+            match result {
+                Ok(block) => {
+                    // we successfully received the payload. Now let's add to store
+                    store
+                        .write_all(block.batches.iter().map(|b| (b.id, b.transactions.clone())))
```
bmwill, line 89:
```suggestion
                        .write_all(block.batches.into_iter().map(|b| (b.id, b.transactions)))
```

executor/src/subscriber.rs
```diff
@@ -1,95 +1,103 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
```
akichidis, line 1: Theoretically I could eliminate the `Subscriber` as well and embed the functionality in the `executor::Core` . However, the separation with the `Subscriber` offers us buffering capabilities, ie if the `Core` for some reason stuck we can still continue downloading the certificates' batches and output in the `tx_executor` channel, effectively saving time. Of course if `tx_executor` fills up we'll stall as well, but that's desired.

asonnino: we may not need a separate thread for the subscriber but I would keep the logic in the core as simple as possible (so Subscriber may remain but perhaps not spawn its own tokio task?)

executor/src/subscriber.rs
```diff
@@ -1,95 +1,103 @@
...
         rx_consensus: metered_channel::Receiver<ConsensusOutput>,
-        tx_batch_loader: metered_channel::Sender<ConsensusOutput>,
         tx_executor: metered_channel::Sender<ConsensusOutput>,
     ) -> JoinHandle<()> {
+        let get_block_retry_policy = ExponentialBackoff {
+            initial_interval: Duration::from_millis(500),
+            randomization_factor: backoff::default::RANDOMIZATION_FACTOR,
+            multiplier: backoff::default::MULTIPLIER,
+            max_interval: Duration::from_secs(10), // Maximum backoff is 10 seconds
+            max_elapsed_time: None, // Never end retrying unless a non recoverable error occurs.
```
asonnino, line 58: If an unrecoverable error happens, we need to panic and stop processing further transactions.

executor/src/subscriber.rs
```diff
@@ -1,95 +1,103 @@
...
-    }
-
     /// Main loop connecting to the consensus to listen to sequence messages.
     async fn run(&mut self) -> SubscriberResult<()> {
+        // It's important to have the futures in ordered fashion as we want
+        // to guarantee that will deliver to the executor the certificates
+        // in the same order we received from rx_consensus. So it doesn't
+        // mater if we somehow managed to fetch the batches from a later
+        // certificate. Unless the earlier certificate's payload has been
+        // fetched, no later certificate will be delivered.
```
asonnino, line 84: good comment

executor/src/subscriber.rs
```diff
@@ -1,95 +1,103 @@
...
                 rx_reconfigure,
                 rx_consensus,
-                tx_batch_loader,
                 tx_executor,
+                tx_get_block_commands,
+                get_block_retry_policy,
             }
             .run()
             .await
             .expect("Failed to run subscriber")
```
asonnino, line 73: ha cool, if an unrecoverable error happens we shut down here.

asonnino (approved):

executor/src/subscriber.rs
akichidis: Exactly, please also see the test [here](https://github.com/MystenLabs/narwhal/pull/738/files#diff-ff908d558501ea2dad76cf598def17b4e8fa49c04205323a81f896d81fd4cc42R174) where I am explicitly checking that subscriber will crash in this case.

You can also see the recoverable / irrecoverable cases [here](https://github.com/MystenLabs/narwhal/pull/738/files#diff-dd570646b526219ac76c68c9546526c43df1bf27085a0949d1355a15bd5d84b8R129) . Errors that are mapped to `Transient` errors will make the retrier retry again. Errors that are marked as `Permanent` will make the retrier stop and eventually make the subscriber crash.

executor/src/subscriber.rs
akichidis: hhm what do you mean no spawn it's own tokio task? I mean, if we don't want to make subscriber block we _need to_ run our loop in a task. Unless you mean something different?

asonnino: We can either keep the subscriber as it is (in this PR) or make it a collection of functions that are called by the Executor Core

primary/src/metrics.rs
```diff
@@ -69,7 +69,7 @@ pub struct PrimaryChannelMetrics {
     pub tx_primary_messages: IntGauge,
     /// occupancy of the channel from the `primary::PrimaryReceiverHandler` to the `primary::Helper`
     pub tx_helper_requests: IntGauge,
-    /// occupancy of the channel from the `primary::ConsensusAPIGrpc` to the `primary::BlockWaiter`
+    /// occupancy of the channel from the `primary::ConsensusAPIGrpc` & `executor::Subscriber` to the `primary::BlockWaiter`
```
huitseeker, line 72: I'd add one line saying that in contexts with an external consensus, it will be the former, and when the consensus is internal, it will be the later. This should make clear we don't actually expect 2 simultaneous senders yet.

huitseeker (approved):
Excellent work, thanks!

executor/src/subscriber.rs
huitseeker: https://github.com/MystenLabs/narwhal/pull/707 should also help with the crashing :)

PR #741, opened 2022-08-11
[crypto] Add Pedersen Commitments and Range Proofs

punwai:
This PR exposes new APIs for Pedersen Commitments and Bulletproofs range proofs. We expose the functionality to:

- Create Pedersen Commitment for some value and blinding factor
- Add and subtract Pedersen Commitments
- Create a Range Proof for some value and blinding factor
- Verify a Range Proof for some Pedersen commitment, given the range

The APIs will be exposed as native functions in Sui Move.

### DIFF
--- a/crypto/src/bulletproofs.rs
+++ b/crypto/src/bulletproofs.rs
@@ -1,28 +1,38 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
 use std::ops;
 
-use bulletproofs::{PedersenGens, BulletproofGens, RangeProof};
-use curve25519_dalek_ng::{scalar::Scalar, ristretto::{RistrettoPoint, CompressedRistretto}};
+use bulletproofs::{BulletproofGens, PedersenGens, RangeProof};
+use curve25519_dalek_ng::{
+    ristretto::{CompressedRistretto, RistrettoPoint},
+    scalar::Scalar,
+};
 use merlin::Transcript;
 use once_cell::sync::OnceCell;
-use serde::{Serialize, Deserialize, de};
+use serde::{de, Deserialize, Serialize};
 
 use crate::traits::ToFromBytes;
 
-// 
+//
 // Pedersen commitments
-// 
+//
+
+const PEDERSEN_COMMITMENT_LENGTH: usize = 32;
 
 #[derive(Debug, Clone)]
 pub struct PedersenCommitment {
     point: RistrettoPoint,
-    bytes: OnceCell<[u8; 32]>,
+    bytes: OnceCell<[u8; PEDERSEN_COMMITMENT_LENGTH]>,
 }
 
 impl PedersenCommitment {
     ///
     /// Creates a new Pedersen commitment from a value, and a blinding factor
     ///
-    pub fn new(value: [u8; 32], blinding_factor: [u8; 32]) -> Self {
+    pub fn new(
+        value: [u8; PEDERSEN_COMMITMENT_LENGTH],
+        blinding_factor: [u8; PEDERSEN_COMMITMENT_LENGTH],
+    ) -> Self {
         let generators = PedersenGens::default();
         let value = Scalar::from_bits(value);
         let blinding = Scalar::from_bits(blinding_factor);
@@ -59,20 +69,17 @@ impl ops::Sub<PedersenCommitment> for PedersenCommitment {
 
 impl AsRef<[u8]> for PedersenCommitment {
     fn as_ref(&self) -> &[u8] {
-        self.bytes.get_or_init(|| {
-            self.point.compress().to_bytes()
-        })
+        self.bytes.get_or_init(|| self.point.compress().to_bytes())
     }
 }
 
 impl ToFromBytes for PedersenCommitment {
     fn from_bytes(bytes: &[u8]) -> Result<Self, signature::Error> {
-        if bytes.len() != 32 {
+        if bytes.len() != PEDERSEN_COMMITMENT_LENGTH {
             return Err(signature::Error::new());
         }
         let point = CompressedRistretto::from_slice(bytes);
-        let decompressed_point = point.decompress()
-            .ok_or(signature::Error::new())?;
+        let decompressed_point = point.decompress().ok_or_else(signature::Error::new)?;
 
         Ok(PedersenCommitment {
             point: decompressed_point,
@@ -85,7 +92,7 @@ impl Serialize for PedersenCommitment {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
-    { 
+    {
         let bytes = self.as_ref();
         serializer.serialize_bytes(bytes)
     }
@@ -97,8 +104,7 @@ impl<'de> Deserialize<'de> for PedersenCommitment {
         D: de::Deserializer<'de>,
     {
         let bytes = Vec::deserialize(deserializer)?;
-        PedersenCommitment::from_bytes(&bytes[..])
-            .map_err(|e| de::Error::custom(e.to_string()))
+        PedersenCommitment::from_bytes(&bytes[..]).map_err(|e| de::Error::custom(e.to_string()))
     }
 }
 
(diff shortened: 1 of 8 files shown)


crypto/src/tests/secp256k1_tests.rs
```diff
@@ -258,11 +258,11 @@ proptest::proptest! {
 
         // same recovered pubkey are recovered
         let recovered_key = signature.sig.recover(&hashed_msg).unwrap();
-        let recovered_key_1 = signature_1.recover_verifying_key(message).expect("couldn't recover pubkey");
+        let recovered_key_1 = signature_1.recover_verify_key(message).expect("couldn't recover pubkey");
         assert_eq!(recovered_key.serialize(),recovered_key_1.to_bytes().as_slice());
 
-        // same signatures produced from both implementations
-        assert_eq!(signature.as_ref(), ToFromBytes::as_bytes(&signature_1));
+        // // same signatures produced from both implementations
```
punwai, line 264: Changes in versioning changes k256's signing implementation.

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,290 @@
...
+        let pc_gens = PedersenGens::default();
+        let bp_gens = BulletproofGens::new(64, 1);
+        let mut prover_transcript = Transcript::new(b"doctest example");
+
+        let blinding = Scalar::from_bits(blinding);
+        let bounding: u64 = u64::max_value();
+        let new_value = value + (bounding - upper_bound);
+
+        // Create a 64-bit rangeproof.
+        let (proof, _) = RangeProof::prove_single(
```
punwai, line 154: Every proof is created in 0 ... 64 range

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,263 @@
...
+
+impl AsRef<[u8]> for PedersenCommitment {
+    fn as_ref(&self) -> &[u8] {
+        self.bytes.get_or_init(|| self.point.compress().to_bytes())
+    }
+}
+
+impl ToFromBytes for PedersenCommitment {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, signature::Error> {
+        if bytes.len() != 32 {
```
akichidis, line 71: Could you please extract the number `32` to a const so we avoid [magic numbers](https://www.codebyamir.com/blog/software-anti-patterns-magic-numbers#:~:text=A%20magic%20number%20is%20a,so%20they%20should%20be%20avoided.) ?

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+impl BulletproofsRangeProof {
+    // Prove that commited value is within the range [0, 2^64)
+    pub fn prove_bound(
+        value: u64,
+        blinding: [u8; 32],
+    ) -> Result<(PedersenCommitment, Self), signature::Error> {
+        let pc_gens = PedersenGens::default();
+        let bp_gens = BulletproofGens::new(64, 1);
```
akichidis, line 141: Same comment as above, I would suggest extracting value `64` to a const

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,263 @@
...
+        let mut prover_transcript = Transcript::new(b"mysten");
```
akichidis, line 142: Not familiar with the `Transcript` but as far as I understood this is being used for domain separation. In this case do we want to go with `mysten` or rather the specific project (ie `narwhal`). I mean is this effectively allow the proofs to work across our products (ex assuming SUI is also using the same domain). We might (or not) want that. Apologies if my observation is totally off - not familiar with the  foundation.

crypto/src/tests/secp256k1_tests.rs
```diff
@@ -258,11 +258,11 @@ proptest::proptest! {
 
         // same recovered pubkey are recovered
         let recovered_key = signature.sig.recover(&hashed_msg).unwrap();
-        let recovered_key_1 = signature_1.recover_verifying_key(message).expect("couldn't recover pubkey");
+        let recovered_key_1 = signature_1.recover_verify_key(message).expect("couldn't recover pubkey");
```
akichidis, line 261: it seems that some tests fail because of the renaming

akichidis (approved):
Additions LGTM - left a few comments. I am approving but I would strongly suggest to wait for someone more experienced in cryptography to also approve before merging.

crypto/src/bulletproofs.rs
punwai: Yes, I believe this is domain separation - I'm happy to go with narwhal for this PR, but I think this entire crate is actually going to be moved to mysten-infra

crypto/src/tests/secp256k1_tests.rs
punwai: Yep, this is reverted!

crypto/src/bulletproofs.rs
kchalkias: Yes let's encode this as const ped commitment length

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(Debug)]
+pub struct BulletproofsRangeProof {
+    proof: RangeProof,
+    bytes: OnceCell<Vec<u8>>,
+}
+
+impl BulletproofsRangeProof {
+    /// Prove that commited value is within the range [0, 2^64). Function only works
+    /// for n = 8, 16, 32, 64
+    pub fn prove_bound(
```
kchalkias, line 137: can you elaborate what `n` is here? if it's the power of two then it's conflicting with the message that says within [0, 2^64) 
-> it should be [0, 2^n)

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,218 @@
...
+        value: u64,
+        blinding: [u8; 32],
+        n: usize,
+        domain: &'static [u8]
+    ) -> Result<(PedersenCommitment, Self), signature::Error> {
+        // Although this is also checked in the bulletproofs library, we check again
+        // to avoid unexpected behaviour in the case of library updates
+        if !(n == 8 || n == 16 || n == 32 || n == 64) {
+            return Err(signature::Error::new());
```
kchalkias, line 146: Better error msg? No worries just add a TODO and we can add in an upcoming PR for all errors in this commit

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,218 @@
...
+                bytes: OnceCell::new(),
+            },
+            BulletproofsRangeProof {
+                proof,
+                bytes: OnceCell::new(),
+            },
+        ))
+    }
+
+    // Verify that commited value is within the range [0, 2^64)
```
kchalkias, line 176: same: 2^n

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,218 @@
...
+    ) -> Result<(), signature::Error> {
+        if !(n == 8 || n == 16 || n == 32 || n == 64) {
+            return Err(signature::Error::new());
+        }
+
+        let pc_gens = PedersenGens::default();
+        let bp_gens = BulletproofGens::new(n, 1);
+        let mut verifier_transcript = Transcript::new(domain);
+
+        // Create a 64-bit rangeproof.
```
kchalkias, line 191: does it depend on n? or it's always 64?

crypto/src/bulletproofs.rs
punwai: That's my error - I didn't see this comment

crypto/src/bulletproofs.rs
punwai: Yes, this depends on n - I need to start reading the comments I write before submitting PRs. I've been ignoring them and looking at the code in the past.

crypto/src/bulletproofs.rs
punwai: Fixed all

crypto/src/bulletproofs.rs
```diff
@@ -0,0 +1,229 @@
+use std::ops;
```
kchalkias, line 1: add license header

kchalkias (approved):
Awesome, this is a good start towards privacy-related api, will approve!

PR #742, opened 2022-08-11
[fix] certificate waiter test

akichidis:
Removed `Ignore` too as this now works and should be executed as part of the unit tests.


huitseeker (approved):
🙏

PR #744, opened 2022-08-11
Input registry to restarter

asonnino:
The restarter should take the registry as input rather than re-creating a new one every time.


akichidis (approved):

huitseeker (approved):
This makes sense to me!

PR #746, opened 2022-08-11
[fix] block_waiter serve blocks with batches of similar digest ids

akichidis:
When created the `block_waiter` the assumption was that each produced batch will resolve to a unique digest id. Given the current design this is not guaranteed. If someone posts the same transactions _we could_ end up producing batches with the same batch id. That became more apparent from the tests on the [reconfigure.rs](https://github.com/MystenLabs/narwhal/blob/84db6c88f9a179ab3e145ecf81a86117cb522348/node/tests/reconfigure.rs#L158) while working on the task to swap the `batch_loader`  and use the `block_waiter` https://github.com/MystenLabs/narwhal/pull/738 . The tests are producing transactions of the exact same payload ending producing batches of same batch ids. Consequently the `block_waiter` was failing to successfully respond to concurrent requests of different certificates with common batch ids leading to receiving error messages like this:
```
Couldn't find pending batch with id ....
```

which is produced when trying to send a batch reply to the aggregator which waits for a certificate's batches to be received. Since `tx_pending_batch` was holding only the latest request's one shot sender, we couldn't basically notify multiple aggregators for the receipt of the batch.

This PR is fixing this. Also a small optimisation is performed to make sure that only `one network request` will be made to fetch a batch from a worker.

Merging this PR is essential in order to unblock the https://github.com/MystenLabs/narwhal/pull/738 as the [reconfigure.rs](https://github.com/MystenLabs/narwhal/blob/84db6c88f9a179ab3e145ecf81a86117cb522348/node/tests/reconfigure.rs#L158) will keep failing.

### DIFF
--- a/primary/src/block_waiter.rs
+++ b/primary/src/block_waiter.rs
@@ -14,7 +14,7 @@ use std::{
     fmt,
     fmt::Formatter,
     sync::Arc,
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::Duration,
 };
 use tokio::{
     sync::{oneshot, watch},
@@ -595,7 +595,7 @@ impl<SynchronizerHandler: Handler + Send + Sync + 'static> BlockWaiter<Synchroni
         debug!("No pending get block for {}", id);
 
         // Add on a vector the receivers
-        let batch_receivers = self.send_batch_requests(certificate.header.clone()).await;
+        let batch_receivers = self.send_batch_requests(id, certificate.header.clone()).await;
 
         let fut = Self::wait_for_all_batches(id, batch_receivers);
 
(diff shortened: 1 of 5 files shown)


asonnino (approved):

primary/src/block_waiter.rs
```diff
@@ -743,11 +767,12 @@ impl<SynchronizerHandler: Handler + Send + Sync + 'static> BlockWaiter<Synchroni
...
         match self.tx_pending_batch.remove(&batch_id) {
-            Some((sender, _)) => {
-                debug!("Sending BatchResult with id {}", &batch_id);
-                sender
-                    .send(result)
-                    .expect("Couldn't send BatchResult for pending batch");
+            Some(respond_to) => {
+                for (id, s) in respond_to {
+                    if let Err(err) = s.send(result.clone()) {
+                        error!("Couldn't send batch result {} message to channel [{:?}] for block_id {}", batch_id, err, id);
```
huitseeker, line 773: Nit: Here and in a few of the lines above you're using a lot of `if let Err(err) = xxx` and logging coupled.

Nothing wrong with that, but some may find it tedious to read. You might want to look at [tap_err](https://docs.rs/tap/latest/tap/tap/trait.TapFallible.html#tymethod.tap_err) and possibly use captured identifiers (`{id}`) to make this a bit lighter. YMMV.

huitseeker (approved):
Great fix, thanks a lot for the explicit comments, context, and tests, too!

primary/src/block_waiter.rs
akichidis: In this specific case I am only consuming the error just to print it. Using the `tap_err` in the end is making me (because of clippy) consume the result anyways and do a `let _ = ....` . Maybe the `tap_err` would be much better fit if I wanted to do something further with the result it self. That being said though, I like the proposal of introducing [tap](https://crates.io/crates/tap) as I can definitely see cases where we can use it - thanks for the recommendation! I've applied the recommendation.

PR #750, opened 2022-08-12
crypto: ensure same length of the aggregated sigs and pubkeys

joyqvq:
previously, ed25519-dalek's batch_verify takes in two arrays of signatures and pubkeys to verify. but for ed25519-consensus library, the user of the library needs to feed in each paired sig and pubkey. 

there is a bug introduced by https://github.com/MystenLabs/narwhal/pull/624 that the number of sigs is more than the number of pubkey provided, the for loop will only verify the first numbers of the sig and pubkey pairs, and terminate as OK and ignores the extra sig. 

example: 

```
Ed25519AggregateSignature::batch_verify(
        &[vec(sig1, sig2, sig3, sig4, bad_sig), vec(sig5, sig6, sig7, sig8, bad_sig)],
        &[vec(pk1, pk2, pk3, pk4), vec(pk5, pk6, pk7, pk8)],
        &[msg1, msg2]
    )
```

this is caught by a test case `test_handle_reject_malicious_signature` in sui that a malicious sig is added to the array to cause one extra sig provided, but not verified.

### DIFF
--- a/crypto/src/ed25519.rs
+++ b/crypto/src/ed25519.rs
@@ -341,6 +341,9 @@ impl AggregateAuthenticator for Ed25519AggregateSignature {
         let mut batch = batch::Verifier::new();
 
         for i in 0..pks.len() {
+            if pks[i].len() != sigs[i].0.len() {
+                return Err(signature::Error::new());
+            }
             for j in 0..pks[i].len() {
                 let vk_bytes = VerificationKeyBytes::from(pks[i][j].0);
                 batch.queue((vk_bytes, sigs[i].0[j], messages[i]));
--- a/crypto/src/tests/ed25519_tests.rs
+++ b/crypto/src/tests/ed25519_tests.rs
@@ -320,6 +320,29 @@ fn verify_batch_aggregate_signature_length_mismatch() {
     )
     .is_err());
 
+    // add an extra signature to both aggregated_signature that batch_verify takes in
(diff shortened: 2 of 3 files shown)


crypto/src/tests/ed25519_tests.rs
```diff
@@ -320,6 +320,29 @@ fn verify_batch_aggregate_signature_length_mismatch() {
     )
     .is_err());
 
+    // add an extra signature to both aggregated_signature that batch_verify takes in
```
joyqvq, line 323: to mirror the sui test where sigs len is one extra longer than the len of pubkeys

mwtian (approved):

crypto/src/ed25519.rs
```diff
@@ -341,6 +341,9 @@ impl AggregateAuthenticator for Ed25519AggregateSignature {
         let mut batch = batch::Verifier::new();
 
         for i in 0..pks.len() {
+            if pks[i].len() != sigs[i].0.len() {
```
mwtian, line 344: 1. Should we apply the same fix to Ed25519PublicKey:: verify_batch() at line 70?
2. Just curious, when `batch` is empty, does the verification succeed or fail?

kchalkias: Great points we should address both. Re 2nd: **empty batch should be considered a fail**; by experience we cannot accepting empty crypto materials as it could lead to surprises.

PR #756, opened 2022-08-12
Update Ed25519Signature ser/de logic

mwtian:
Implement struct ser/de logic for `Ed25519Signature`. The current ser/de logic of `Ed25519Signature` just returns a `&[u8]`, making it being treated as `[u8]` by `serde_reflection` and causing ser/de failures.

I investigated if we can rely on `serde` annotations and avoid manual implementation of struct ser/de. But it seems there is no straightforward way unless we want to compromise on
- Using cached `bytes` for ser/de.
- And / or using base64 for readable serialization.

### DIFF
--- a/crypto/src/ed25519.rs
+++ b/crypto/src/ed25519.rs
@@ -251,11 +251,8 @@ impl Serialize for Ed25519Signature {
     where
         S: serde::Serializer,
     {
-        if serializer.is_human_readable() {
-            base64ct::Base64::encode_string(self.as_ref()).serialize(serializer)
-        } else {
-            self.as_ref().serialize(serializer)
-        }
+        let str = base64ct::Base64::encode_string(self.as_ref());
+        serializer.serialize_newtype_struct("Ed25519Signature", &str)
     }
 }
 
@@ -264,14 +261,9 @@ impl<'de> Deserialize<'de> for Ed25519Signature {
     where
         D: serde::Deserializer<'de>,
     {
-        let bytes = if deserializer.is_human_readable() {
-            let s = String::deserialize(deserializer)?;
-            base64ct::Base64::decode_vec(&s).map_err(|e| de::Error::custom(e.to_string()))?
-        } else {
-            Vec::deserialize(deserializer)?
-        };
-        <Ed25519Signature as signature::Signature>::from_bytes(&bytes)
-            .map_err(|e| de::Error::custom(e.to_string()))
+        let s = String::deserialize(deserializer)?;
+        let value = Self::decode_base64(&s).map_err(|e| de::Error::custom(e.to_string()))?;
+        Ok(value)
     }
 }
 
--- a/crypto/src/ed25519.rs
+++ b/crypto/src/ed25519.rs
@@ -54,6 +54,7 @@ pub struct Ed25519KeyPair {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Ed25519Signature {
     pub sig: ed25519_consensus::Signature,
+    // Helps implementing AsRef<[u8]>.
     pub bytes: OnceCell<[u8; ED25519_SIGNATURE_LENGTH]>,
 }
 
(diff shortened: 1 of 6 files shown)


mwtian:
I figured out the reason of `trace_value()` failures: calling `serialize_newtype_struct()` makes `serde-reflection` think the type is just an alias, so it seems to not remember the type and value. Using `serialize_struct()` fixes the issue but it is a bit more complex. An alternative is to use the pattern we have for `BLS12381Signature`. Will update the PR later.

crypto/src/ed25519.rs
```diff
@@ -264,14 +278,73 @@ impl<'de> Deserialize<'de> for Ed25519Signature {
     where
         D: serde::Deserializer<'de>,
     {
-        let bytes = if deserializer.is_human_readable() {
-            let s = String::deserialize(deserializer)?;
-            base64ct::Base64::decode_vec(&s).map_err(|e| de::Error::custom(e.to_string()))?
+        struct Ed25519SignatureVisitor {
```
mwtian, line 281: If we just put `#[derive(Deserialize, Serialize)]` on top of `Ed25519Signature` and `#[serde(skip)]` on top of `bytes`, we can save the custom ser/de logic here. The downsides are:
1. Human readable data is not base64, but something less compact.
2. We will not be using the `bytes` once_cell "cache".

mwtian (commented):
Should be ready now.

crypto/src/tests/ed25519_tests.rs
```diff
@@ -51,6 +52,38 @@ fn serialize_deserialize() {
     assert!(bincode::deserialize::<Ed25519PublicKey>(&bytes).is_err());
 }
 
+#[test]
+fn custom_serde_reflection() {
```
huitseeker, line 56: 👍

node/src/generate_format.rs
```diff
@@ -126,18 +126,6 @@ fn get_registry() -> Result<Registry> {
     tracer.trace_type::<HeaderDigest>(&samples)?;
     tracer.trace_type::<CertificateDigest>(&samples)?;
 
-    // Caveat: the following (trace_type) won't work, but not because of generics.
```
huitseeker, line 129: Indeed, this is quite an antiquated comment.

crypto/src/ed25519.rs
```diff
@@ -3,7 +3,12 @@
 use base64ct::{Base64, Encoding};
 use ed25519_consensus::{batch, VerificationKeyBytes};
 use once_cell::sync::OnceCell;
-use serde::{de, Deserialize, Serialize};
+use serde::{
+    de::{self, MapAccess, SeqAccess, Visitor},
+    ser::SerializeStruct,
+    Deserialize, Serialize,
+};
+use serde_bytes::{ByteBuf, Bytes};
```
huitseeker, line 11: Nit (that you can ignore): we already use `serde_with` elsewhere, and it has analogous byte types. But that may be a bit too much OCD on my part.
https://docs.rs/serde_with/latest/serde_with/

huitseeker (commented):
This is excellent! 

I think that the only thing that is missing is a few lines of english in a comment of `impl (De)Serialize` summarizing what invariants the programmer should maintain when editing the inner representation of our types:
- we use this disjunction between human-readable and non-human readable formats since #460,
- at the same time we aim to have serde_reflection processable types, which works poorly [with aliases and nameless types](https://docs.rs/serde-reflection/latest/serde_reflection/index.html#unsupported-idioms).

Here's why I think a comment about this is necessary: you're making specific choices here (e.g. the marker fields for disjunction between b64 and raw formats) that we may debate in the future. If someone was to make different ones, what shouldn't they break?

crypto/src/ed25519.rs
mwtian: Thanks for the suggestion. `serde_with::Bytes` seems to be missing some features compared to `serde_bytes::Bytes`, e.g. wrapping `&[u8]`. It is a bit hard to use it in the situation here.

crypto/src/ed25519.rs
mwtian: > This is excellent!
> 
> I think that the only thing that is missing is a few lines of english in a comment of `impl (De)Serialize` summarizing what invariants the programmer should maintain when editing the inner representation of our types:
> 
> * we use this disjunction between human-readable and non-human readable formats since #460,
> * at the same time we aim to have serde_reflection processable types, which works poorly [with aliases and nameless types](https://docs.rs/serde-reflection/latest/serde_reflection/index.html#unsupported-idioms).
> 
> Here's why I think a comment about this is necessary: you're making specific choices here (e.g. the marker fields for disjunction between b64 and raw formats) that we may debate in the future. If someone was to make different ones, what shouldn't they break?

Thanks for the feedback. Added a comment with the information outlined above.

joyqvq:
lgtm! should we add similar tracer test to other serializer for Secp256k1 and BLS signatures (in a follow up PR)? or, port over the format tests from sui to narwhal as well?

joyqvq (approved):

mwtian:
> lgtm! should we add similar tracer test to other serializer for Secp256k1 and BLS signatures (in a follow up PR)? or, port over the format tests from sui to narwhal as well?

Good idea, Narwhal does have a similar `generate_format.rs` test and I'm thinking of adding tests for structs containing signatures to it.

huitseeker (approved):
Thanks @mwtian !

PR #760, opened 2022-08-13
[crypto] systematize batch verification testing

punwai:
Systematize tests on batch_verify for aggregated signatures.


huitseeker (approved):
Excellent! Thanks a lot!

PR #762, opened 2022-08-14
fix: Make leader election select from the whole Committee

### LINKED ISSUE #755: [consensus] Rotate through all the leader roster
The leader selection process is meant to deterministically chose a leader, for example in a round robin way or even pseudorandom keyed by round.
The function is here [1] [2] :
```rust
    /// Returns a leader node in a round-robin fashion.
    pub fn leader(&self, seed: usize) -> PublicKey {
        let mut keys: Vec<_> = self.authorities.keys().cloned().collect();
        keys.sort();
        keys[seed % self.size()].clone()
```

- Problem 1: this does not select on the basis of stake, but only number, which we should change.
- Problem 2: more subtle. The round (seed) we pass in is filtered by a r % 2 != 0 condition. Meaning that I think we will be selecting a subset of only odd or even validators.
[1]: https://github.com/MystenLabs/narwhal/blob/9c1862af04e8c5ecf9c7504900e9d5d00a088859/consensus/src/bullshark.rs#L145-L161
[2]: https://github.com/MystenLabs/narwhal/blob/9c1862af04e8c5ecf9c7504900e9d5d00a088859/consensus/src/tusk.rs#L145-L164

huitseeker:
Fixes #755

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -11,6 +11,7 @@
 use arc_swap::ArcSwap;
 use crypto::{traits::EncodeDecodeBase64, PublicKey};
 use multiaddr::Multiaddr;
+use rand::{rngs::StdRng, seq::SliceRandom, SeedableRng};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
@@ -377,11 +378,21 @@ impl Committee {
         (total_votes + 2) / 3
     }
 
-    /// Returns a leader node in a round-robin fashion.
+    /// Returns a leader node as a weighted choice seeded by the provided integer
     pub fn leader(&self, seed: usize) -> PublicKey {
-        let mut keys: Vec<_> = self.authorities.keys().cloned().collect();
-        keys.sort();
-        keys[seed % self.size()].clone()
+        let mut seed_bytes = [0u8; 32];
+        seed_bytes[32 - 8..].copy_from_slice(&seed.to_le_bytes());
+        let mut rng = StdRng::from_seed(seed_bytes);
+        let choices = self
+            .authorities
+            .iter()
+            .map(|(name, authority)| (name, authority.stake as f32))
+            .collect::<Vec<_>>();
+        choices
+            .choose_weighted(&mut rng, |item| item.1)
+            .expect("Weighted choice error: stake values incorrect!")
+            .0
+            .clone()
     }
 
     /// Returns the primary addresses of the target primary.
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -379,7 +379,7 @@ impl Committee {
     }
 
     /// Returns a leader node as a weighted choice seeded by the provided integer
-    pub fn leader(&self, seed: usize) -> PublicKey {
+    pub fn leader(&self, seed: u64) -> PublicKey {
         let mut seed_bytes = [0u8; 32];
         seed_bytes[32 - 8..].copy_from_slice(&seed.to_le_bytes());
         let mut rng = StdRng::from_seed(seed_bytes);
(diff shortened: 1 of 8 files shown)


config/tests/config_tests.rs
```diff
@@ -29,6 +29,24 @@ use std::{fs::File, io::Write};
...
 
+#[test]
+fn leader_election_rotates_through_all() {
+    // this committee has equi-sized stakes
+    let committee = test_utils::committee(None);
+    let mut leader_counts = HashMap::new();
+    let mut leader_counts_stepping_by_2 = HashMap::new();
+    for i in 0..100 {
+        let leader = committee.leader(i);
+        let leader_stepping_by_2 = committee.leader(i * 2);
```
akichidis, line 40: It's not immediately clear to me what we are trying to test by having the `leader_stepping_by_2` here. @huitseeker could you please provide some info?

config/tests/config_tests.rs
```diff
@@ -29,6 +29,24 @@ use std::{fs::File, io::Write};
...
+        *leader_counts.entry(leader).or_insert(0) += 1;
+        *leader_counts_stepping_by_2
+            .entry(leader_stepping_by_2)
+            .or_insert(0) += 1;
+    }
+    assert!(leader_counts.values().all(|v| *v > 1));
```
akichidis, line 46: Since we are weighting based on stake would it make sense to also check the diff between the number of times the leaders have been elected and ensure they don't diverge much? (since we use normal distribution).

akichidis (approved):
Great fix @huitseeker , changes make sense to me (from a non-consensus expert point of view compared to the other peer reviewers).

config/tests/config_tests.rs
huitseeker: It would be an improvement, but I don't see as a main . 
We use a uniform distribution for the `test_utils::committee` function, which means we are doing Bernouilli sampling. 

If we want to be as general as we can, this test should generate a binomial distribution, something we can test [with a goodness-of-fit test adapted for the discrete case](http://www.stat.yale.edu/~jay/EmersonMaterials/DiscreteGOF.pdf). We can also test exact numbers (the test is deterministic), but this may break the second we get out of our current discipline of iterating through seeds in a particular way. A good trade-off would be to use a simple inequality (e.g. Chebysheff).

I'll open an issue & follow-up.

config/tests/config_tests.rs
huitseeker: I've added a comment: we iterate through even rounds in practice, and so I wanted to make sure we hit the whole roster (each available validator) reliably.

config/src/lib.rs
```diff
@@ -377,11 +378,21 @@ impl Committee {
...
     }
 
-    /// Returns a leader node in a round-robin fashion.
+    /// Returns a leader node as a weighted choice seeded by the provided integer
     pub fn leader(&self, seed: usize) -> PublicKey {
-        let mut keys: Vec<_> = self.authorities.keys().cloned().collect();
-        keys.sort();
-        keys[seed % self.size()].clone()
+        let mut seed_bytes = [0u8; 32];
+        seed_bytes[32 - 8..].copy_from_slice(&seed.to_le_bytes());
```
asonnino, line 384: why `32-8`?

asonnino (approved):

config/src/lib.rs
huitseeker: The seed is a u64 (8 bytes)
Ah, yes, you're right.We should include the epoch in there.

asonnino: got it

config/tests/config_tests.rs
akichidis: Thanks for the detailed reply on this! Great to see that you did with solution (2). Definitely (3) is something we could apply here - and now that I can think off we could use in other places in the end to end tests .

PR #763, opened 2022-08-14
fix: Avoid spammy logging on hitting the network's BoundedExecutor

### LINKED ISSUE #759: Logs fill up with 'available tasks is 0' upon network saturation
### Description
```
2022-08-12T20:50:17.765387Z  WARN network::primary: Executor in network:primary and module:primary_core available tasks is 0 for client address: /dns/validator-3/tcp/8381/http
```

The log files fill up and are spammed with the above.

### Explanation from @huitseeker : 

It means that the network has reached its maximum amount of concurrent network messages. Further messages are queued and will be sent when further semaphore tickets are available.

A network sending attempt can move off the semaphore in 3 ways :
* the message send finishes,
* the message send fails (some will be rescheduled on an exponential backoff)
* the message is canceled because the network has progressed past the point where this would be relevant. (that cancellation cancels the retries for those nodes that are retried)

### Steps to reproduce
Run a docker-compose with multiple Narwhal nodes.  Saturate the network such that the network queue fills up.

### Possible solutions:

Have the network throttling code not spam warnings every time it is full -- maybe set a flag and warn just once out of N invocations?

huitseeker:
## Context
We operate an executor with a bound on the concurrent number of messages (see #463, #559, #706).
PR #472 added logging when the bound is hit.

## The issue
We expect the executors to operate for a long time at this limit (e.g. in recovery situation).
The spammy logging is not usfeful

## The fix
This removes the logging of the concurrency bound being hit.
Fixes #759

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -57,4 +57,4 @@ impl<T> std::future::Future for CancelOnDropHandler<T> {
 //
 // The exact number here probably isn't important, the key things is that it should be finite so
 // that we don't create unbounded numbers of tasks.
-pub const MAX_TASK_CONCURRENCY: usize = 200;
+pub const MAX_TASK_CONCURRENCY: usize = 500;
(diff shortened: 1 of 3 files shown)


akichidis (approved):
Thanks for reducing logging 👍 !

velvia (commented):
Thank you @huitseeker !

PR #800, opened 2022-08-16
crypto: fix Ed25519AggregateSignature serialization

joyqvq:


crypto/src/serde_helpers.rs
```diff
@@ -67,3 +67,39 @@ pub fn keypair_decode_base64<T: KeyPair>(value: &str) -> Result<T, eyre::Report>
     }
     Ok(kp)
 }
+
+pub struct Ed25519Signature;
+
+impl SerializeAs<ed25519_consensus::Signature> for Ed25519Signature {
```
mwtian, line 73: Hopefully in future we can reuse this logic across signatures.

mwtian (approved):

PR #810, opened 2022-08-18
crypto: use fastcrypto library

joyqvq:


crypto/src/bls12381.rs
```diff
@@ -401,7 +401,7 @@ impl KeyPair for BLS12381KeyPair {
     type PrivKey = BLS12381PrivateKey;
     type Sig = BLS12381Signature;
 
-    #[cfg(feature = "copy_key")]
+    #[cfg(any(test, feature = "copy_key"))]
```
joyqvq, line 404: temporary workaround - not needed as /crypto folder will be removed

asonnino (approved):
So we basically move the crypto crate to its own repo (and then import it in Narwhal and Sui)?

joyqvq:
> So we basically move the crypto crate to its own repo (and then import it in Narwhal and Sui)?

yup! the goal is to decouple the crypto crate without narwhal, it can be used by narwhal, sui or external parties to upgrade the pointer or release version independently

PR #811, opened 2022-08-18
[test] showcase node shutdown issue

akichidis:
I believe that the use of `TaskGroup` in MystenLabs/narwhal#707 doesn't allow the node to shutdown properly. This test here showcase that. We should expect this test to pass, but instead it fails because we are still getting a successful response from the node even after stopping. This can also explain the issue in https://github.com/MystenLabs/sui/issues/5323 - because the nodes don't shutdown when we try to start them again they hold already a lock and thus we fail.

Please see https://github.com/MystenLabs/narwhal/pull/812 which proves that reverting the MystenLabs/narwhal#707 fixes the problem. I timeboxed some quick investigation to identify the issue with the `TaskGroup` but didn't manage to find the root cause (_I suspect_ it has to do with the fact that we use the `task_group.spawn` and we pass in the handlers instead of using directly on the tasks) , so I propose the revert MystenLabs/narwhal#812 to get unblocked in the integration tests (I'll need them for my upcoming PR) until we fix it and re-apply , unless it's something quick/easy to fix.

### DIFF
--- a/primary/tests/nodes_bootstrapping_tests.rs
+++ b/primary/tests/nodes_bootstrapping_tests.rs
@@ -4,6 +4,43 @@ use std::time::Duration;
 use telemetry_subscribers::TelemetryGuards;
 use test_utils::cluster::Cluster;
 
+use types::{PublicKeyProto, RoundsRequest};
+
+#[tokio::test]
+async fn test_shutdown_bug() {
+    // Enabled debug tracing so we can easily observe the
+    // nodes logs.
+    let _guard = setup_tracing();
+
+    let delay = Duration::from_secs(10); // 10 seconds
+
+    // A cluster of 4 nodes will be created
+    let cluster = Cluster::new(None, None, false);
+
+    // ==== Start first authority ====
+    let authority = cluster.authority(0);
+    authority.start(false, Some(1)).await;
+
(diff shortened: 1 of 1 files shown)


primary/tests/nodes_bootstrapping_tests.rs
```diff
@@ -4,6 +4,43 @@ use std::time::Duration;
...
+    // send a sample rounds request
+    let request = tonic::Request::new(RoundsRequest {
+        public_key: Some(PublicKeyProto::from(authority.name.clone())),
+    });
+    let response = client.rounds(request).await;
+
+    // Should get back an error response - however this test will fail
+    // as we keep getting an OK response back , which shouldn't happen , as we
+    // stopped the node.
+    assert!(response.is_err());
```
akichidis, line 41: This asset here will fail, although we would expect it to succeed under normal circumstances as we did shutdown the node

huitseeker:
This should be superseded by MystenLabs/narwhal#812, no?

akichidis:
> This should be superseded by MystenLabs/narwhal#812, no?

Exactly, as [this](https://github.com/MystenLabs/narwhal/pull/812) has been merged this is now not needed. Will close.

PR #812, opened 2022-08-18
[test] revert task group change

akichidis:
Follow up of https://github.com/MystenLabs/narwhal/pull/811 . When we revert the changes we can see that now the `test_shutdown_bug` successfully passes as we would expect.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2356,7 +2356,6 @@ dependencies = [
  "serde-reflection",
  "serde_yaml",
  "structopt",
- "task-group",
  "telemetry-subscribers",
  "test_utils",
  "thiserror",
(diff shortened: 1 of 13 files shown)


huitseeker (approved):
Thanks @akichidis!
Yes, as #811 and #765 demonstrate, it makes sense to revert this for now. Your thinking in #811 seems correct: Our `{Task1, Task2}::spawn` calls should take a reference to a controlling taskgroup and spawn on it, rather than having the taskgroup manage the collection of `JoinHandle`s from those tasks. Indeed, because the `TaskManager` will keep polling those `JoinHandles`s, the tasks will never drop by themselves, which leads to the zombie node problem you outline.

mwtian:
I see. Probably there is a misunderstanding of the interaction between join handle and tokio task. Will investigate later.

Issue #813, opened 2022-08-18
[refactor] certificate storage to keep a secondary index to use across consensus & core for recovery


akichidis:
### Description
TBD

huitseeker:
Fixed by #825

PR #814, opened 2022-08-18
[recovery] core to recover the last certificates and forward to proposer

akichidis:
Part of MystenLabs/sui#5324 

### Problem

Upon node bootstrap we want to recover the last received certificates and forward them to the proposer so it can have the opportunity to propose a new header and make progress.

### Solution
Read from the `certificate_store` the certificates that are after the `last_commit_round` and figure out the ones with the highest round number.

The current solution is blocking - meaning that core will not initiate its loop until it has finished with recovery. Thought about various approaches but decided to go with that for now. Since our networking stack gets initialised before that it's possible that our node might start receiving certificates, headers etc while we perform the recovery fill up our buffers if our recovery is slow. In the TODO list there is proposals to speed up the recovery.

Also thought about the recovery process in general as:
* making it blocking Vs non-blocking. In some cases makes sense to block and wait for a component to finish its recovery before proceeding
* dependencies between components while we recover
* networking conditions

☝️ points for further discussion

---
TODO:

- [ ] In order for core to successfully recover we need the `last_commit_round` which should be driven/emitted by consensus. To achieve that we first need to make sure that consensus has recovered. We need to make consensus block until we have figured out the last commit round to propagate downstream
- [ ] currently consensus & core are following similar logic in recovery - filtering out the certificate_store to figure out the latest certificates. Building a secondary index https://github.com/MystenLabs/narwhal/issues/813 where certificates are stored by their rounds would help us quickly skip certificates and quickly perform the recovery. The storage can be re-used by both components (and probably more)

### DIFF
--- a/primary/src/core.rs
+++ b/primary/src/core.rs
@@ -17,6 +17,7 @@ use std::{
     time::Instant,
 };
 use store::Store;
+use tap::Tap;
 use tokio::{sync::watch, task::JoinHandle};
 use tracing::{debug, error, info, instrument, warn};
 use types::{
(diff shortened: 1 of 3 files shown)


primary/src/core.rs
```diff
@@ -132,11 +133,82 @@ impl Core {
                 cancel_handlers: HashMap::with_capacity(2 * gc_depth as usize),
                 metrics,
             }
+            .recover()
+            .await
```
asonnino, line 137: I like this programming pattern

primary/src/core.rs
```diff
@@ -132,11 +133,82 @@ impl Core {
...
+            .iter()
+            .filter_map(|(_, certificate)| {
+                if certificate.round() == last_round {
+                    Some(certificate)
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<&Certificate>>()
+            .tap(|c| info!("Number of certificates to repopulate: {}", c.len()));
```
asonnino, line 202: What does it do?

asonnino (approved):

huitseeker:
@akichidis following our offline chat, should we close in favor of MystenLabs/narwhal#817?

akichidis:
> @akichidis following our offline chat, should we close in favor of MystenLabs/narwhal#817?

Yes, I am doing so now as part of this comment

PR #815, opened 2022-08-18
Add ExecutionState::handle_consensus_without_transactions

### LINKED ISSUE #758: Pass empty batches/transactions to ExecutorState
Currently the narwhal executor core simply skips empty batches. It seems however better to pass them forward to the executor state (who can make informed decision based on those data).

We should modify the ExecutionState interface in some way to let us know about the beginning end of batches. This would be similar to the [Tendermint ABCI](https://docs.tendermint.com/master/spec/abci/abci.html#beginblock).

### LINKED ISSUE #5322: Pass empty batches/transactions to ExecutorState
Currently the narwhal executor core simply skips empty batches. It seems however better to pass them forward to the executor state (who can make informed decision based on those data).

We should modify the ExecutionState interface in some way to let us know about the beginning end of batches. This would be similar to the [Tendermint ABCI](https://docs.tendermint.com/master/spec/abci/abci.html#beginblock).

aakoshh:
# Changes
This PR introduces an extra method, `ExecutionState::handle_consensus_without_transactions` that informs the executor about certificates that contain no transactions. 

## Motivation
The motivation to do so is because we want to use Narwhal as the transaction ordering layer to replace the consensus in an existing blockchain, by creating a deterministic chain of blocks from the certificates that Narwhal and Bullshark produce. In that blockchain, there might be a need for blocks to be produced even when there are no transactions (same as in Narwhal), and to make it deterministic, the only source of truth we can rely on that indicates the passage of time across all nodes are the certificates themselves. 
`
## Backwards Compatibility
The change is backwards compatible because the method has a default implementation that does nothing, so it doesn't break Sui in any way. 

This means, however, that the code doesn't put any expectations on what the executor does with the information, namely it doesn't pass `ExecutionIndices` for atomic persistence. This means that in theory these certificates could be passed again after a crash, and the executor has to be able to deduplicate them based on the `consensus_index` in the `ConsensusOutput`. 

## Alternative Designs

The first idea I had was to introduce something like the [BeginBlock / EndBlock](https://docs.tendermint.com/master/spec/abci/abci.html#beginblock) pair in Tendermint's ABCI. We'd have to be careful after a restart not to call _begin_ again, because Narwhal expects that the `ExecutionIndices` passed to `handle_consensus_transaction` are persisted, so like transactions, this event should be skipped. But then I realised that because `ConsensusOutput` and `ExecutionIndices` are passed to `handle_consensus_transaction` every time, it is easy for the executor to detect when a certificate starts or ends. Having _begin_ and _end_ would only make sense if this wasn't the case, to put the responsibility of signaling and persisting indices squarely on those methods, but that would be a breaking change for no obvious benefit, and it would open up different questions about where to replay from.

Another option would be to pass all transactions in the certificate as a single call, which is how the [FinalizeBlock](https://docs.tendermint.com/master/spec/abci++/abci++_methods.html#finalizeblock) method works in Tendermint's ABCI++. That would put the onus of partial batch execution on the executor itself. It would also be a breaking change for Sui. 

## Open Questions

I see in the docstring of `handle_consensus_transaction` that it should be able to return committee changes, but I couldn't find anything about this happening in the current codebase, the `Output` type seems completely opaque. 

If it was possible though, then `handle_consensus_without_transactions` should also be able to return such a result. An example of where this could be useful is to define epochs as a fixed number of blocks; transactions to register new validators and delegations should be allowed to go through during the epoch, but they should only take effect at the end of the epoch, which should not require further transactions to take place. In this case the committee change can be triggered by empty blocks as well as non-empty ones. In Tendermint ABCI, this happens in `EndBlock`. 

With the current `ExecutionState` I would have to detect the last transaction of the certificate, and do extra execution there, then "attach" the result to the `Outcome` of the transaction. That's not exactly the right place to do it, though, since it might not have been that transaction which the effects belong to. 

Currently the `handle_consensus_without_transactions` returns `Result<(), Self::Error>`. I didn't think it would be right to return `Outcome` from it because: 
1. There is no default value for the default method implementation. I could return `Option<Outcome>` from it with `None` begin the default.
2. But it seems tied to the actual transactions, being returned as pairs and passed on to an observer. If I returned `Some(outcome)` I wouldn't have a transaction to pair it with for the observer. 

For these reasons I opted not to return anything, until it becomes clear how committee changes should be carried out.

The situation would be easier if the whole certificate was passed on to the executor, and the return value could optionally contain the new committee. Perhaps the one-by-one executor could be just a special case implementation of it.

EDIT: This alternative with the one-by-one being a special case is available in https://github.com/MystenLabs/narwhal/pull/818

# Related Issues
Resolves MystenLabs/sui#5322

### DIFF
--- a/executor/src/core.rs
+++ b/executor/src/core.rs
@@ -110,21 +110,19 @@ where
 
     /// Execute a single certificate.
     async fn execute_certificate(&mut self, message: &ConsensusOutput) -> SubscriberResult<()> {
-        // Skip the certificate if it contains no transactions.
         if message.certificate.header.payload.is_empty() {
-            self.execution_indices.skip_certificate();
-            return Ok(());
-        }
-
-        // Execute every batch in the certificate.
-        let total_batches = message.certificate.header.payload.len();
-        for (index, digest) in message.certificate.header.payload.keys().enumerate() {
-            // Skip batches that we already executed (after crash-recovery).
-            if self
-                .execution_indices
-                .check_next_batch_index(index as SequenceNumber)
-            {
-                self.execute_batch(message, *digest, total_batches).await?;
+            self.execute_empty_certificate(message).await?;
+        } else {
+            // Execute every batch in the certificate.
+            let total_batches = message.certificate.header.payload.len();
+            for (index, digest) in message.certificate.header.payload.keys().enumerate() {
+                // Skip batches that we already executed (after crash-recovery).
+                if self
+                    .execution_indices
+                    .check_next_batch_index(index as SequenceNumber)
+                {
+                    self.execute_batch(message, *digest, total_batches).await?;
+                }
             }
         }
         Ok(())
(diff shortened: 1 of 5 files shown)


asonnino:
Do we still need this PR since we have this https://github.com/MystenLabs/narwhal/pull/818 ?

aakoshh:
No, if you agree that one's more forward looking I would just close this one.

PR #816, opened 2022-08-18
[refactor] replace the certificate_store with a wrapper with secondary index

akichidis:


node/src/certificate_store.rs
```diff
@@ -0,0 +1,240 @@
...
+    certificates_by_id: DBMap<CertificateDigest, Certificate>,
+    /// A secondary index that keeps the certificate digest ids
+    /// by the certificate rounds. That helps us to perform
+    /// range requests based on rounds. We avoid storing again the
+    /// certificate here to not waste space. To dereference we use
+    /// the certificates_by_id storage.
+    certificate_ids_by_round: DBMap<(Round, CertificateDigest), u8>,
+    /// Senders to notify for a write that happened for
+    /// the specified certificate digest id
+    obligations: Arc<RwLock<HashMap<CertificateDigest, VecDeque<oneshot::Sender<Certificate>>>>>,
```
akichidis, line 29: I could instead use a [DashMap](https://crates.io/crates/dashmap) which you have also used in the dag implementation

akichidis: My intention with having interior mutability here is to be able to share/clone the `CertificateStore` across structs - the "penalty" is having a lock around the obligations I guess, but it might not be that much of an issue

node/src/certificate_store.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// are returned in the same sequence as the provided keys.
+    pub fn multi_read(
+        &self,
+        ids: impl IntoIterator<Item = CertificateDigest>,
+    ) -> StoreResult<Vec<Option<Certificate>>> {
+        self.certificates_by_id.multi_get(ids)
+    }
+
+    /// Waits to get notified until the requested certificate becomes available
+    pub async fn notify_read(&self, id: CertificateDigest) -> StoreResult<Option<Certificate>> {
```
akichidis, line 96: To completely ditch the `certificate_store` we need the pub/sub system the `Store` offers. Here I preferred I followed a different approach than Store - meaning not use a tokio task to handle the commands. Let me know yourr thoughts @huitseeker

PR #818, opened 2022-08-18
Batch execution with single execution adapter

### LINKED ISSUE #5322: Pass empty batches/transactions to ExecutorState

aakoshh:
This PR is a more aggressive alternative for https://github.com/MystenLabs/narwhal/pull/815

The basic motivation is the same: let the executor know about every certificate, whether it's empty or not. But instead of having a different method for empty certificates, this changes the `executor::Core` to expect a new `BatchExecutionState` which takes the whole certificate with all the batches it includes. It takes `SerializedTransaction` so it can decide how to deal with deserialization errors.

The trait also has a `load_next_certificate_index` method to support replay (which we discussed @asonnino is currently missing). The replay is expected to happen from the index it returns, and it's the responsibility of the state to skip any transactions it has already processed, if the whole certificate wasn't processed atomically.

For backwards compatibility, the the PR also has a `SingleExecutionState` that has the same method that takes singular transactions, and a `SingleExecutor` adapter which maintains the `ExecutionIndices` for it. This way you don't have to change Sui too much, only in the application root where the execution state is instantiated. 

The `SingleExecutor` also takes over the responsiblity of feeding the output channel, that is, it feeds the pairs of `(outcome, transaction_bytes)` to any observers. This didn't seem to make sense for the batch execution state, which could have extra outcome without transactions. Arguably the execution state itself can forward its own output to where it has to go, so threading the `tx_confirmation` or `tx_output` through the whole system seems to serve vestigial purposes; and it's also easy to do using some adapter that inspects the return value. 

There is a slight discrepancy between the `BatchExecutionState` and the `SingleExecutionState` in that the latter takes a deserialized `Transaction` type, rather than bytes. The `SingleExecutor` preserves the current behaviour of reporting outcome even on malformed transactions, and only passing valid ones to the execution state. It also assumes `bincode` is used.

Overall I think this is a cleaner solution than the previous PR, and it makes it easier to expand the solution to patterns I mentioned there; for example to enact committee changes at the end of (potentially empty) batches based on how many certificates we have seen, rather than upon individual transactions.

Resolves https://github.com/MystenLabs/sui/issues/5322


executor/src/lib.rs
```diff
@@ -50,6 +50,41 @@ pub trait ExecutionState {
...
+    /// Load the last consensus index from storage.
+    ///
+    /// It should return the index from which it expects a replay, so one higher than
+    /// the last certificate index it successfully committed. This is so it has the
+    /// same semantics as `ExecutionIndices`.
+    async fn load_next_certificate_index(&self) -> Result<SequenceNumber, Self::Error>;
+
+    /// Execute the transactions and atomically persist the consensus index.
+    ///
+    /// TODO: This function should be allowed to return a new committee to reconfigure the system.
```
asonnino, line 77: Reconfiguration can simply be done by sending the following message to the Primary and the Workers through the network: `ReconfigureNotification::NewCommittee(new_committee)`.

Below is an example:
https://github.com/MystenLabs/narwhal/blob/ad05db4c5d6f851402efc7d2b285e9b59d513e64/node/src/restarter.rs#L98-L101

executor/src/core.rs
```diff
@@ -143,54 +147,175 @@ where
...
+            .into_iter()
+            .map(
+                |serialized| match bincode::deserialize::<State::Transaction>(&serialized) {
+                    Ok(x) => Ok(x),
+                    Err(e) => Err(SubscriberError::ClientExecutionError(format!(
+                        "Failed to deserialize transaction: {e}"
+                    ))),
+                },
+            )
+            .collect::<Result<Vec<_>, _>>()?;
```
asonnino, line 168: Could we collect a Vec<Result> rather than a Result<Vec> so that we can skip faulty transactions rather than the whole batch?

executor/src/core.rs
```diff
@@ -143,54 +147,175 @@ where
...
+
+/// Executor that feeds transactions one by one to the execution state.
+pub struct SingleExecutor<State>
+where
+    State: SingleExecutionState,
+{
+    /// The (global) state to perform execution.
+    execution_state: Arc<State>,
+    /// The indices ensuring we do not execute twice the same transaction.
+    execution_indices: Mutex<ExecutionIndices>,
```
asonnino, line 182: This Mutex seems dangerous. Do we absolutely need it?

executor/src/lib.rs
aakoshh: Thanks for the link! True, we discussed that we can send reconfiguration messages through gRPC. 

I included this TODO because the [current docs](https://github.com/MystenLabs/narwhal/blob/52dd0c4efa36ae303b5ef80e015c108a2fb6955d/executor/src/lib.rs#L57-L58) of the `handle_consensus_transaction` say that it can return a new committee. While that's not true, as it clearly cannot, I thought the idea that it should do so was a good one. It capture committee transitions on the level of the `ExecutionState`, if it is to happen as part of transaction execution. The gRPC way seems like a technical detail that the implementors of the `ExecutionState` should not have to concern themselves with.

But it's up to you, I just wanted it to be consistent with the other docs.

executor/src/core.rs
aakoshh: I thought about this while debugging one of the unit tests that failed to deserialise (the compiler decided to send `i32` instead of `u64`). I think this is a bit of a grey area. I see that you label the error as `ClientExecutionError` which, during execution, is treated as a non-fatal one. However, it's not clear whether treating deseralisation errors as non-fatal would be correct. 

That's because it's difficult to say why we can't deserialise a transaction: Is it because some malicious validator put a malformed message into one of the batches? Or is it because the `Transaction` type on our side is somehow incorrect, and our machine is the only one that fails to deserialize something? Perhaps other validators have added an extra variant to an enum, but we forgot to update our node? Skipping such transactions could lead to consensus failure down the road. 

What we should do depends on how much pre-validation happens on the contents _before_ quorum is reached over their availability. Can malicious validators use Narwhal to atomically broadcast absolute rubbish content? Or is there some vetting before we vote on it to see if it at least conforms to some basic expectations about format?

If there is such validation, then we could be sure that the honest majority thinks this transaction looks legit, so if we can't deserialize it, we should stop and fix our software. If there is no such validation, then we can't decide whose fault is it, ours or the batcher's, in which case I don't know what to do.

With that in mind, do you want to just issue a warning and skip?

executor/src/core.rs
aakoshh: It is unfortunate, I'm not sure if there's another way. The only reason it's here is because the `SingleExecutor` instance has to be `Sync` and `Send` so we can call its async methods, and it also cannot be mutable (the traits methods aren't, and it's shared through an `Arc`), so the internally mutable indices must be put in something. Unfortunately `RefCell` cannot be shared across threads, and the next option was the `Mutex`. It is held between async calls, so it's the future specific variant.

It's only ever used by this instance, though, and it's only called sequentially, so there's no risk of a deadlock I think, but it must have a negative performance impact.

executor/src/core.rs
```diff
@@ -106,35 +104,41 @@ where
...
     async fn execute_certificate(&mut self, message: &ConsensusOutput) -> SubscriberResult<()> {
-        // Skip the certificate if it contains no transactions.
-        if message.certificate.header.payload.is_empty() {
-            self.execution_indices.skip_certificate();
-            return Ok(());
+        // Collect all transactions in all the batches.
+        let mut batches = Vec::new();
+
+        for batch_digest in message.certificate.header.payload.keys() {
+            batches.push(self.collect_batch(batch_digest).await?);
```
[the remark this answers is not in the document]
aakoshh: In this instance it would mean stopping the `Core::run` altogether with the `"Failed to run core"` message. 

I know it's not what you want, thanks for bringing it up. See my [other comment](https://github.com/MystenLabs/narwhal/pull/818/files/6432daae2e69d813942837c058d34e0865192035#r949976679) about why it's difficult to say what the right thing to do is. 

For example in Tendermint you have some guarantees about the content because transactions have to conform to Protobuf message formats to be admitted into the blocks, so this problem doesn't come up. 

Arguably if you _know_ what type of `Transaction` you are dealing with, ie. you aren't passing raw bytes to `ExecutionState`, then you might want to think about validating their content up front before batching, and also after downloading it, before casting a vote on the batch. That way if it fails here, we are right to stop because it must in a minority who can't deal with something everyone else okayed already.

But if you trust that most of your nodes will run the same version of the software and you aren't doing any vetting of input from clients, then you're right it's best to skip rubbish transactions than to halt all validators.

executor/src/core.rs
aakoshh: Changed it back to skipping.

executor/src/core.rs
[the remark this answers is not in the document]
aakoshh: Changed the transaction deserialization to skipping. Now we only stop if a batch cannot be retrieved from the store.

executor/src/core.rs
```diff
@@ -143,54 +147,175 @@ where
...
         };
 
+        // The consensus simply orders bytes, so we first need to deserialize the transaction.
+        // If the deserialization fail it is safe to ignore the transaction since all correct
+        // clients will do the same. Remember that a bad authority or client may input random
+        // bytes to the consensus.
+        let transactions = transactions
+            .into_iter()
+            .map(
+                |serialized| match bincode::deserialize::<State::Transaction>(&serialized) {
```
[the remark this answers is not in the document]
aakoshh: You mean to push the deserialization into the `ExecutorState` implementation itself?

Like I said I think letting the system know about the `Transaction` type is a good thing, it keeps your options open for hardening in the future by rejecting malicious input at the peripheries.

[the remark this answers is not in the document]
aakoshh: I highlighted another change in the PR description, which is that `tx_output` will only contain the outcome for transactions that we managed to at least deserialize (previously it had raw bytes, now it's the model). It doesn't look like this would be a problem though, because you weren't able to tell the difference in the errors anyway, so you didn't now whether the raw bytes were legit without trying to deserialize again.

Let me know, though, because pushing deserialization into the state itself could restore full freedom.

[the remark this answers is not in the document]
aakoshh: Moving it out would also put the decision of what serialization to use back with the application, ie. `bincode` would not be prescribed by this library. But you can achieve this by adding your own traits as well.

executor/src/core.rs
aakoshh: On further thought, changed it to return raw bytes at this stage.

executor/src/core.rs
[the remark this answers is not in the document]
aakoshh: Pushed a change which moves the deserialisation into the `SingleExecutor`, to restore the current behaviour of reporting the outcome of all transactions, even if they are malformed. The `BatchExecuttionState` now takes raw transactions, and it's up to the application to decide how much of it to process, and what deserialisation scheme to use.

executor/src/lib.rs
asonnino: Cool, let's leave the TODO then (we haven't fully deployed Narwhal reconfiguration in DevNet yet, so this part may still change)

executor/src/core.rs
asonnino: It's a good point. Validators can input any rubbish as transactions, there are no checks on transactions format upon voting. Before voting, validators simply verify that the payload is available (and that the header doesn't break any safety or GC rules).

You are right that if the transaction serialisation format changes and some validators are late to the party, then it may be a safety problem. Should we open an issue for that and see what others think about it?

aakoshh: Opening an issue can't hurt, at least there would be a story about it it. Someone using this library might want to protect themselves from the clients sending invalid content. If it's encoded into a trait, anyone happy with raw bytes can trivially allow everything in.

asonnino (approved):

PR #823, opened 2022-08-19
crypto: remove all except bls12377

joyqvq:


huitseeker (changes requested):
There is another issue with this: if we let this in, NW would be using an alias for public keys, private keys and signature types that is defined in the fastcrypto crate.

The goal of defining this alias is to ensure we can check the crypto agility of the code base, for example in rewriting it to something different (e.g. Ed25519 -> Bls12381, or any other scheme change), and checking the code base still passes all tests.

This is what we did, for example, to "test" the PoC BLS12-377 implementation, and check that this switch worked + that it did not impact benchmarks.

This switch can therefore not live in an upstream library: the whole point of keeping some code for bls12-377 here is to be able to switch again (and possibly have a long-running nightly integration test that does this automatically).

Long story short: would it be possible to move the alias to this (NW) repo, and check that instantiating it with BLS12-377 (rather than the current Ed25519) doesn't break functionality? This would validate we are keeping here all the elements that are essential to the BLS12377 implementation.

joyqvq:
> There is another issue with this: if we let this in, NW would be using an alias for public keys, private keys and signature types that is defined in the fastcrypto crate.
> 
> The goal of defining this alias is to ensure we can check the crypto agility of the code base, for example in rewriting it to something different (e.g. Ed25519 -> Bls12381, or any other scheme change), and checking the code base still passes all tests.
> 
> This is what we did, for example, to "test" the PoC BLS12-377 implementation, and check that this switch worked + that it did not impact benchmarks.
> 
> This switch can therefore not live in an upstream library: the whole point of keeping some code for bls12-377 here is to be able to switch again (and possibly have a long-running nightly integration test that does this automatically).
> 
> Long story short: would it be possible to move the alias to this (NW) repo, and check that instantiating it with BLS12-377 (rather than the current Ed25519) doesn't break functionality? This would validate we are keeping here all the elements that are essential to the BLS12377 implementation.

good point, also removed from fastcrypto https://github.com/MystenLabs/fastcrypto/pull/10

huitseeker (approved):
Perfect! Thank you @joyqvq this is excellent!

PR #824, opened 2022-08-20
restore certificates in restarted executor

lanvidr:
https://github.com/MystenLabs/sui/issues/5222
https://github.com/MystenLabs/sui/issues/5327


todd-mystenlabs:
Adding François as well to get more eyeballs on this.

todd-mystenlabs:
Looks like code coverage dropped below a threshold.

executor/src/lib.rs
```diff
@@ -153,15 +166,54 @@ impl Executor {
...
+    ) -> Vec<ConsensusOutput>
+    where
+        State: BatchExecutionState + Send + Sync + 'static,
+        State::Error: Debug,
+    {
+        let mut restored_consensus_output = Vec::new();
+        let consensus_max_seq = consensus_store
+            .read_last_consensus_index()
+            .expect("Failed to load consensus index from store");
+
```
akichidis, line 195: looking on the semantics of the `read_last_consensus_index` and how that is being stored/used I believe it would be better to rename the variable to something like `consensus_next_index` , as practically this is what is being stored in there. Looking in the consensus code it seems we assign the current `consensus_index` to a sequenced certificate, we increment its value +1 , and then we store the index in store. So in the end, when you read the `read_last_consensus_index` is the "next" index that will be assigned to the next certificate that will be sequenced.

executor/src/tests/executor_tests.rs
```diff
@@ -322,3 +328,120 @@ async fn crash_recovery() {
...
+
+    // Ensure the execution state is as expected.
+    let _ = rx_output.recv().await;
+    let expected = ExecutionIndices {
+        next_certificate_index: 7,
+        next_batch_index: 0,
+        next_transaction_index: 0,
+    };
+    assert_eq!(execution_state.get_execution_indices().await, expected);
+}
```
akichidis, line 447: This test looks great! Could you please also add a test for the `Executor` as a whole component (not only the `Core` one? At the moment we don't test the recovery part that has to do with the `get_restored_consensus_output` and I believe is essential

akichidis (dismissed):
Thanks so much for this @laura-makdah ! So I believe we are missing some important bits that have to do with the certificate processing it self, so let me give some info around it. 

* When a certificate gets sequenced from `Consensus`, then the next part is to be sent to the `Subscriber`. 
* The `Subscriber` is "listening" to all those sequenced certificates via the `rx_consensus` channel and has as its main mission to "download" their batches by fetching them from the worker nodes. 
* Once the respective batches have been downloaded, _and only then_, they are handed over to `Core`  via the `tx_executor` channel in the same order those have been received from Consensus. Please see the [description](https://github.com/MystenLabs/narwhal/blob/437b3b25a929b7e28695da1a0b334d5a97acbc74/executor/src/subscriber.rs#L27) in the Subscriber for more info

If a certificate reaches `Core` without its batches been previously downloaded, then we'll be looking to a misbehaviour.

So when you recover the certificates that have been sequenced, but not processed via the `Core`, we have to make sure that we run them first via the `Subscriber` to ensure that we have their batches available.

Given the above I would probably point you to perform the recovery by sending all those `ConsensusOutput` messages to the `Subscriber` (in order) and then let the protocol play as is (so maybe you won't need to touch the `Core` at all).

----
**P.S** flagged as "Request changes" as merging this will probably lead to a "broken" behaviour - wanted to make sure this is captured.

executor/src/core.rs
```diff
@@ -64,21 +66,27 @@ where
...
             .expect("Failed to load execution indices from store");
 
-        // TODO: Replay certificates from the store.
+        // Process any messages that were restored after a restart, by order of sequence.
+        for message in restored_messages
+            .into_iter()
+            .sorted_by(|a, b| a.consensus_index.cmp(&b.consensus_index))
+        {
+            self.execute_certificate(&message).await?;
+        }
```
asonnino, line 89: Is there a scenario where this loop can take a very long time?

executor/src/lib.rs
```diff
@@ -153,15 +166,54 @@ impl Executor {
...
+                for cert_digest in missing.iter().flatten() {
+                    if let Some(cert) = certificate_store.read(*cert_digest).await.unwrap() {
+                        // Save the missing sequence / cert pair as ConsensusOutput to re-send to the executor.
+                        restored_consensus_output.push(ConsensusOutput {
+                            certificate: cert,
+                            consensus_index: seq,
+                        })
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
laura-makdah: 👍

executor/src/core.rs
laura-makdah: The longer the executor was down, the greater the amount of restored messages there could be. However, as I understand, it is essential that ensure they are executed in order, right?

executor/src/tests/executor_tests.rs
laura-makdah: I looked into testing `get_restored_consensus_output` in isolation, which would involve setting up the consensus and certificate stores, and making sure that the ones sequenced after the certificate index are returned. I think this is better tested manually than via unit test, since the amount of test code required to set up the two databases with state is disproportionate to the two operations in the function.

test_utils/src/lib.rs
```diff
@@ -14,11 +14,11 @@ use futures::Stream;
 use indexmap::IndexMap;
 use multiaddr::Multiaddr;
 use rand::{rngs::StdRng, Rng, SeedableRng as _};
+use std::sync::Arc;
 use std::{
     collections::{BTreeSet, VecDeque},
     ops::RangeInclusive,
     pin::Pin,
-    sync::Arc,
```
asonnino, line 21: that's strange, have you run `cargo fmt`?

asonnino (approved):
LGTM

test_utils/src/lib.rs
laura-makdah: I have a file watcher that does this automatically

PR #825, opened 2022-08-21
[refactor] replace the certificate_store with a wrapper struct

akichidis:
Currently when we need to fetch the certificates from storage for a specific round , or higher/lower than a round, it is quite inefficient and requires a whole scan of the `certificates` store. As this storage grows (see DevNet) this can be quite costly and can introduce latency. We mainly do that for the recovery of the nodes.

This PR is replacing the certificate_store ( `Store<CertificateDigest, Certificate>`). with a wrapper struct that is keeping as well a secondary index referencing the certificate digests by their rounds. That will allow us to do efficiently scan the stored certificates by their rounds.

The PR is:
- [x] introducing the `CertificateStore` under a dedicated create `storage`. We can move all the related store implementations there in follow up PRs (ex see ConsensusStore, NodeStorage etc)
- [x] is replacing the instances of `Store<CertificateDigest, Certificate>` with the `CertificateStore`
- [x] is leveraging the DBMap iterator method `prev()` introduced as part of this PR https://github.com/MystenLabs/mysten-infra/pull/125 (please review as well)

Note: this PR is quite big but quite many changes just deal with the point 2 from above (replacing the `certificate_store` occurancies) - so nothing fancy there. The main point to focus on should probably be the [CertificateStore](https://github.com/MystenLabs/narwhal/compare/feat/use-certificate-store?expand=1#diff-2515bd4b1b8c546469e49fedb1f2ab47ff66a16e7a1bb445a3273fcc132f077eR1) implementation


storage/src/certificate_store.rs
```diff
@@ -0,0 +1,603 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.certificates_by_id.is_empty()
+    }
+
+    /// Notifies the subscribed ones that listen on updates for the
+    /// certificate with the provided id. The obligations are notified
+    /// with the provided value. The obligation entries under the certificate id
+    /// are removed completely. If we fail to notify an obligation we don't
+    /// fail and we rather print a warn message.
+    fn notify_subscribers(&self, id: CertificateDigest, value: Certificate) {
```
akichidis, line 294: We might want to notify the subscribers in an async way - ex spin up a tokio task (probably used a bound executor / task pool for that) to notify them so we make sure we don't block the write/read path. Right now I don't consider it a huge issue as sending to the `oneshot` channel is non blocking and under normal circumstances we don't expect to have that many `notify_read` for the same values - even if we do I believe we can handle ok.

storage/src/certificate_store.rs
```diff
@@ -0,0 +1,603 @@
...
+            })?;
+
+            certificates.push(certificate);
+        }
+
+        Ok(certificates)
+    }
+
+    /// Retrieves the certificates of the last round
+    pub fn last_round(&self) -> StoreResult<Vec<Certificate>> {
```
akichidis, line 238: @huitseeker

consensus/src/consensus.rs
```diff
@@ -98,14 +98,10 @@ impl ConsensusState {
         );
 
         let min_round = last_committed_round.saturating_sub(gc_depth);
-        let cert_map = cert_store
-            .iter(Some(Box::new(move |(_dig, cert)| {
-                cert.header.round > min_round
-            })))
-            .await;
+        let cert_map = cert_store.after_round(min_round + 1).unwrap();
```
asonnino, line 101: Maybe add a comment saying "get all >= min_round"?

node/src/lib.rs
```diff
@@ -83,6 +87,7 @@ impl NodeStorage {
         ) = reopen!(&rocksdb,
             Self::HEADERS_CF;<HeaderDigest, Header>,
             Self::CERTIFICATES_CF;<CertificateDigest, Certificate>,
+            Self::CERTIFICATE_ID_BY_ROUND_CF;<(Round, CertificateDigest), u8>,
```
asonnino, line 90: What is the `u8` field?

asonnino (approved):
LGTM but we may want to wait for @huitseeker review on the `certificate_store`

node/src/lib.rs
akichidis: @asonnino nothing really - since it's a key-value store we need to have something on the value side. We use this table just for the ordering purposes of the keys by round. That tells me that I might need to use a type alias so to make things a bit clearer

storage/src/certificate_store.rs
huitseeker: Yes, that seems reasonable to me.

consensus/src/consensus.rs
huitseeker: <pedantic> That's actually a > </pedantic>

huitseeker (approved):
This looks great to me, thank you so much @akichidis!
I added a few commits at the end to fix some completely auxiliary efficiency issues.

PR #826, opened 2022-08-21
Fix clippy failure on beta channel

mwtian:
From nightly failure here: https://github.com/MystenLabs/narwhal/runs/7935149802?check_suite_focus=true#step:5:850

Also fix a typo.

### DIFF
--- a/primary/src/header_waiter.rs
+++ b/primary/src/header_waiter.rs
@@ -263,20 +263,15 @@ impl HeaderWaiter {
                     }
                 },
                 // we poll the availability of a slot to send the result to the core simultaneously
-                (Some(result), permit) = try_fut_and_permit!(waiting.try_next(), self.tx_core) => match result {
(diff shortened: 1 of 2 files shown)


primary/src/header_waiter.rs
```diff
@@ -263,20 +263,15 @@ impl HeaderWaiter {
...
-                        for x in header.payload.keys() {
-                            let _ = self.batch_requests.remove(x);
-                        }
-                        for x in &header.parents {
-                            let _ = self.parent_requests.remove(x);
-                        }
-                        permit.send(header);
-                    },
-                    None => {
-                        // This request has been canceled.
```
huitseeker, line 278: I'd like not to lose that comment, perhaps reinsert it on the line of the closing bracket of the destructuring `if let`?

huitseeker (approved):
Thanks @mwtian

PR #839, opened 2022-08-22
Two small maintenance fixes

huitseeker:
- the long-running tests have been on a livelock [for a while](https://github.com/MystenLabs/narwhal/actions/workflows/nightly.yml), for reasons that are hard to reproduce. This puts them under a reasonable timeout in the nightly job.
- we stop retrying message sends on some clearly irrecoverable errors.
- ~we import fixes from #670 in the hope of fixing the local benchmarks, which have been broken (but the breakage is hard to reproduce) since https://github.com/MystenLabs/narwhal/commit/babd3d8ee6278652a8fee3b41764787203c17cc1~

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -21,4 +21,5 @@ status-level = "skip"
 fail-fast = false
 # Retry failing tests in order to not block builds on flaky tests
 retries = 2
-# no killing slow tests, by design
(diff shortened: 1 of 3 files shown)


arun-koshy (approved):
Thanks for this! Benchmark changes should be in now 😄

Issue #191, opened 2022-04-28
Securely cleanup the executor storage


asonnino:
It is not clear how to security cleanup the executor storage. The following code does not work: https://github.com/MystenLabs/narwhal/blob/fecbfc9385e5599aed504d2b63de4b8763d4b6a2/executor/src/core.rs#L76-L85

The following problem may happen:

1. The subscriber adds a certificate C1 to its waiter until its payload X becomes available in the store
2. The corresponding payload downloaded
3. The executor core is about to execute the payload 
4. The subscriber adds a second certificate C2 to its waiter with the same payload X as certificate C1
5. C1 is executed and the store is purged right before the following future is polled: https://github.com/MystenLabs/narwhal/blob/fecbfc9385e5599aed504d2b63de4b8763d4b6a2/executor/src/subscriber.rs#L160-L170

As a result, C2 remains in the waiter forever. (Yes, this happened during a unit test)

akichidis:
@asonnino is our assumption even in the case of `C1` that we'll _eventually_ receive the payload? I am sure there is some theory behind this, but how much would we break the protocol by putting a `timeout` when waiting for the payloads? Could this be a trigger to "reprocess" the message?

asonnino:
No we can't have the timeout or we break safety or liveness. you break safety is you decide to ignore the timed out transactions and continue with the next ones, or you break liveness if you decide to shut down the node and panic

asonnino:
I think the proper approach to this is a feedback communication between the core and the subscriber (but I haven't had time to properly thing about it)

asonnino:
Also I think that other people using the codebase (somelier/celo?) will have the same issue

huitseeker:
> Also I think that other people using the codebase (somelier/celo?) will have the same issue

I don't think so: we hash intents (including round #) in the digest definition for batches in our spec for them. As a consequence the mapping from headers to their payload should be injective.

Linked to #86 #156 #188

PR #842, opened 2022-08-23
[refactor] clean up executor's temporary batch store

akichidis:
Resolves: https://github.com/MystenLabs/narwhal/issues/191

The same batch can be referenced by multiple certificates, something that doesn't allow us to securely clean up storage after a certificate's execution. This PR is refactoring the temporary storage in order to reference a batch both by its `id` and the corresponding `certificate's id`. So effectively we are storing a certificate's batches under the key tuples `(certificate_id, batch_id)`. Now that allow us to securely clean up the storage whenever we execute a certificate as we have 1 to 1 semantics.

### DIFF
--- a/executor/src/core.rs
+++ b/executor/src/core.rs
@@ -7,14 +7,18 @@ use crate::{
     ExecutionState, ExecutorOutput, SerializedTransaction,
 };
 use consensus::ConsensusOutput;
+use fastcrypto::Hash;
 use std::{fmt::Debug, sync::Arc};
+use store::rocks::TypedStoreError;
 use store::Store;
 use tokio::{
     sync::{mpsc::Sender, watch},
     task::JoinHandle,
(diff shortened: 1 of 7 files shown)


asonnino (approved):
LGTM, I didn't see any concurrency issues between executing multiple copies of the a certificates while cleaning up the store

executor/src/core.rs
```diff
@@ -103,13 +107,32 @@ where
         }
     }
 
+    /// Cleans up the temporary batch store for the batches stored
+    /// for the specified certificate. We are storing the batches per
+    /// certificate as bathes of same id can be referenced by multiple
```
huitseeker, line 112: Nit:typo

huitseeker (approved):
This is excellent & straightforward, thanks for the tests as well!

PR #844, opened 2022-08-24
Committee type safety

punwai:
This makes Committee structs type-safe by ensuring that the list of underlying validators in an epoch is immutable. This is required for the following PRs that we will then later push:
1. Replace Public Keys in certificate with Bitmaps.
2. Replace with vector of signatures with Aggregated Signatures

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3214,6 +3214,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "retain_mut"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4389f1d5789befaf6029ebd9f7dac4af7f7e3d61b69d4f30e2ac02b57e7712b0"
+
+[[package]]
 name = "rfc6979"
 version = "0.3.0"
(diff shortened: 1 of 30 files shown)


huitseeker (changes requested):
The `BTreeMap` is already ordered:
https://doc.rust-lang.org/std/collections/struct.BTreeMap.html
> Iterators obtained from functions such as [BTreeMap::iter](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.iter), [BTreeMap::values](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.values), or [BTreeMap::keys](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.keys) produce their items in order by key, and take worst-case logarithmic and amortized constant time per item returned.

This is unecessary.

punwai:
@huitseeker Do we still want this PR for committee's type safety?

huitseeker:
@punwai if by this you mean making authorities a read-only field (i.e. private with a public getter), yes! But it's a small subset of this PR in the current state.

punwai:
Closing for now

PR #846, opened 2022-08-24
[crypto] Signature aggregations and pk bitmap

punwai:
This PR makes two optimizations.
1. It replaces the vector of public keys with a bitmap that points to which encodes which authorities in the Committee object have signed the certificate
2. We replace the vector of signatures with the `AggregateSignature` type (note that the Ed25519 AggregateSignature is simply a vector of Signatures.

The main change to the certificate data structure is the following:
```
pub struct Certificate {
    pub header: Header,
    pub votes: Vec<(PublicKey, <PublicKey as VerifyingKey>::Sig)>,
}
```
into
```
pub struct Certificate {
    pub header: Header,
    aggregated_signature: AggregateSignature,
    #[serde_as(as = "NarwhalBitmap")]
    signed_authorities: roaring::RoaringBitmap,
}
```


config/src/lib.rs
```diff
@@ -475,14 +480,51 @@ impl std::fmt::Display for Committee {
...
+    pub fn new(authorities: BTreeMap<PublicKey, Authority>, epoch: Epoch) -> Self {
+        let mut ordered_authorities: Vec<PublicKey> =
+            authorities.iter().map(|(pk, _)| pk.clone()).collect();
+
+        ordered_authorities.sort_by_key(|pk| pk.clone());
+
+        let index_map: BTreeMap<PublicKey, u64> = ordered_authorities
+            .iter()
+            .enumerate()
+            .map(|(index, addr)| (addr.clone(), index as u64))
```
joyqvq, line 492: addr should be pubkey?

punwai: Removed this

punwai:
@huitseeker Changes made and ready for review.

types/src/primary.rs
```diff
@@ -381,6 +386,48 @@ impl Certificate {
...
+        let aggregated_signature =
+            AggregateSignature::aggregate(votes.iter().map(|(_, sig)| sig).cloned().collect())
+                .map_err(DagError::InvalidSignature)?;
+
+        Ok(Certificate {
+            header,
+            aggregated_signature,
+            signed_authorities,
+        })
+    }
```
huitseeker, line 429: We will need tests on this. More specifically:
- that a set of votes using a `PublicKey` twice results in an error,
- unless it's providing twice the exact same set of bytes for the signatures in which case we should just log a warning and include the signature once,
- that a set of votes with an aggregated stake lower than the `committee` threshold results in an error, (same logic as in validation)
- please also insert a randomized test (you can use [proptest](https://docs.rs/proptest/latest/proptest/)) showing that given an arbitrary committee, and an arbitrary set of valid votes by 50% of it (such that 50% passes the threshold stake), the output of `Certificate::new` passes `verify()`.

huitseeker (changes requested):
Both the `Certificate::new` and `Certificate::verify` methods will require rigorous testing.

punwai:
@huitseeker Ready for review

huitseeker:
@punwai I'm going to checkout your PR and play with it in order to review. That's a heck of a lot easier to do if CI passes.

huitseeker (commented):
Thanks for the work you put into this. I'll stamp as soon as the proptest can compile and is `#[ignore]` rather than commented as it is now.

punwai:
@huitseeker squashed!

huitseeker (approved):

PR #850, opened 2022-08-25
crypto: zeroize private key in BLS12377

joyqvq:


crypto/src/bls12377/mod.rs
```diff
@@ -592,3 +593,31 @@ impl From<&BLS12377PublicKey> for BLS12377PublicKeyBytes {
         BLS12377PublicKeyBytes::from_bytes(pk.as_ref()).unwrap()
     }
 }
+
+impl zeroize::Zeroize for BLS12377PrivateKey {
+    fn zeroize(&mut self) {
```
kchalkias, line 598: what about the `privkey`? In BLS12_381 we zeroize that too.

joyqvq: talked to pun about this today, in BLS12-381 it is automatically dropped without explicitly drop here. in fact i commented out https://github.com/MystenLabs/fastcrypto/blob/main/src/bls12381.rs#L618 and all tests still passed

joyqvq: added a comment here, this is done by the library under the hood, the test suggested this. in addition, the SecertKey in BLS12377 library prohibits a mutable reference, there is actually no easy to explicitly call zeroize on the object.

crypto/src/bls12377/mod.rs
```diff
@@ -592,3 +593,32 @@ impl From<&BLS12377PublicKey> for BLS12377PublicKeyBytes {
...
+        // PrivateKey.zeroize here is not necessary here because the underlying implicitly zeroizes.
```
kchalkias, line 599: typo: using the word _here_ twice

crypto/src/tests/bls12377_tests.rs
```diff
@@ -370,3 +370,52 @@ async fn signature_service() {
...
+                .iter()
+                .enumerate()
+                .take(CELO_BLS_PRIVATE_KEY_LENGTH)
+            {
+                assert_eq!(*ptr.add(i + 41), byte);
+            }
+        }
+        let sk_memory: &[u8] =
+            unsafe { ::std::slice::from_raw_parts(bytes_ptr, CELO_BLS_PRIVATE_KEY_LENGTH) };
+        // Assert that this is equal to sk_bytes before deletion
```
kchalkias, line 402: Prefer adding period (.) on every comment

kchalkias (approved):
LGTM

PR #851, opened 2022-08-25
Add WorkerInfo Request and rename public primary & worker interface

arun-koshy:
Part of Dynamic Worker Integration -> [Design Doc](https://www.notion.so/mystenlabs/Dynamic-Worker-Design-Doc-160e43560031474bba42d15e84d8e96f#54a6013acf434efba811cd076e381b74)

There are four network interfaces currently being used: PrimaryToPrimary, WorkerToWorker, WorkerToPrimary, PrimaryToWorker. The WorkerToPrimary and PrimaryToWorker interfaces are meant only for messaging between a primary and its workers and vice versa, so these will be considered to be more private or internal interfaces. The PrimaryToPrimary & WorkerToWorker interfaces are the public interfaces that are used to communicate with the external primaries and external workers. As workers will also communicate with external primaries too identify their worker peers we will rename these “public” interfaces to be more clear about the use case.

Also introducing the messages for how `WorkerInfo` will be retrieved from an external primary.

What has been tested - 
- [x] Unit/Integration Tests
- [x] Local & Remote Benchmark
- [x] Client Demo
- [ ] Docker configurations (NW & Sui)
- [x] Sui Integration


bmwill:
Just an FYI that i remember having a chat with @huitseeker about simplifying our lives a bit and merging all of these interfaces to be hosted off of a single network interface. So this was something that I was hoping to be able to do once we migrate to the quic network stack. We can still use ACLs to protect those "private" interfaces of course (especially since the quic stack requires there to be network identities this is easy to do)

huitseeker:
> 
Just an FYI that i remember having a chat with @huitseeker about simplifying our lives a bit and merging all of these interfaces to be hosted off of a single network interface. So this was something that I was hoping to be able to do once we migrate to the quic network stack. We can still use ACLs to protect those "private" interfaces of course (especially since the quic stack requires there to be network identities this is easy to do)

Yup, I see no contradiction with this work, which does not change the number of interfaces, but clarifies their purpose by renaming them (as the `primary_to_primary` interface is soon to be used to carry messages coming from a worker).

primary/src/tests/primary_tests.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2021, Facebook, Inc. and its affiliates
```
huitseeker, line 1: Nit: this one copyright stanza isn't needed in a new file

huitseeker (approved):
Thanks for the care in making sure this works E2E, this is amazing! This will will need a rebase + care in merging, and coordination with merging https://github.com/MystenLabs/sui/pull/4425, as you no doubt have noticed. LGTM!

arun-koshy:
PR as it stands is no longer required now that [PR#4628](https://github.com/MystenLabs/sui/pull/4628) has been merged. Will move the `worker_info` messaging in `PrimaryReceiverHandler` and the tests in separate PR.

PR #856, opened 2022-08-26
Update worker index properly

arun-koshy:


huitseeker (approved):
Ah, I see, it's not just the shadowing. Good fix, @arun-koshy !

PR #857, opened 2022-08-26
[crypto] Switch signature scheme to BLS + Add Networking key.

punwai:
This PR switches the signature scheme for signing consensus transactions to aggregatable BLS12-381 signatures. That said, we now need the validators to own another ed25519 key, `NetworkKeypair`, so that a QUIC connection can be established. Moreover, as the worker's only need their keypairs to interact with the primary, we switch the worker's keypair from BLS12381 to Ed25519.

Must be followed by:
https://github.com/MystenLabs/sui/pull/4408 (sui)
https://github.com/MystenLabs/sui-operations/pull/366 (production configs)

### DIFF
--- a/benchmark/benchmark/config.py
+++ b/benchmark/benchmark/config.py
@@ -155,6 +155,7 @@ class Committee:
                     "primary_to_primary": x.x.x.x:x,
                     "worker_to_primary": x.x.x.x:x,
                 },
+                "network_key: NETWORK_KEY==
             },
             ...
         }
@@ -170,17 +171,17 @@ class Committee:
         assert isinstance(addresses, OrderedDict)
         assert all(isinstance(x, str) for x in addresses.keys())
         assert all(
-            isinstance(x, list) and len(x) >= 1 for x in addresses.values()
+            isinstance(address, list) and len(address) >= 1 for (_, address) in addresses.values()
         )
         assert all(
-            isinstance(x, str) for y in addresses.values() for x in y
+            isinstance(x, str) for (_, address) in addresses.values() for x in address
         )
         assert len({len(x) for x in addresses.values()}) == 1
         assert isinstance(base_port, int) and base_port > 1024
 
         port = base_port
         self.json = {'authorities': OrderedDict(), 'epoch': 0}
-        for name, hosts in addresses.items():
+        for name, (network_name, hosts) in addresses.items():
             host = hosts.pop(0)
             primary_addr = {
                 'primary_to_primary': f'/ip4/{host}/tcp/{port}/http',
@@ -190,7 +191,8 @@ class Committee:
 
             self.json['authorities'][name] = {
                 'stake': 1,
-                'primary': primary_addr
+                'primary': primary_addr,
+                'network_key': network_name
             }
 
     def primary_addresses(self, faults=0):
@@ -241,11 +243,12 @@ class Committee:
 
 
 class LocalCommittee(Committee):
-    def __init__(self, names, port):
+    def __init__(self, names, network_names, port):
         assert isinstance(names, list)
         assert all(isinstance(x, str) for x in names)
         assert isinstance(port, int)
-        addresses = OrderedDict((x, ['127.0.0.1']) for x in names)
+        assert len(names) == len(network_names)
+        addresses = OrderedDict((name, (network_name, ['127.0.0.1'])) for name, network_name in zip(names, network_names))
         super().__init__(addresses, port)
 
 
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -10,7 +10,7 @@
 #![allow(clippy::mutable_key_type)]
 
 use arc_swap::ArcSwap;
-use crypto::PublicKey;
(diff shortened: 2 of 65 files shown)


kchalkias (dismissed):
Awesome! Don't forget to add a meaningful commit msg when merged

huitseeker (changes requested):
This is a earth-shaking change for our code bases, changing the signature scheme from Ed25519 to BLS, and should be managed as such.

1. Please add a meaningful commit message
2. please label as breaking, and announce the breakage where appropriate,
3. Please coordinate with the Sui folks, as this will break NW's ability to receive an Ed25519 key, hence we need the switch to BLS ready there as well,
4. please check the keytool there (sp. for any PoP generation), sui-operations (both Sui and Narwhal configs therein), sui-genesis, so that they are all ready to receive a switch to BLS at the drop of a hat.

punwai:
I'm really sorry about making this PR. I forgot to turn on drafts. I've pushed this branch purely for the reason of seeing diffs while developing.

punwai:
Sorry for the confusion, I was supposed to open up this PR as a draft. I have this branch so that I have a remote revision which I can import into Sui to begin a PR which switches it to BLS there. I did not intend for this PR to be merged. That said, these are super helpful comments and I will ensure that I keep these in mind!

config/src/lib.rs
```diff
@@ -450,6 +450,8 @@ pub struct Authority {
     pub stake: Stake,
     /// The network addresses of the primary.
     pub primary: PrimaryAddresses,
+    /// Network key of the primary.
+    pub network_key: NetworkPublicKey,
```
huitseeker, line 454: 👍

crypto/src/lib.rs
```diff
@@ -26,11 +26,11 @@ pub use fastcrypto::traits;
 // to change all four aliases to point to concrete types that work with each other. Failure to do
 // so will result in a ton of compilation errors, and worse: it will not make sense!
 
-pub type PublicKey = ed25519::Ed25519PublicKey;
-pub type Signature = ed25519::Ed25519Signature;
-pub type AggregateSignature = ed25519::Ed25519AggregateSignature;
-pub type PrivateKey = ed25519::Ed25519PrivateKey;
-pub type KeyPair = ed25519::Ed25519KeyPair;
+pub type PublicKey = bls12381::BLS12381PublicKey;
```
huitseeker, line 29: So, in theory this looks good, but in practice since #909 we also have worker keys. That worker key is used exclusively for the purpose of validating the authentication of workers over the network (TLS 1.3-powered), so it should morally be a `NetworkKey`

At the moment that key is defined in the field `name` of the `WorkerInfo` struct in `config`, which type is `PublicKey`. I think we need to switch `PublicKey` -> `NetworkingKey` there:
https://github.com/MystenLabs/narwhal/blob/a383ad626355035cf5e388be6ee668fff17bb646/config/src/lib.rs#L300-L302

Once that is done, note the changes needed to support worker keys in Sui are in https://github.com/MystenLabs/sui/pull/4541, so that the companion of the present PR (https://github.com/MystenLabs/sui/pull/4408) should be adapted to take 4541 into account.

huitseeker (commented):
This looks fantastic! Thanks a lot for the huge amount of work it takes to bring this over the finish line!

I have only one remark : switching the key type in the `WorkerInfo` to a `NetworkPublicKey`. Otherwise, this LGTM!

bmwill:
Have you also given all primaries their own network key? that would be a blocker for landing this correct?

crypto/src/lib.rs
punwai: Done!

config/src/lib.rs
```diff
@@ -299,7 +299,7 @@ impl Parameters {
 #[derive(Clone, Serialize, Deserialize, Eq, Hash, PartialEq, Debug)]
 pub struct WorkerInfo {
     /// The public key of this worker.
-    pub name: PublicKey,
+    pub name: NetworkPublicKey,
```
huitseeker, line 302: Awesome!

huitseeker (approved):
Thank you so much @punwai! 🎉

benchmark/benchmark/config.py
```diff
@@ -155,6 +155,7 @@ class Committee:
                     "primary_to_primary": x.x.x.x:x,
                     "worker_to_primary": x.x.x.x:x,
                 },
+                "network_key: NETWORK_KEY==
```
kchalkias, line 158: :)

kchalkias (changes requested):
IMO we should have split this PR into "adding Network Key" and "switch to BLS". If we merge it as a single PR, let's please update the title of this PR + commit message to reflect both changes.

benchmark/benchmark/config.py
punwai: Oh yeah, I think that's a bit random haha.

punwai: Yes, this should be merged right before the other one.

kchalkias (approved):
Fantastic!

Issue #858, opened 2022-08-28
Failing benchmarks


huitseeker:
### Description
The benchmarks (run on CI on every pushed commit, see [the script](https://github.com/MystenLabs/narwhal/blob/main/.github/workflows/bench.yml)) are failing since https://github.com/MystenLabs/narwhal/commit/babd3d8ee6278652a8fee3b41764787203c17cc1 (aka https://github.com/MystenLabs/narwhal/pull/738).

This is impossible to reproduce outside Github for me. I had to resort to [hacks]

PR #859, opened 2022-08-29
Revert 818 + increase the batch timeout

huitseeker:
- MystenLabs/narwhal#818 introduced a Batch/Single Executor organization, and attempting to port it in https://github.com/MystenLabs/sui/pull/4219 has faced issues (as the livelocked CI will show)
- We have also observed that `test_full_node_sync_flood` in Sui had its runtime multiplied by 7 on a simple update to the precise commit of MystenLabs/narwhal#818.

We have not managed to clear up the issue, and would therefore like to revert MystenLabs/narwhal#818, allowing more time to consider it for a later inclusion.

*Besides this*, we have noticed that the introduction of the block waiter in the executor (#738) has created issues (#858) in the speed of recovery when a batch transmission hits the batch request timeout. This bumps that batch request timeout, mitigating MystenLabs/sui#5212 as well as Sui issues suspected to be linked to this.

The hope is that this PR will allow closing https://github.com/MystenLabs/sui/pull/4219, and rebasing https://github.com/MystenLabs/sui/pull/4316 (which aims to update the NW pointer in Sui past the merge of MystenLabs/narwhal#670) on the merge commit of the present PR.

/cc @aakoshh


primary/src/block_waiter.rs
```diff
@@ -30,7 +30,7 @@ use types::{
 };
 use Result::*;
 
-const BATCH_RETRIEVE_TIMEOUT: Duration = Duration::from_secs(1);
+const BATCH_RETRIEVE_TIMEOUT: Duration = Duration::from_secs(10);
```
arun-koshy, line 33: Is this a temporary fix? If so should we add a TODO here to change it back with some tracking issue?

arun-koshy (approved):

primary/src/block_waiter.rs
huitseeker: Done!

PR #881, opened 2022-09-01
chore(ci): don't auto-close issues

huitseeker:
As this already tripped us on MystenLabs/sui#5318


mwtian (approved):
Maybe have a comment that this workflow is currently disabled.

PR #890, opened 2022-09-06
chore(ci): Auto update mysten-infra

huitseeker:
This is an adaptation of https://github.com/MystenLabs/sui/pull/4393 to automatically update the `fastcrypto` & `mysten_infra` pointer in NW.

The next step (after this PR) would be to write a job that updates the fastcrytpo pointer in Sui. Because this `fastcrypto` & `mysten_infra` pointer in Sui *must* match the `fastcrypto` & `mysten_infra` pointer in NW (the 2 libraries exchange keys w/ each other), I would actually advise making this part of the `scripts/update_narwhal.sh` script:
- we would find the `fastcrypto` & `mysten_infra` hash in NW (a variant of the `current_nw_revision` function),
- and if it's not the same as the `fastcrypto` & `mysten_infra` hash in Sui (found through a variant of the `current_nw_revision` but applied to a different repo),
- we would update the pointer to that in NW (possibly checking it's newer, see `latest_nw_revision`)

@Edit: this now only updates mysten-infra, fastcrypto to be updated through dependabot


bmwill:
what if we release fastcrypto to crates.io (using version 0.0.X) and let dependabot update fastcrypto?

bmwill (approved):

huitseeker:
@bmwill Good point! I'll socialize this.

Issue #891, opened 2022-09-06
Failure of test `test_restart_with_new_committee_change`


huitseeker:
Failures are very consistent in CI on windows:
https://github.com/MystenLabs/narwhal/pull/886
https://github.com/MystenLabs/narwhal/pull/888
https://github.com/MystenLabs/narwhal/pull/889
https://github.com/MystenLabs/narwhal/pull/887

Failure message:
https://gist.github.com/huitseeker/4c2a0ed94eaf151c009cb00c4c7ff6e2

The bisection points to:
https://github.com/MystenLabs/narwhal/pull/668

PR #892, opened 2022-09-06
primary: don't panic on inbound RPCs when the node is shutting down

bmwill:
Fixes a bug where the primary would panic if an RPC comes in while a
node is shutting down. Instead this patch just returns an error to
requester.

Fixes: #891


huitseeker (approved):
Looks great!
Thanks @bmwill

PR #898, opened 2022-09-06
Add heap usage tracking metric for DAG.

aschran:
Towards MystenLabs/sui#5213.

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -315,6 +315,15 @@ where
                             tracing::info!("Committed {} -> {:?}", certificate.header, digest);
                         }
 
+                        // Update DAG size metric periodically to limit computation cost.
+                        // TODO: this should be triggered on collection when library support for
+                        // closure metrics is available.
+                        if output.consensus_index % 1_000 == 0 {
+                            self.metrics
+                                .dag_size_bytes
+                                .set((mysten_util_mem::malloc_size(&state.dag) + std::mem::size_of::<Dag>()) as i64);
+                        }
+
                         self.tx_primary
                             .send(certificate.clone())
                             .await
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -857,6 +857,7 @@ dependencies = [
  "futures",
  "indexmap",
  "match_opt",
+ "mysten-util-mem",
  "pprof",
  "prometheus",
  "rand 0.8.5",
@@ -1079,6 +1080,7 @@ dependencies = [
  "hkdf",
  "k256",
  "merlin",
+ "mysten-util-mem",
  "once_cell",
  "proptest",
  "proptest-derive",
@@ -1994,6 +1996,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "impl-trait-for-tuples"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "11d7a9f6330b71fea57921c9b61c47ee6e84f72d394754eff6163ae67e7395eb"
+dependencies = [
+ "proc-macro2 1.0.40",
+ "quote 1.0.20",
+ "syn 1.0.98",
+]
+
+[[package]]
 name = "indenter"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


consensus/src/consensus.rs
```diff
@@ -325,6 +325,9 @@ where
                         .consensus_dag_rounds
                         .with_label_values(&[])
                         .set(state.dag.len() as i64);
+                    self.metrics
+                        .dag_size_bytes
+                        .set((mysten_util_mem::malloc_size(&state.dag) + std::mem::size_of::<Dag>()) as i64);
```
huitseeker, line 330: What is the order of magnitude of the error we expect from the second argument of the sum?

huitseeker (commented):
So we have a [micro-bench in the consensus](https://github.com/MystenLabs/narwhal/blob/main/consensus/benches/process_certificates.rs), which you can run with 
```bash
cargo-criterion  --bench process_certificates --features "benchmark, pprof"
```
The benchmark can also create flamegraphs, which should help in measuring where the CPU time goes along this processing, see [this blog post](https://www.jibbow.com/posts/criterion-flamegraphs/) for details.

That benchmark is based on `process_certificates`, so it exercises none of the logic that you add, which is based on receiving a message in the `run` loop of the consensus, which then calls `process_certificates`, followed by your metric update.

But I'm wondering if it would be possible to tweak the benchmark itself to call perhaps an internal auxiliary function in the measurement loop, which would compute
```rust
(mysten_util_mem::malloc_size(&state.dag) + std::mem::size_of::<Dag>()) as i64
```
and then measure the e2e impact of this on the overall cost of processing one certificate.

What I'm trying to figure out is the overall computational burden of this metric computation on the performance-sensitive consensus loop. If it's negligible, fine. If it's not, there are several ways to go about diminishing this burden, among which:
- sampling (updating the metric once in a few calls),
- memoizing (getting the latest value out of the metric and only adding the size of the elements we added to this add-only structure)

Deciding which to go with is essentially predicated on having a quantitative sense of this computational cost.

consensus/src/consensus.rs
aschran: Probably quite small, but this is a fixed value so I'd hope that the call to size_of is inlined as a constant? Not sure how to check that though.

huitseeker: We actually came up on this a while ago: https://github.com/MystenLabs/sui/pull/3920#discussion_r943290604

aschran: In what way do you mean you came up on it? Clearly the heap allocated part needs to be measured (which is what that discussion seemed to be about), but what's the harm in also adding the stack allocated part to the measurement? Is there a nuance I'm missing here?

huitseeker: Sorry about that, my comment above was misguided and not very constructive, I was more thinking out loud, and that led you on a wild goose chase, which I apologize for.

aschran:
> So we have a [micro-bench in the consensus](https://github.com/MystenLabs/narwhal/blob/main/consensus/benches/process_certificates.rs), which you can run with
> 
> ```shell
> cargo-criterion  --bench process_certificates --features "benchmark, pprof"
> ```
> 
> The benchmark can also create flamegraphs, which should help in measuring where the CPU time goes along this processing, see [this blog post](https://www.jibbow.com/posts/criterion-flamegraphs/) for details.
> 
> That benchmark is based on `process_certificates`, so it exercises none of the logic that you add, which is based on receiving a message in the `run` loop of the consensus, which then calls `process_certificates`, followed by your metric update.
> 
> But I'm wondering if it would be possible to tweak the benchmark itself to call perhaps an internal auxiliary function in the measurement loop, which would compute
> 
> ```rust
> (mysten_util_mem::malloc_size(&state.dag) + std::mem::size_of::<Dag>()) as i64
> ```
> 
> and then measure the e2e impact of this on the overall cost of processing one certificate.
> 
> What I'm trying to figure out is the overall computational burden of this metric computation on the performance-sensitive consensus loop. If it's negligible, fine. If it's not, there are several ways to go about diminishing this burden, among which:
> 
> * sampling (updating the metric once in a few calls),
> * memoizing (getting the latest value out of the metric and only adding the size of the elements we added to this add-only structure)
> 
> Deciding which to go with is essentially predicated on having a quantitative sense of this computational cost.

I added sampling for now. The benchmark showed enough of a performance drop that it seemed worthwhile. I think the right long term solution is to add support for a metric that computes its value at collection time, at that point we should switch this over. (I may work on that soon since it'd be useful for other purposes too.)

huitseeker (approved):
Thanks a crapton for running through the paces of benchmarking the perf impact! I really appreciate it, @aschran !
The sampling looks great, and so does your thinking on how to further progress on this.

PR #899, opened 2022-09-06
test-utils: introduce CommitteeFixture and friends

bmwill:
Today for testing in narwhal we heavily rely on a number of fixture generation utility functions for generating keys, committees, etc. One present issue with them is that if you can a test helper you generally don't know if it's expecting to use a fixture generated at a fixed seed or not. This could result in the helper using a different committee than you expected.

This pr introduces the concept of a CommitteeFixture as well as corresponding types for workers and primaries in order to make testing helpers a bit more transparent and easier to use.

### DIFF
--- a/test_utils/src/lib.rs
+++ b/test_utils/src/lib.rs
@@ -14,10 +14,12 @@ use fastcrypto::{
 use futures::Stream;
 use indexmap::IndexMap;
 use multiaddr::Multiaddr;
+use rand::rngs::OsRng;
 use rand::{rngs::StdRng, Rng, SeedableRng as _};
 use std::sync::Arc;
 use std::{
-    collections::{BTreeSet, VecDeque},
+    collections::{BTreeMap, BTreeSet, VecDeque},
+    num::NonZeroUsize,
     ops::RangeInclusive,
     pin::Pin,
 };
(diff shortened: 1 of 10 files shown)


config/src/lib.rs
```diff
@@ -298,6 +298,7 @@ impl Parameters {
 
 #[derive(Clone, Serialize, Deserialize, Eq, Hash, PartialEq, Debug)]
 pub struct WorkerInfo {
+    // pub public_key: PublicKey,
```
aschran, line 301: Add a TODO to explain this maybe? Or was it meant to be deleted?

test_utils/src/lib.rs
```diff
@@ -936,6 +938,274 @@ pub fn mock_signed_certificate(
             .unwrap();
         votes.push((pk.clone(), sig))
     }
-    let cert = Certificate::new(&committee(None), header, votes).unwrap();
+    let cert = Certificate::new(committee, header, votes).unwrap();
     (cert.digest(), cert)
 }
+
+use rand::rngs::OsRng;
```
aschran, line 945: should this go at the top?

test_utils/src/lib.rs
```diff
@@ -936,6 +938,274 @@ pub fn mock_signed_certificate(
...
+
+pub struct Builder<R = OsRng> {
+    rng: R,
+    committee_size: NonZeroUsize,
+    number_of_workers: NonZeroUsize,
+    randomize_ports: bool,
```
aschran, line 951: this reads like a function name to me, it's more verbose but maybe consider "should_randomize_ports"? or feel free to ignore :)

test_utils/src/lib.rs
bmwill: They don't have to go at the top. Depending on the situation it may may sense to have a use inside a function or a block. In this case it probably makes the most sense at the top of the file :)

config/src/lib.rs
bmwill: ah thanks this should be removed

test_utils/src/lib.rs
bmwill: I don't have a strong preference but do lean slightly towards the shorter name 🤷🏼‍♂️

aschran (approved):

PR #900, opened 2022-09-06
Revert #738 & #842

### LINKED ISSUE #858: Failing benchmarks

### LINKED ISSUE #5212: Failing benchmarks
### Description
The benchmarks (run on CI on every pushed commit, see [the script](https://github.com/MystenLabs/narwhal/blob/main/.github/workflows/bench.yml)) are failing since https://github.com/MystenLabs/narwhal/commit/babd3d8ee6278652a8fee3b41764787203c17cc1 (aka https://github.com/MystenLabs/narwhal/pull/738).

This is impossible to reproduce outside Github for me. I had to resort to [hacks](https://github.com/huitseeker/narwhal/commit/0a78e0ab51e0fb48406e6649c5138f75a3550f34) in order to reproduce this.

This generated the following log collating all the logs from the primaries:
https://gist.github.com/huitseeker/c9e94b03f1ded5155ab6f7fae8197a9c

This shows a *lot* of ERROR messages:
```

2022-08-28T00:28:40.938450005Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: WEoJCTeNr8hGZH6h, error type: BatchTimeout
2022-08-28T00:28:41.009256736Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: ZSmM9/zTy0rchjVe, error type: BatchTimeout
2022-08-28T00:28:41.18922286Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 3tBtWhKDXo0Z1lOt, error type: BatchTimeout
2022-08-28T00:28:41.190253959Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: PWvIQwgrjsBU9zVh, error type: BatchTimeout
2022-08-28T00:28:41.190262359Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 6cdRBZ0BjSzCKTD/, error type: BatchTimeout
2022-08-28T00:28:41.190270059Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: wZmmum8O9Fji92bQ, error type: BatchTimeout
2022-08-28T00:28:41.190276659Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: GDpuvWoV3imlr11J, error type: BatchTimeout
2022-08-28T00:28:41.233743717Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: T4aZ3TVPYNOquRTy, error type: BatchTimeout
2022-08-28T00:28:43.355264748Z ERROR handle_get_block_command{block_id=kR1FgPrIAvPBgeUI1hlJy36+hIafP2B+6mBCFMGdMG4=}: primary::block_synchronizer::handler: Error for payload synchronization with block id kR1FgPrIAvPBgeUI, error: Payload for block with kR1FgPrIAvPBgeUI couldn't be synchronized: Block with id kR1FgPrIAvPBgeUI could not be retrieved, timeout while retrieving result
2022-08-28T00:28:43.355795947Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: kR1FgPrIAvPBgeUI, error type: BatchError
2022-08-28T00:28:43.355811547Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: ez1A69+h0dSPWy9S, error type: BatchTimeout
2022-08-28T00:28:45.358290095Z ERROR handle_get_block_command{block_id=8fBpctxds6AsKN0wiRB3VVhFgMOkRHNgLbh7pUjYX8Y=}: primary::block_synchronizer::handler: Error for payload synchronization with block id 8fBpctxds6AsKN0w, error: Payload for block with 8fBpctxds6AsKN0w couldn't be synchronized: Block with id 8fBpctxds6AsKN0w could not be retrieved, timeout while retrieving result
2022-08-28T00:28:45.359836393Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 8fBpctxds6AsKN0w, error type: BatchError
2022-08-28T00:28:45.359848193Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: +SOaQXV+9d25O7Xr, error type: BatchTimeout
2022-08-28T00:28:45.359854893Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: RlIGqn8nBJiuZzsP, error type: BatchTimeout
2022-08-28T00:28:45.359861493Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 5EZZEkGzUplc5Fbm, error type: BatchTimeout
2022-08-28T00:28:45.359867393Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: ZisR6+IKv3Ox2V9W, error type: BatchTimeout
2022-08-28T00:28:45.359878593Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: oBvDNGqJ24KwLVft, error type: BatchTimeout
2022-08-28T00:28:45.360496692Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: j5w8zYnFTaEfp/Pe, error type: BatchTimeout
2022-08-28T00:28:45.360664692Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 9AQp+iKRg/9xRJbl, error type: BatchTimeout
2022-08-28T00:28:45.360672392Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: cBPOlHz8S0e0qqZ/, error type: BatchTimeout
2022-08-28T00:28:45.360678192Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: Y/sf8ljv0fjuoxc1, error type: BatchTimeout
2022-08-28T00:28:45.362037691Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: KK2rt+Z4auYyVRCR, error type: BatchTimeout
2022-08-28T00:28:45.362047091Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 3y26qDLPg5jkZN5R, error type: BatchTimeout
2022-08-28T00:28:45.362053491Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: hMobCuAHhM7ERFMX, error type: BatchTimeout
2022-08-28T00:28:45.362059691Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 1PZ6pzwpMQ4PvmHG, error type: BatchTimeout
2022-08-28T00:28:45.362065391Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: xASlHPo863vlw4dc, error type: BatchTimeout
2022-08-28T00:28:45.36319459Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: yX+oXs2WkyNsdGxU, error type: BatchTimeout
2022-08-28T00:28:45.36320419Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: 5PPiV86Hvj9NMIXP, error type: BatchTimeout
2022-08-28T00:28:45.36320979Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: LvUmM69H51TG8CNi, error type: BatchTimeout
2022-08-28T00:28:45.36321569Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: Yey4HkTQC7xUGJTq, error type: BatchTimeout
100 2397k  100 2397k    0     0  8789k      0 --:--:-- --:--:-- --:--:-- 8911k
2022-08-28T00:28:47.366147134Z ERROR handle_get_block_command{block_id=t8Bvt5jwALnig7b8j+fIfIdFvdt2TPpyo+FnGOqZRCo=}: primary::block_synchronizer::handler: Error for payload synchronization with block id t8Bvt5jwALnig7b8, error: Payload for block with t8Bvt5jwALnig7b8 couldn't be synchronized: Block with id t8Bvt5jwALnig7b8 could not be retrieved, timeout while retrieving result
2022-08-28T00:28:47.366415633Z ERROR executor::subscriber: Error while retrieving block via block waiter: block id: t8Bvt5jwALnig7b8, error type: BatchError
```

huitseeker:
This means to start a discussion re: reverting MystenLabs/narwhal#842 and MystenLabs/narwhal#738 (with the former reverted for building on top of the later)

On the benchmarks, in CI, this shows a 64% increase in throughput and 60% decrease in latency:
- the base of this PR (main) shows [17K TPS at 1.6s latency](https://github.com/MystenLabs/narwhal/commit/e2cae6382c0b988195990d3102c549dfdbdd6a30#commitcomment-83224735)
- the tip of this PR shows [28K TPS at 1.0s latency](https://github.com/MystenLabs/narwhal/commit/293cecd0b997df937ad141f7e70f0f4b7bf5b2e8#commitcomment-83225149)

At home, on a (voluntarily) puny laptop (which will do much better on e.g. serde):
- the base of this PR (main) shows [42.9K TPS at 709ms latency](https://gist.github.com/huitseeker/31963368870f5d707b48fb5dd745e05d)
- the tip of this PR shows [45.4K TPS at 500ms latency](https://gist.github.com/huitseeker/733b621ef40886ce080b9499f8ae22e8)

It also seems this might improve some long-running integration tests:
- [before](https://gist.github.com/huitseeker/7dac7a894cd93054a06af513b00cb6ea)
- [after](https://gist.github.com/huitseeker/0aea1bb43dc8b14f7380869939ff93d9)

This fixes MystenLabs/sui#5212, and might improve MystenLabs/sui#5217.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1485,12 +1485,12 @@ name = "executor"
 version = "0.1.0"
 dependencies = [
  "async-trait",
- "backoff",
  "bincode",
  "blake2",
  "bytes",
  "config",
  "consensus",
+ "crypto",
  "fastcrypto",
  "futures",
  "indexmap",
--- a/executor/Cargo.toml
+++ b/executor/Cargo.toml
@@ -23,10 +23,10 @@ tokio-util = { version = "0.7.3", features = ["codec"] }
 tonic = "0.7.2"
 tracing = "0.1.36"
 prometheus = "0.13.1"
-backoff = { version = "0.4.0", features = ["tokio"] }
 storage = { path = "../storage" }
 itertools = "0.10.3"
 
+crypto = { path = "../crypto" }
 types = { path = "../types" }
 mysten-network = { git = "https://github.com/mystenlabs/mysten-infra.git", rev = "352091e92070c2ecfcccad444361a78249ecfe59" }
 
@@ -38,7 +38,6 @@ match_opt = "0.1.2"
 indexmap = { version = "1.9.1", features = ["serde"] }
 rand = "0.8.5"
 tempfile = "3.3.0"
-primary = { path = "../primary" }
 test_utils = { path = "../test_utils" }
 types = { path = "../types" }
 telemetry-subscribers = { git = "https://github.com/mystenlabs/mysten-infra.git", rev = "352091e92070c2ecfcccad444361a78249ecfe59" }
--- a/executor/src/batch_loader.rs
+++ b/executor/src/batch_loader.rs
@@ -1,0 +1,204 @@
+// Copyright (c) 2022, Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use crate::errors::{SubscriberError, SubscriberResult};
+
+use config::WorkerId;
+use consensus::ConsensusOutput;
+use futures::stream::StreamExt;
+use multiaddr::Multiaddr;
+use std::collections::{HashMap, HashSet};
+use store::Store;
+use tokio::{
+    sync::{
+        mpsc::{channel, Receiver, Sender},
+        watch,
+    },
+    task::JoinHandle,
+};
+use tracing::{error, warn};
+use types::{
+    metered_channel, serialized_batch_digest, BatchDigest, BincodeEncodedPayload,
+    ClientBatchRequest, ReconfigureNotification, SerializedBatchMessage, WorkerToWorkerClient,
+};
+
+/// Download transactions data from the consensus workers and notifies the called when the job is done.
+pub struct BatchLoader {
+    /// The temporary storage holding all transactions' data (that may be too big to hold in memory).
+    store: Store<BatchDigest, SerializedBatchMessage>,
+    /// Receive reconfiguration updates.
+    rx_reconfigure: watch::Receiver<ReconfigureNotification>,
+    /// Receive consensus outputs for which to download the associated transaction data.
+    rx_input: metered_channel::Receiver<ConsensusOutput>,
+    /// The network addresses of the consensus workers.
+    addresses: HashMap<WorkerId, Multiaddr>,
+    /// A map of connections with the consensus workers.
+    connections: HashMap<WorkerId, Sender<Vec<BatchDigest>>>,
+}
+
+impl BatchLoader {
+    /// Spawn a new batch loaded in a dedicated tokio task.
+    #[must_use]
+    pub fn spawn(
+        store: Store<BatchDigest, SerializedBatchMessage>,
+        rx_reconfigure: watch::Receiver<ReconfigureNotification>,
+        rx_input: metered_channel::Receiver<ConsensusOutput>,
+        addresses: HashMap<WorkerId, Multiaddr>,
+    ) -> JoinHandle<()> {
+        tokio::spawn(async move {
+            Self {
+                store,
+                rx_reconfigure,
+                rx_input,
+                addresses,
+                connections: HashMap::new(),
+            }
+            .run()
+            .await
+            .expect("Failed to run batch loader")
+        })
+    }
+
+    /// Receive consensus messages for which we need to download the associated transaction data.
+    async fn run(&mut self) -> SubscriberResult<()> {
+        loop {
+            tokio::select! {
+                // Receive sync requests.
+                Some(message) = self.rx_input.recv() => {
+                    let certificate = &message.certificate;
+
+                    // Send a request for every batch referenced by the certificate.
+                    // TODO: Can we write it better without allocating a HashMap every time?
+                    let mut map = HashMap::with_capacity(certificate.header.payload.len());
+                    for (digest, worker_id) in certificate.header.payload.iter() {
+                        map.entry(*worker_id).or_insert_with(Vec::new).push(*digest);
+                    }
+                    for (worker_id, digests) in map {
+                        let address = self
+                            .addresses
+                            .get(&worker_id)
+                            .ok_or(SubscriberError::UnexpectedWorkerId(worker_id))?;
+
+                        let sender = self.connections.entry(worker_id).or_insert_with(|| {
+                            let (sender, receiver) = channel(primary::CHANNEL_CAPACITY);
+                            SyncConnection::spawn(
+                                address.clone(),
+                                self.store.clone(),
+                                receiver,
+                            );
+                            sender
+                        });
+
+                        sender
+                            .send(digests)
+                            .await
+                            .expect("Sync connections are kept alive and never die");
+                    }
+                }
+
+                // Check whether the committee changed.
+                result = self.rx_reconfigure.changed() => {
+                    result.expect("Committee channel dropped");
+                    let message = self.rx_reconfigure.borrow().clone();
+                    if let ReconfigureNotification::Shutdown = message {
+                        return Ok(());
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Connect (and maintain a connection) with a specific worker. Then download batches from that
+/// specific worker.
+struct SyncConnection {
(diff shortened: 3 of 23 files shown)


asonnino:
Are we re-introducing the batch-loader instead of the batch waiter?

asonnino (approved):

akichidis:
Regarding the batch_loader and the performance changes (@huitseeker we talked a little bit offline about those today) . A couple of info regarding that which unfortunately didn't express more vocally on my PR MystenLabs/narwhal#738 (added in description but probably got lost there). 

Executor [was configured](https://github.com/MystenLabs/narwhal/commit/babd3d8ee6278652a8fee3b41764787203c17cc1#diff-c1fa00b27f461be84b2e8f9f93974c238229bb7a87964886c0fe316ed825f4e0L299) to use the [batch_store](https://github.com/MystenLabs/narwhal/blob/ef92fc0d455494e7b97aaa02a528356f45169a24/node/src/lib.rs#L47) which as part of MystenLabs/narwhal#738 is changed to use a separate store (the `temp_batch_store` was introduced) to keep the batches fetched from the workers. The `batch_store` is the exact same one that [a worker node is using as well](https://github.com/MystenLabs/narwhal/blob/ef92fc0d455494e7b97aaa02a528356f45169a24/node/src/lib.rs#L393). In a SUI deployment when we initialise the Narwhal's storage we do it only once and we pass the same `NodeStorage` instance to both the `worker` and the `primary` nodes. I see that currently we use the [NodeRestarter](https://github.com/MystenLabs/sui/blob/dced2e77ac9f93e66311ba9bd536e820d868696e/crates/sui-core/src/authority_server.rs#L267) - before the nodes were initialised manually - but the underneath practice is the same, NodeRestarter will pass [the same store instance to both nodes](https://github.com/MystenLabs/narwhal/blob/ef92fc0d455494e7b97aaa02a528356f45169a24/node/src/restarter.rs#L84). So practically both the `primary` and `worker` would share the same storage. The worker would use it to store in the `batch_store` its (and others) batches and the primary would use the same `batch_store`, via the Executor, to effectively store the "fetched" batches from the worker.

The [Subscriber](https://github.com/MystenLabs/narwhal/commit/babd3d8ee6278652a8fee3b41764787203c17cc1#diff-dd570646b526219ac76c68c9546526c43df1bf27085a0949d1355a15bd5d84b8L65) was implemented to first check in the `batch_store` about the required certificate's batches. If all existed (`notify_read` etc) then the [certificate would be delivered](https://github.com/MystenLabs/narwhal/commit/babd3d8ee6278652a8fee3b41764787203c17cc1#diff-dd570646b526219ac76c68c9546526c43df1bf27085a0949d1355a15bd5d84b8L65) for execution. Now, given that the `Subscriber` would look directly to the worker's store (batch_store) my understanding is that we would never observe any sort of network issues - or at least get affected by them as we practically never leverage the network communication for that purpose. Also the `batch_loader` wouldn't really do much , apart from fetching via network the batch from the worker and "rewriting" the same data again to the same (shared) store - `batch_store`. So if we revert the PRs lets keep in mind that we'll re-introduce the above behaviour - both primary & worker use the exact same store (`batch_store`) to keep their batches. Although it will make things definitely more robust (as we are talking now about a monolith) it will give us the false impression about the primary <--> worker communication when we request the batches which I would strongly recommend then refactor asap.

huitseeker:
Yep,I think the overall outcome we want to get to is a better story for the `BlockWaiter` in the presence of large messages, i.e. an Executor-focused variant of https://github.com/MystenLabs/sui/issues/5300. I no longer think this PR, which was there for expediency anyway, is the best way to address this.

PR #909, opened 2022-09-07
Add Worker Keys

arun-koshy:
Part of Dynamic Worker Integration -> [Design Doc](https://www.notion.so/mystenlabs/Dynamic-Worker-Design-Doc-160e43560031474bba42d15e84d8e96f#54a6013acf434efba811cd076e381b74)

Give keys to workers as part of genesis and store the public keys as part of the `WorkerInfo`. Eventually when workers can be dynamically discovered the keys will not be provided at genesis but rather created dynamically in code.

Worker reconfiguration requires some additional thought and the bare minimum was done in this PR in that regard. Only allowed an avenue for the updated worker information to be sent from Sui to NW via `NodeRestarter`s channels. Leaving proper worker reconfiguration changes for a future PR. @asonnino let me know if that is okay.

What has been tested - 
- [x] Unit/Integration Tests
- [x] Local & Remote Benchmark
- [x] Client Demo
- [x] Docker configurations (NW)
- [ ] Sui Integration


config/tests/snapshots/config_tests__worker_cache.snap
```diff
@@ -6,87 +6,103 @@ expression: worker_cache
   "workers": {
     "ildi4hrBzbOHBELHe0w69Yx87bh3nQJw5tTx4vc2fXQ=": {
       "0": {
+        "name": "zGIzLjS7LVzWn2Dvuyo2y5FsfrRYMB6jZjbE27ASvYg=",
         "transactions": "/ip4/127.0.0.1/tcp/2/http",
         "worker_to_worker": "/ip4/127.0.0.1/tcp/3/http",
         "primary_to_worker": "/ip4/127.0.0.1/tcp/1/http"
       },
       "1": {
+        "name": "ildi4hrBzbOHBELHe0w69Yx87bh3nQJw5tTx4vc2fXQ=",
```
bmwill, line 15: This doesn't quite look right but this is in a snap-shot file so maybe its just that?

in particular the pubkey for worker 1 here is the same as the primary and you can see that the pubkeys repeat for other workers for other primaries. They'll all need to be unique from each other.

node/tests/reconfigure.rs
```diff
@@ -66,17 +66,30 @@ impl ExecutionState for SimpleExecutionState {
...
+            let keypair = k
                 .into_iter()
                 .enumerate()
                 .filter(|(i, _)| i == &self.index)
                 .map(|(_, x)| x)
                 .collect::<Vec<_>>()
                 .pop()
                 .unwrap();
+
+            let worker_keypairs = keys(None);
```
bmwill, line 80: Yeah this probably will cause some issues where the worker keys need to be unique from each other and from their primary's keys. maybe when we generate worker keys we can use a seed to start or you can look at using the infra i just added here: https://github.com/MystenLabs/narwhal/pull/899

worker/src/worker.rs
```diff
@@ -60,7 +64,8 @@ const INADDR_ANY: Ipv4Addr = Ipv4Addr::new(0, 0, 0, 0);
 
 impl Worker {
     pub fn spawn(
-        name: PublicKey,
+        primary_name: PublicKey,
+        keypair: KeyPair,
```
bmwill, line 68: perfect! Thanks for threading that through here.

bmwill (commented):
This looks great! I only had the one question about how we generate these keys for workers inside of our tests. Maybe to start (so that we can quickly get this landed) we can use the `keys` function with a seed that is unique from the primaries and is unique among the workers. eg:

primary 0 worker 0: `keys(Some(100))`
primary 0 worker 1: `keys(Some(101))`

primary 1 worker 0: `keys(Some(110))`
primary 1 worker 1: `keys(Some(111))`

bmwill:
Also, converting Sui to use this change should hopefully be fairly easy since each node just has 1 worker so we'd just need to generate an extra key as apart of the node config

config/tests/snapshots/config_tests__worker_cache.snap
arun-koshy: Thanks for fixing it for me 😄

node/tests/reconfigure.rs
arun-koshy: I tried to use the new infra for this test but the test kept hanging (I believe it wasn't an issue with the keys but with how I was updating the committee epoch 🤷🏽 ). So I will just use the old method with a new seed for the worker keys to unblock this PR for now.

arun-koshy: Test is fixed now 👍🏽

bmwill (approved):
🚀

config/src/lib.rs
```diff
@@ -298,6 +298,8 @@ impl Parameters {
 
 #[derive(Clone, Serialize, Deserialize, Eq, Hash, PartialEq, Debug)]
 pub struct WorkerInfo {
+    /// The public key of this worker.
+    pub name: PublicKey,
```
huitseeker, line 302: I think this new field breaks sui genesis, where we construct such a `WorkerInfo`

huitseeker (commented):
This should be flagged as breaking-change.

bmwill:
Yep, I believe that @arun-koshy is planning on making the changes in sui next.

PR #910, opened 2022-09-07
chore(deps): Switch to fastcrypto published version

huitseeker:


joyqvq (approved):
we are at 0.1.2 now on latest, but good to test if dependabot picks up correctly

PR #911, opened 2022-09-07
chore(deps) Update the Mysten-infra pointer

huitseeker:
- Update Mysten-infra pointer

Auto-generated by [create-pull-request][1] an the scripts/update_dependencies.sh script

[1]: https://github.com/peter-evans/create-pull-request

PR #915, opened 2022-09-07
chore(deps): switch to packages for mysten-infra

huitseeker:
Continues #910 
Subsumes & obsoletes #911


executor/Cargo.toml
```diff
@@ -41,4 +41,4 @@ tempfile = "3.3.0"
 primary = { path = "../primary" }
 test_utils = { path = "../test_utils" }
 types = { path = "../types" }
-telemetry-subscribers = { git = "https://github.com/mystenlabs/mysten-infra.git", rev = "352091e92070c2ecfcccad444361a78249ecfe59" }
+telemetry-subscribers = "0.1.0"
```
bmwill, line 44: :/ one thing that this makes me realize is that i don't really like the names we've selected for these things that have now been published to crates.io. They should probably be prefixed with `mysten-` but 🤷🏼‍♂️

bmwill (approved):

mwtian (approved):

PR #924, opened 2022-09-08
[crypto] Update fastcrypto, re-export its traits module

huitseeker:
This will ease the Sui-NW interop on key type conversions


crypto/src/lib.rs
```diff
@@ -10,6 +10,9 @@
 
 use fastcrypto::ed25519;
 
+// This re-export allows using the trait-defined APIs
```
kchalkias, line 13: <3

kchalkias (approved):

joyqvq (approved):

PR #936, opened 2022-09-08
Do not block workers during unreliable_send

andll:
`UnreliableNetwork::unreliable_send` invokes blocking RPC on remote peer in order to send the message. In order to make the RPC call `unreliable_send` spawns a new task on the bounded executor, so that `unreliable_send` normally is not blocked on remote peer processing the RPC.

However, if remote peer is slow or unresponsive, the bounded executor can get full. In this case `UnreliableNetwork::unreliable_send` will block the caller until executor has permits available.

This leads to a situation where single slow/unresponsive peer can block entire critical task by not replying on send message RPC, if too many of them are in-flight.

We [see](http://grafana.shared.internal.sui.io:3000/explore?left=%7B%22datasource%22:%22zIREmpMVz%22,%22queries%22:%5B%7B%22expr%22:%22%7B__name__%3D~%5C%22tx_.%2A%5C%22,job%3D%5C%22mh1-suival-2%5C%22%7D%3E200%22,%22refId%22:%22A%22,%22interval%22:%22%22,%22editorMode%22:%22code%22,%22range%22:true,%22instant%22:true%7D,%7B%22refId%22:%22B%22,%22editorMode%22:%22code%22,%22expr%22:%22primary_network_available_tasks%7Bjob%3D%5C%22mh1-suival-2%5C%22%7D%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%221662553513874%22,%22to%22:%221662586403070%22%7D%7D&orgId=1) this happening in the stuck cluster. The single slow peer `ty6-suival-1` here completely blocks `helper` task on `mh1-suival-2`, so that this validator can not reply to requests from other (healthy) peer.

The solution is to use `try_spawn` and if the bounded executor for worker is full we simply drop the message.

It is hard to predict whether this particular fix will completely solve the issue we are seeing, but the behavior seem to be more in line with how BFT system should handle unresponsive peers.

Some implementation details:

Ideally we want to just remove return value from `UnreliableNetwork::unreliable_send`, but it is currently used by some unit tests and reconfigure.rs. We can figure out later on how to fix this.

### DIFF
--- a/network/src/primary.rs
+++ b/network/src/primary.rs
@@ -94,7 +94,7 @@ impl UnreliableNetwork for PrimaryToWorkerNetwork {
         address: Multiaddr,
         message: BincodeEncodedPayload,
     ) -> Option<JoinHandle<()>> {
-        let mut client = self.client(address.clone());
+        let mut client = self.client(address);
         let handler = self
             .executor
             .try_spawn(async move {
--- a/network/src/primary.rs
+++ b/network/src/primary.rs
@@ -93,14 +93,14 @@ impl UnreliableNetwork for PrimaryToWorkerNetwork {
         &mut self,
         address: Multiaddr,
         message: BincodeEncodedPayload,
-    ) -> JoinHandle<()> {
+    ) -> Option<JoinHandle<()>> {
         let mut client = self.client(address.clone());
         let handler = self
             .executor
-            .spawn(async move {
+            .try_spawn(async move {
                 let _ = client.send_message(message).await;
             })
-            .await;
+            .ok();
 
         self.update_metrics();
 
--- a/network/src/traits.rs
+++ b/network/src/traits.rs
@@ -24,13 +24,13 @@ pub trait UnreliableNetwork: BaseNetwork {
         &mut self,
         address: Multiaddr,
         message: BincodeEncodedPayload,
-    ) -> JoinHandle<()>;
+    ) -> Option<JoinHandle<()>>;
 
     async fn unreliable_send(
         &mut self,
         address: Multiaddr,
         message: &Self::Message,
-    ) -> JoinHandle<()> {
+    ) -> Option<JoinHandle<()>> {
         let message =
             BincodeEncodedPayload::try_from(message).expect("Failed to serialize payload");
         self.unreliable_send_message(address, message).await
@@ -47,8 +47,12 @@ pub trait UnreliableNetwork: BaseNetwork {
             BincodeEncodedPayload::try_from(message).expect("Failed to serialize payload");
         let mut handlers = Vec::new();
         for address in addresses {
-            let handle = { self.unreliable_send_message(address, message.clone()).await };
-            handlers.push(handle);
+            if let Some(handle) = self.unreliable_send_message(address, message.clone()).await {
+                // todo - eventually unreliable_xxx should stop returning join handlers
+                // Currently there are few places (unit tests and p2w network reconfigure) that uses it so having this as a work around
+                // This current usage is broken anyway so we are not making things much worse here
+                handlers.push(handle);
+            }
         }
         handlers
     }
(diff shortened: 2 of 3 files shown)


huitseeker (commented):
Thanks @andll for this PR! This is a good stab at fixing step 1. of #930.
There are further things I'd love to see. First:
> Ideally we want to just remove return value from UnreliableNetwork::unreliable_send, but it is currently used by some unit tests and reconfigure.rs. We can figure out later on how to fix this.

What is the blocker towards fixing that in this PR? Your envisioned approach seems good.

Second, would it be possible to create a unit test that creates an instance of a network using the bounded executor with a non-default tiny limit (e.g. 2, it may require tweaking the constructor), connects to a  non responding (loopback) address,  and sends more messages that the limit all at once, checking all sends conclude successfully under a timeout smaller than the message send timeout? This would effectively check that `unreliable_send_message` is non-blocking irrespective of whether we are past the semaphore limit or not.
WDYT?

andll:
>What is the blocker towards fixing that in this PR? Your envisioned approach seems good.

If you look how this join handle is used in reconfigure.rs, i don't think this can be easily fixed in this PR. Since it does not introduce new problems, I think this PR can be kept as is.
In terms of how to fix this in the future I think we should just remove Primary2Worker network and replace it with rust channels, so the issue will go away. When this is done we can remove return values from those functions.

> Second, would it be possible to create a unit test that creates an instance of a network using the bounded executor with a non-default tiny limit (e.g. 2, it may require tweaking the constructor), connects to a non responding (loopback) address, and sends more messages that the limit all at once

What would be the purpose of this unit test? Are we testing `BoundedExecutor` implementation? I don't think you can cover every possible aspect of the code with unit test, and it feels like what you propose will probably be more work to write and support then any potential benefit

huitseeker:
> In terms of how to fix this in the future I think we should just remove Primary2Worker network and replace it with rust channels, so the issue will go away. When this is done we can remove return values from those functions.

I don't think this is a possibility for the foreseeable future, since the  ability to have multiple workers on distinct machines is pivotal to NW performance. Further, even if it was an option, bounded tokio channel sends by default become blocking when the channel is full, same as the bounded executor here: we'd essentially have to solve the same issue.

> What would be the purpose of this unit test? Are we testing BoundedExecutor implementation?

We are testing that the network implementation (by which I mean the 2 implementations of the Unreliable network trait, that you are modifying), while maintaining a concurrency limit, are not making sends blocking. The assumption that those sends are non blocking in any situation is pervasive throughout the code base (we had a long offline meeting on our Tokio select patterns today that went over this in depth), the crux of the bug MystenLabs/sui#5204, and before your fix in the present PR, wrong.

dmitri-perelman:
1. We all agree that the fix is a step in the right direction.
2. There is tech debt associated with the change, which should be paid at some point (in this PR, or in later PRs).

One approach is to apply this fix now, and to explicitly mention the tech debt (return value of the function and the unit tests) in the corresponding issue (#930). This way, we can resume the iterative experimentations on the network while not sacrificing the long-term quality of the codebase.

huitseeker:
@dmitri-perelman another approach is to try to come up with an attempt at doing it right: https://github.com/MystenLabs/narwhal/pull/940

That extension PR may be missing something, but it was an interesting ~1h  spent thinking about the problem, as far as I'm concerned. If it turns out in review that it takes much more work than MystenLabs/narwhal#940 to address things comprehensively, then applying a partial fix may be appropriate indeed.

I see the general wisdom in checkpointing work, I do not see why it applies this early to save 1 hour, especially since we can't ship either of MystenLabs/narwhal#936 or MystenLabs/narwhal#940 to Sui main for another few hours just yet (the NW -> Sui pipe is blocked on other adjustment work).

If experiments on the network are what is needed, they do not require using main, and should not be blocked on stamps. In fact, we have already shipped the fix in the present PR to a test network, which proves that point.

dmitri-perelman:
Sounds good, let's review the extension PR MystenLabs/narwhal#940!

huitseeker:
Closing in favor of MystenLabs/narwhal#940, which I'll merge as soon as Alberto has had a fair chance to have a look.

PR #937, opened 2022-09-08
Increase `BlockSynchronizer` timeouts

mwtian:
Currently `BlockSynchronizer` steps have 2s timeout. They seem a bit short considering
- Roundtrip to other nodes may take multiple secs
- Other nodes may be busy
- Large payload may need to be loaded from disk.

For `certificates_synchronize_timeout` and `payload_availability_timeout`, only one response is needed within timeout to make progress. However 2s still seems too short. Maybe these timeouts can be 10s~20s instead.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -177,18 +177,32 @@ impl Default for ConsensusAPIGrpcParameters {
     }
 }
 
+fn block_synchronizer_params_default_value() -> Duration {
+    Duration::from_secs(30)
+}
+
 #[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(default)]
 pub struct BlockSynchronizerParameters {
     /// The timeout configuration when requesting certificates from peers.
-    #[serde(with = "duration_format")]
+    #[serde(
+        with = "duration_format",
+        default = "block_synchronizer_params_default_value"
+    )]
     pub certificates_synchronize_timeout: Duration,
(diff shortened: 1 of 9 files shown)


config/src/lib.rs
```diff
@@ -177,36 +177,53 @@ impl Default for ConsensusAPIGrpcParameters {
     }
 }
 
+fn block_synchronizer_params_default_value() -> Duration {
```
huitseeker, line 180: Would it be possible to avoid using block_synchronizer_params_default_value, and rather have 4 distinct default values (which can be consts defined in an `impl BlockSynchronizerParameters` block), one for each timeout, referred to on both the the serialize and the implementation of `Default`?

huitseeker (commented):
I'm in favor of increasing the default delays for the block synchronizer indeed, and something in the dozen of seconds makes sense to me. 

These were originally thought of for the purpose of a node doing historical queries on blocks seconds in the past already, not for streaming block data as blocks go through consensus, which explains the discrepancy.

However, I'm not sure those timeouts should all have the same default, as some timeouts reflect starkly different realities.

config/src/lib.rs
mwtian: Agree that we may need to adjust these values independently, although I don't have a good sense of the default to use other than 30s for now. Changed to separate default values.

huitseeker (approved):
Tip top, thanks @mwtian !

PR #940, opened 2022-09-09
Do not block workers during unreliable_send (extended)

huitseeker:
This is an extension of MystenLabs/narwhal#936 that:
- avoids the cumbersome `JoinHandle`s in `unreliable_send_message` 
- includes rustdoc & tests for the non-blocking behavior of `unreliable_send_message`, 
- switches the reconfiguration message from primary to worker to a reliable send (the reliable send implementation is not de novo, of course), since we rely on its delivery to sequence reconfiguration => asking @asonnino for review,

### DIFF
--- a/network/src/primary.rs
+++ b/network/src/primary.rs
@@ -41,6 +41,15 @@ impl PrimaryToWorkerNetwork {
         }
     }
 
+    // used for testing non-blocking behavior
+    #[cfg(test)]
+    fn new_with_concurrency_limit(concurrency_limit: usize) -> Self {
+        Self {
+            executor: BoundedExecutor::new(concurrency_limit, Handle::current()),
+            ..Self::default()
+        }
+    }
+
     fn update_metrics(&self) {
         if let Some(m) = &self.metrics {
             m.set_network_available_tasks(self.executor.available_capacity() as i64, None);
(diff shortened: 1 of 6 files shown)


andll (approved):
If we can use a reliable broadcast in primary2worker network for reconfiguration then yes, it seem reasonable to switch to it so that we can get rid of `Vec<JoinHandle<()>>`.

network/src/primary.rs
```diff
@@ -93,18 +106,63 @@ impl UnreliableNetwork for PrimaryToWorkerNetwork {
...
+
+#[async_trait]
+impl ReliableNetwork for PrimaryToWorkerNetwork {
+    // Safety
+    // Since this spawns an unbounded task, this should be called in a time-restricted fashion.
+    // Here the callers are [`PrimaryNetwork::broadcast`] and [`PrimaryNetwork::send`],
+    // at respectively N and K calls per round.
+    //  (where N is the number of primaries, K the number of workers for this primary)
+    // See the TODO on spawn_with_retries for lifting this restriction.
+    async fn send_message(
```
dmitri-perelman, line 129: I have a question about the new reliable send implementation: what's the reason to introduce it here? Do I understand it right that it's not solving the issue described in #930, i.e., one irresponsive validator would drain the bounded executor out of tickets?

huitseeker: > what's the reason to introduce it here? 

It's used in two places, [restart](https://github.com/MystenLabs/narwhal/pull/940/files#diff-40cc4e55db0adaed3d6a5aa9728a109866d47486a89131e2ccfbf3cf472f1c72R104) and [reconfiguration](https://github.com/MystenLabs/narwhal/pull/940/files#diff-1f96f5aebfa12edc1699a5a69f0ed2817092622810dc9de207b3287713b37c60R306), which allow the `Primary` to communicate with its `Worker`s. 

Those two places drive network restarts and network reconfigurations (epoch changes). The consequence of losing a network message due to e.g. a transient error in those contexts is a zombie worker.

The difference between reliable sends and unreliable sends (besides the concurrency limitation behavior) is that the reliable send retries on transient errors.

Hence, it's seems sensible on first approach to use a reliable send here.

> Do I understand it right that it's not solving the issue described in #930 ?

That is correct: if the `BoundedExecutor` runs out of tickets, then yes, the reliable send on the `PrimaryToWorker` interface will become blocking.

> i.e., one irresponsive validator would drain the bounded executor out of tickets?

No. The behavior of other validators is irrelevant here, on this `PrimaryToWorker` interface. For the bug described in #930 to come into play, you would need one worker to:
1. request so much information from its primary that the primary fills up its `BoundedExecutor` to the concurrency limit with responses (_mass responses_),
2. behave so that those responses to stay on the semaphore, i.e. the worker does not acknowledge them (_irresponsivity_)
3. Then yes, the primary's answer to *other* workers would be blocked, and consequently a restart / reconfiguration would be delayed (_interface-wide impact_).

However, there are a few reasons this scenario is implausible or irrelevant:
- in cases where we deploy a single worker per primary (as in today's Sui deployments), the _interface-wide impact_ is irrelevant since there are no other workers to speak to than the one that's irresponsive,
- in cases where we co-locate workers with the primary (as in today's Sui deployments), the _irresponsiveness_ of workers is implausible,
- the _mass responses_ is implausible, since the worker is never (for now) the _source_ of responses from the primary (it's rather the worker than answers requests from its primary, as the name indicates). For such a mass of worker-bound messages to happen, it would need to originate in primary-driven requests: concretely, we would have a code path that triggers a _flood_ of restarts / reconfig messages, in which case the network behavior may be the least of our problems,
- but even if that flood were the case, forcing the primary to block on its network semaphore before sending even more requests might be a good idea.

With that said there is for sure a leftover issue of addressing the other half of #930 in a further PR, as laid out in this issue.

dmitri-perelman: Thanks for the context!

huitseeker:
Heard back from offline from @asonnino  on reliability of primary-worker connection, this is good to go.

PR #941, opened 2022-09-09
network: convert worker-to-primary interface to quic/anemo

### LINKED ISSUE #884: network: convert worker-to-primary interface to quic/anemo
We recently converted the primary-to-primary interface to use a new p2p quic networking stack (anemo) in https://github.com/MystenLabs/narwhal/pull/668. As a next step we should convert the worker-to-primary interface that is hosted on the primary and is the interface that worker's send information to their primaries. This should be done by doing the following:
- Instantiate an anemo network on the workers by using a randomly generated key at startup (@arun-koshy is currently working on assigning ed25519 keys to all workers at genesis time. While we're waiting on that work to land we can just use a random key on workers since the primaries won't need to know a stable key to send messages on the existing gRPC primary-to-worker interface)
- On the primary, convert the worker-to-primary server from gRPC to anemo and add the handler to the already existing anemo network that was added in https://github.com/MystenLabs/narwhal/pull/668. This will mean that we'll be able to remove the extra networking interface on the primaries (as well as in the configs). (e.g today primaries bind on two ports, primary-to-primary and worker-to-primary, since we'll be able to host these two services on the same port we can drop the worker-to-primary information from configs).

aschran:
Fixes #884

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -468,27 +468,19 @@ impl WorkerCache {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct PrimaryAddresses {
-    /// Address to receive messages from other primaries (WAN).
-    pub primary_to_primary: Multiaddr,
-    /// Address to receive messages from our workers (LAN).
-    pub worker_to_primary: Multiaddr,
-}
-
-#[derive(Clone, Serialize, Deserialize, Debug)]
+#[derive(Clone, Serialize, Deserialize, Debug, Eq, PartialEq)]
 pub struct Authority {
     /// The voting power of this authority.
     pub stake: Stake,
-    /// The network addresses of the primary.
-    pub primary: PrimaryAddresses,
+    /// The network address of the primary.
+    pub primary_address: Multiaddr,
     /// Network key of the primary.
     pub network_key: NetworkPublicKey,
 }
 
 pub type SharedCommittee = Arc<ArcSwap<Committee>>;
 
-#[derive(Clone, Serialize, Deserialize, Debug)]
+#[derive(Clone, Serialize, Deserialize, Debug, Eq, PartialEq)]
 pub struct Committee {
     /// The authorities of epoch.
     pub authorities: BTreeMap<PublicKey, Authority>,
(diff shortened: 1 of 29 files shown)


worker/src/tests/worker_tests.rs
```diff
@@ -54,7 +54,7 @@ async fn handle_clients_transactions() {
     // Spawn a `Worker` instance.
     Worker::spawn(
         name.clone(),
-        worker_keypair,
+        worker_keypair.copy(),
```
bmwill, line 57: technically i don't think we need to copy this

worker/src/worker.rs
```diff
@@ -45,7 +51,6 @@ pub struct Worker {
     /// The public key of this authority.
     primary_name: PublicKey,
     // The private-public key pair of this worker.
-    // TODO: utilize keypair in network communication
     #[allow(dead_code)]
```
bmwill, line 54: remove this annotation too

bmwill (commented):
looks great! Just a couple things to maybe fix and a test failure to fix.

node/src/restarter.rs
```diff
@@ -89,12 +88,13 @@ impl NodeRestarter {
...
-                .worker_to_primary;
-            let message = WorkerPrimaryMessage::Reconfigure(ReconfigureNotification::Shutdown);
-            let primary_cancel_handle = primary_network.send(address, &message).await;
+            // TODO: This needs to be refactored to work with anemo.
+            // let address = committee
+            //     .primary(&name)
+            //     .expect("Our key is not in the committee")
+            //     .worker_to_primary;
+            // let message = WorkerPrimaryMessage::Reconfigure(ReconfigureNotification::Shutdown);
+            // let primary_cancel_handle = primary_network.send(address, &message).await;
```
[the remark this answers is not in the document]
aschran: Would you have any concern with using randomly generated private keys outside of test code?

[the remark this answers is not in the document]
bmwill: In order to do this conversion, no i don't have any concerns with using a randomly generated key just to send this shutdown message. But lets place a TODO here in the code to rethink how this shutdown message to sent to the primary.

config/src/lib.rs
```diff
@@ -468,27 +468,19 @@ impl WorkerCache {
...
-    pub worker_to_primary: Multiaddr,
-}
-
-#[derive(Clone, Serialize, Deserialize, Debug)]
+#[derive(Clone, Serialize, Deserialize, Debug, Eq, PartialEq)]
 pub struct Authority {
     /// The voting power of this authority.
     pub stake: Stake,
-    /// The network addresses of the primary.
-    pub primary: PrimaryAddresses,
```
bmwill, line 484: I *think* there may need to be some configs that need to be changed in the narwhal repo, specifically for the benchmark, due to this change.

node/tests/reconfigure.rs
```diff
@@ -189,6 +189,7 @@ async fn run_client(
 }
 
 #[tokio::test]
+#[ignore]
```
bmwill, line 192: lets un-ignore this test.

bmwill (approved):

PR #946, opened 2022-09-10
worker: convert worker-to-worker network to quic/anemo

bmwill:


types/src/tests/batch_serde.rs
```diff
@@ -83,17 +83,3 @@ fn test_bincode_serde_batch_message() {
...
-    #[test]
-    fn test_batch_and_serialized(
-        batch in Batch::arbitrary()
-    ) {
-        let digest = batch.digest();
-        let message = WorkerMessage::Batch(batch);
-        let serialized = bincode::serialize(&message).expect("Failed to serialize our own batch");
-        let digest_from_serialized = serialized_batch_digest(&serialized).expect("Failed to hash serialized batch");
-        assert_eq!(digest, digest_from_serialized);
-    }
```
huitseeker, line 98: The goal of this is to check the computation of the digest on the serialized form of the batch is equivalent to that on the batch itself. Moreover, we have some benches proving they take the same time. TL;DR, it's important to not lose this test.

types/benches/batch_digest.rs
```diff
@@ -1,44 +0,0 @@
-// Copyright (c) 2022, Mysten Labs, Inc.
-// SPDX-License-Identifier: Apache-2.0
-use criterion::{
-    criterion_group, criterion_main, BenchmarkId, Criterion, SamplingMode, Throughput,
-};
-use fastcrypto::Hash;
-use rand::Rng;
-use types::{serialized_batch_digest, Batch, WorkerMessage};
-
-pub fn batch_digest(c: &mut Criterion) {
```
huitseeker, line 10: This is the benchmark that checks the equivalence in performance between computing on a serialized batch and a deserialized one. This piece of code gets us away from a Great Debate over whether we should define the digest over the deserialized form of the Batch (as it is with nearly every other piece of data in the code base) or if we should only define it over the serialized bytes (to save on having to deserialize for the sole purpose of computing a digest).

TL;DR: it's important to not lose this code.

huitseeker (commented):
We should chat offline about the purpose of the `SerializedBatch` structure in general.

types/src/tests/batch_serde.rs
bmwill: While taking a look at this i discovered that this file is not included in the module hierarchy and thus is not compiled and these tests are not run on main today.

bmwill:
@huitseeker Ok i've added back that test and the corresponding tests to verify correctness. I also fixed it so that those batch serialization tests were actually run, turned out that they weren't being compiled and tested on main today.

I ran the benchmark on my M1 laptop and verified that the perf between the two impls is indeed identical (or rather very close to identical) which does support the changes that this PR makes in moving away from leveraging the more error prone process of passing around and calculating digests based on serialized batches vs using the Batch type directly.

I do understand that there would be a slight perf win from reading the raw serialized batch from the DB and then just sending that over the wire instead of going through a de/ser pass before writing it to the wire. Today this is sort of possible because we're just sending those bytes wrapped in a protobuf (although that does go through a protobuf ser stage before hitting the wire). This optimization is possible with anemo's request/response paradigm but does require writing a bit more code by hand vs leveraging the generic code-gen process this PR uses. Given this migration to a new networking stack is a large endeavor already I'd be inclined to postpone doing this small optimization  till after the network migration is complete (as well as after some rearchitecting of how nodes communicate which will happen once the migration is finished).

types/src/tests/batch_serde.rs
huitseeker: 🤦

huitseeker (approved):
This LGTM! thanks so much @bmwill !

Before:
```
-----------------------------------------
 SUMMARY:
-----------------------------------------
 + CONFIG:
 Faults: 0 node(s)
 Committee size: 4 node(s)
 Worker(s) per node: 1 worker(s)
 Collocate primary and workers: True
 Input rate: 50,000 tx/s
 Transaction size: 512 B
 Execution time: 20 s

 Header size: 1,000 B
 Max header delay: 200 ms
 GC depth: 50 round(s)
 Sync retry delay: 10,000 ms
 Sync retry nodes: 3 node(s)
 batch size: 500,000 B
 Max batch delay: 200 ms
 Max concurrent requests: 500,000

 + RESULTS:
 Consensus TPS: 48,729 tx/s
 Consensus BPS: 24,949,377 B/s
 Consensus latency: 631 ms

 End-to-end TPS: 48,254 tx/s
 End-to-end BPS: 24,706,255 B/s
 End-to-end latency: 814 ms
-----------------------------------------
```

After:
```
-----------------------------------------
 SUMMARY:
-----------------------------------------
 + CONFIG:
 Faults: 0 node(s)
 Committee size: 4 node(s)
 Worker(s) per node: 1 worker(s)
 Collocate primary and workers: True
 Input rate: 50,000 tx/s
 Transaction size: 512 B
 Execution time: 20 s

 Header size: 1,000 B
 Max header delay: 200 ms
 GC depth: 50 round(s)
 Sync retry delay: 10,000 ms
 Sync retry nodes: 3 node(s)
 batch size: 500,000 B
 Max batch delay: 200 ms
 Max concurrent requests: 500,000

 + RESULTS:
 Consensus TPS: 49,633 tx/s
 Consensus BPS: 25,412,182 B/s
 Consensus latency: 447 ms

 End-to-end TPS: 49,151 tx/s
 End-to-end BPS: 25,165,264 B/s
 End-to-end latency: 630 ms
-----------------------------------------
```

PR #947, opened 2022-09-11
[crypto] bls12377 code cleanup + fix duplicated code

kchalkias:


crypto/src/tests/bls12377_tests.rs
```diff
@@ -81,22 +135,6 @@ fn verify_invalid_signature() {
     assert!(kp.public().verify(digest.as_ref(), &signature).is_err());
 }
 
-fn signature_test_inputs() -> (Vec<u8>, Vec<BLS12377PublicKey>, Vec<BLS12377Signature>) {
-    // Make signatures.
-    let message: &[u8] = b"Hello, world!";
-    let digest = message.digest();
-    let (pubkeys, signatures): (Vec<BLS12377PublicKey>, Vec<BLS12377Signature>) = keys()
```
kchalkias, line 88: We had 3 copies of that snippet

crypto/src/bls12377/mod.rs
```diff
@@ -483,9 +483,9 @@ impl Default for BLS12377AggregateSignature {
 }
 
 impl AggregateAuthenticator for BLS12377AggregateSignature {
-    type PrivKey = BLS12377PrivateKey;
-    type PubKey = BLS12377PublicKey;
     type Sig = BLS12377Signature;
+    type PubKey = BLS12377PublicKey;
```
kchalkias, line 487: We should maintain the members order according to trait for consistency and readability.

huitseeker (approved):
FYI, this is a PoC aimed just at proving we can use something like https://github.com/celo-org/celo-bls-snark-rs in NW. But I'm grateful for the cleanup!

PR #976, opened 2022-09-13
primary: convert primary-to-worker interface to quic/anemo

bmwill:


network/src/traits.rs
```diff
@@ -1,109 +1,13 @@
...
-            let handle = self.send_message(address, message.clone()).await;
-            handlers.push(handle);
-        }
-        handlers
-    }
-}
 
 #[async_trait]
-pub trait UnreliableNetwork2<Message: Clone + Send + Sync> {
+pub trait UnreliableNetwork<Message: Clone + Send + Sync> {
```
aschran, line 10: Would it be worth removing the `2` suffix now that these are the only one again? Or do we not really care?

node/src/restarter.rs
```diff
@@ -48,7 +48,16 @@ impl NodeRestarter {
         let mut committee = committee.clone();
 
         let mut handles = Vec::new();
-        let mut worker_network = PrimaryToWorkerNetwork::default();
+        let network = anemo::Network::bind("127.0.0.1:0")
+            .server_name("narwhal")
+            .private_key(
+                NetworkKeyPair::generate(&mut rand::rngs::OsRng)
```
aschran, line 54: Maybe add a short explanation for why we generate a random private key here and why it's ok?

node/src/restarter.rs
```diff
@@ -114,21 +123,29 @@ impl NodeRestarter {
...
-                .map(|x| x.primary_to_worker)
-                .collect();
-            let message = PrimaryWorkerMessage::Reconfigure(ReconfigureNotification::Shutdown);
-            let worker_cancel_handles = worker_network.broadcast(addresses, &message).await;
+            {
+                let address = network::multiaddr_to_address(&worker.worker_address).unwrap();
+                let peer_id = anemo::PeerId(worker.name.0.to_bytes());
+                network
+                    .connect_with_peer_id(address, peer_id)
+                    .await
```
aschran, line 142: Is there a way this could be refactored to start all the connections in parallel instead of serially? (Or is that already how this works and I'm misunderstanding the semantics of `connect_wtih_peer_id`?)

node/tests/reconfigure.rs
```diff
@@ -323,7 +323,16 @@ async fn epoch_change() {
         let name_clone = name.clone();
         let worker_cache_clone = worker_cache.clone();
         tokio::spawn(async move {
-            let mut worker_network = PrimaryToWorkerNetwork::default();
+            let network = anemo::Network::bind("127.0.0.1:0")
+                .server_name("narwhal")
+                .private_key(
```
aschran, line 328: Can this use `random_network`?

primary/src/block_synchronizer/mod.rs
```diff
@@ -900,6 +895,7 @@ impl BlockSynchronizer {
         let mut peers = Peers::<Certificate>::new(SmallRng::from_entropy());
 
         loop {
+            println!("hello");
```
aschran, line 898: remove?

primary/src/block_synchronizer/mod.rs
```diff
@@ -950,6 +946,7 @@ impl BlockSynchronizer {
                     }
                 },
                 () = &mut timer => {
+                    println!("timer");
```
aschran, line 949: remove?

test_utils/src/lib.rs
```diff
@@ -917,3 +919,12 @@ pub fn mock_network(keypair: NetworkKeyPair, address: &Multiaddr) -> anemo::Netw
         .start(anemo::Router::new())
         .unwrap()
 }
+
+pub fn random_network() -> anemo::Network {
+    let network_key = NetworkKeyPair::generate(&mut OsRng).private().0.to_bytes();
```
aschran, line 924: can this be implemented in terms of `mock_network`?

Also: is it too late to pedantically rename `mock_network`? It's not really a mock, I'd just call it `test_network`

primary/src/tests/block_remover_tests.rs
```diff
@@ -459,21 +474,24 @@ async fn test_unlocking_pending_requests() {
...
     expected_batch_ids: Vec<BatchDigest>,
     tx_delete_batches: metered_channel::Sender<DeleteBatchResult>,
 ) -> JoinHandle<()> {
-    println!("[{}] Setting up server", &address);
-    let mut recv = PrimaryToWorkerMockServer::spawn(address.clone());
+    let pubkey = keypair.public().clone();
+    println!("[{}] Setting up server", pubkey);
+    let (mut recv, network) = PrimaryToWorkerMockServer::spawn(keypair, address);
     tokio::spawn(async move {
+        let _network = network;
```
aschran, line 486: What's this for? is it some kind of closure capture thing? worth explaining or is it idiomatic and I just don't know Rust well enough yet :)

bmwill: Yeah this is just so that the network handle is kept alive, since it will shutdown the network once all handles have been dropped.

test_utils/src/lib.rs
bmwill: yep we can do both

node/tests/reconfigure.rs
bmwill: yep

network/src/traits.rs
bmwill: Yeah the old traits were for grpc networks and now that we've finished the conversion we can just rename it.

node/src/restarter.rs
bmwill: yeah i'll add a comment

node/src/restarter.rs
bmwill: yeah i suppose we could, but idk if its strictly worth it, this will only be an issue the first time a connection is established and on subsequent iterations the connection will already have been established so this should result in a noop.

aschran (approved):

PR #979, opened 2022-09-14
Pass a host string to get_available_port.

mystenmark:
Previously, `get_available_port` was using `"localhost"`, which may resolve to the ipv6 loopback instead of ipv4. We then use the obtained port in an ipv4 multiaddr in every case that was calling `get_available_port()`

This was probably working 99.9% of the time before by chance, since there isn't much ephemeral port contention most of the time.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -145,8 +145,9 @@ pub struct PrometheusMetricsParameters {
 
 impl Default for PrometheusMetricsParameters {
     fn default() -> Self {
+        let host = "127.0.0.1";
         Self {
-            socket_addr: format!("/ip4/127.0.0.1/tcp/{}/http", get_available_port())
+            socket_addr: format!("/ip4/{}/tcp/{}/http", host, get_available_port(host))
                 .parse()
                 .unwrap(),
         }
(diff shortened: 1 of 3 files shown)


huitseeker (approved):
Indeed, if you bind on `::1` to probe and then bind on `127.0.0.1` for actual use, you haven't actually reserved a port. Thanks!

@mystenmark I note the sui-side variant of this change at https://github.com/MystenLabs/sui/pull/4525 did not excise `localhost` from sui's `crates/sui-config/src/utils.rs`'s `get_ephemeral_port`. It would probably be wise to remove it there as well, even if sui doesn't use the same network as NW yet.

mystenmark:
> Indeed, if you bind on `::1` to probe and then bind on `127.0.0.1` for actual use, you haven't actually reserved a port. Thanks!
> 
> @mystenmark I note the sui-side variant of this change at [MystenLabs/sui#4525](https://github.com/MystenLabs/sui/pull/4525) did not excise `localhost` from sui's `crates/sui-config/src/utils.rs`'s `get_ephemeral_port`. It would probably be wise to remove it there as well, even if sui doesn't use the same network as NW yet.

Thanks for the pointer - I know I'd seen something like that go by recently, but I couldn't figure out where it was. Will fix that as well.

bmwill:
We may need to rethink port reservation because we now use udp for a few interfaces and reserving a tcp port doesn't do anything to reserve a udp one. Also the same trick (the whole time_wait state with tcp) does not exist with udp unfortunately...

PR #980, opened 2022-09-14
Reduce timeout value for leader

asonnino:
In partial synchrony, if this node is going to be the leader of the next round, we set a lower timeout value to increase its chance of committing the leader committed.

### DIFF
--- a/primary/src/proposer.rs
+++ b/primary/src/proposer.rs
@@ -135,7 +135,23 @@ impl Proposer {
         self.last_parents = Certificate::genesis(&self.committee);
     }
 
-    // Main loop listening to incoming messages.
+    /// Compute the timeout value of the proposer.
+    fn timeout_value(&self) -> Instant {
+        match self.network_model {
+            // In partial synchrony, ff this node is going to be the leader of the next
+            // round, we set a lower timeout value to increase its chance of committing
+            // the leader committed.
+            NetworkModel::PartiallySynchronous
+                if self.committee.leader(self.round + 1) == self.name =>
+            {
+                Instant::now() + self.max_header_delay / 2
+            }
+
+            // Otherwise we keep the default timeout value.
+            _ => Instant::now() + self.max_header_delay,
+        }
+    }
+
     /// Update the last leader certificate. This is only relevant in partial synchrony.
     fn update_leader(&mut self) -> bool {
         let leader_name = self.committee.leader(self.round);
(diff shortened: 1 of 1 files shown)


primary/src/proposer.rs
```diff
@@ -135,7 +135,23 @@ impl Proposer {
...
+    /// Compute the timeout value of the proposer.
+    fn timeout_value(&self) -> Instant {
+        match self.network_model {
+            // In partial synchrony, if this node is going to be the leader of the next
+            // round, we set a lower timeout value to increase its chance of committing
+            // the leader committed.
+            NetworkModel::PartiallySynchronous
+                if self.committee.leader(self.round + 1) == self.name =>
+            {
+                Instant::now() + self.max_header_delay / 2
```
akichidis, line 147: Why `max_header_delay / 2` ? Is there some logic why we picked to half the time or this is a good enough estimate for what we want to achieve?

primary/src/proposer.rs
```diff
@@ -135,7 +135,23 @@ impl Proposer {
         self.last_parents = Certificate::genesis(&self.committee);
     }
 
-    // Main loop listening to incoming messages.
+    /// Compute the timeout value of the proposer.
+    fn timeout_value(&self) -> Instant {
```
akichidis, line 139: Can we somehow test the changes of this? Also what would be the way to see whether this actually makes an improvement? Can you think of some metric we could introduce that would show this?

akichidis (commented):
@asonnino if I understand correctly this change will try to make our node propose a header for the next round fast enough to make sure that it doesn't "miss its chance" to commit its own stuff? I might describe it quite vague, but want to make sure I understand the change here.

akichidis:
Is this still something we want to go forward with @asonnino ?

primary/src/proposer.rs
asonnino: No deep reason (and open to suggestions). I just wanted to set a lower time than other nodes.

asonnino:
> @asonnino if I understand correctly this change will try to make our node propose a header for the next round fast enough to make sure that it doesn't "miss its chance" to commit its own stuff? I might describe it quite vague, but want to make sure I understand the change here.

Yes you got it right

asonnino:
> Is this still something we want to go forward with @asonnino ?

Ha! I somehow missed your comments on this PR. Yes I think we can go ahead if you think it is a good idea.

PR #990, opened 2022-09-14
Refactor worker-to-worker RPC flows to be synchronous.

aschran:
This sets the stage for simplifying RPC call graph, retry logic, etc., when primary-to-worker RPCs are similarly refactored in future PR(s).


network/src/p2p.rs
```diff
@@ -296,3 +317,43 @@ impl ReliableNetwork<WorkerMessage> for P2pNetwork {
...
+impl ReliableNetwork<WorkerBatchRequest> for P2pNetwork {
+    type Response = WorkerBatchResponse;
+    async fn send(
+        &mut self,
+        peer: NetworkPublicKey,
+        message: &WorkerBatchRequest,
+    ) -> CancelOnDropHandler<anyhow::Result<anemo::Response<WorkerBatchResponse>>> {
+        let message = message.to_owned();
+        let f = move |peer| {
+            let message = message.clone();
```
mwtian, line 349: Why another clone() is needed?

network/src/p2p.rs
```diff
@@ -296,3 +317,43 @@ impl ReliableNetwork<WorkerMessage> for P2pNetwork {
...
+            WorkerToWorkerClient::new(peer)
+                .request_batches(message)
+                .await
+        };
+        self.unreliable_send(peer, f).await
+    }
+}
+
+#[async_trait]
+impl ReliableNetwork<WorkerBatchRequest> for P2pNetwork {
```
mwtian, line 340: Not for this PR, but the two definitions of unreliable and reliable networks feels like a bit of boilerplate. Is it possible to avoid defining the `send()` and `unreliable_send()` functions for other types of request and response?

And just curious, what is the difference in implementation between unreliable and reliable networks?

worker/src/worker.rs
```diff
@@ -332,21 +330,8 @@ impl Worker {
         &self,
         tx_reconfigure: &watch::Sender<ReconfigureNotification>,
         tx_primary: Sender<WorkerPrimaryMessage>,
-        rx_worker_request: types::metered_channel::Receiver<(Vec<BatchDigest>, PublicKey)>,
         rx_worker_processor: types::metered_channel::Receiver<Batch>,
-        network: anemo::Network,
     ) -> Vec<JoinHandle<()>> {
```
mwtian, line 334: Maybe we can just return `JoinHandle<()>` now.

mwtian (approved):
Looks great overall!

worker/src/worker.rs
```diff
@@ -457,16 +442,25 @@ impl WorkerToWorker for WorkerReceiverHandler {
...
         .map_err(|e| anemo::rpc::Status::internal(e.to_string()))
     }
+    async fn request_batches(
+        &self,
+        request: anemo::Request<types::WorkerBatchRequest>,
+    ) -> Result<anemo::Response<types::WorkerBatchResponse>, anemo::rpc::Status> {
+        let message = request.into_body();
+        let batches: Vec<Batch> = self
+            .store
+            .read_all(message.digests)
```
akichidis, line 456: I am thinking that maybe we should keep a TODO comment as before (`// TODO [issue #7]: Do some accounting to prevent bad actors from monopolizing our resources`) , but updated, as a byzantine node could send a huge list here and stress our node. Something to not forget about.

akichidis (commented):
This is great work @aschran 💯  and a pleasure to see us adopting a request/response pattern for such things. I am already thinking of other places where this would simplify things. Left a few comments. Could you also please run the benchmarks (apologies if you have done already) to ensure we don't see any significant change in performance - although I am not 100% sure the [current benchmark](https://github.com/MystenLabs/narwhal/tree/main/benchmark)s will reveal anything as this flow is not a frequently executed one unless we are under some kind of stress (ex network delays etc)

worker/src/worker.rs
```diff
@@ -457,16 +442,25 @@ impl WorkerToWorker for WorkerReceiverHandler {
...
+            .await
+            .map_err(|e| anemo::rpc::Status::from_error(Box::new(e)))?
+            .into_iter()
+            .flatten()
+            .collect();
+        Ok(anemo::Response::new(types::WorkerBatchResponse { batches }))
```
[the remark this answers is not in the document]
bmwill: If we expect batches to be large then it may be better to have this rpc only handle a single batch request and instead request a Vec of batches by initiating N requests.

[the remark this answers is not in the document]
aschran: I could implement a max request size of N batches, which would also help towards addressing the issue above? Can any of you advise on a good value for N? I don't have a sense of how big these are, either.

network/src/p2p.rs
aschran: I agree these are starting to look repetitive enough that a macro or something might be worthwhile (if it's not possible to do via a default impl in the trait itself). +1 for a future PR.

Essentially the reliable variant does a bunch of retrying in case of errors or missing peer connections, while unreliable does not.

network/src/p2p.rs
aschran: It's cloned so it can be moved into the async block. `send` needs to be able to call this closure multiple times in case of retry so it has to be `Fn` and not just `FnOnce`

worker/src/synchronizer.rs
```diff
@@ -309,14 +290,19 @@ impl Synchronizer {
...
-                        // We got the batch, remove it from the pending list.
-                        self.pending.remove(&digest);
-                    },
-                    Ok(None) => {
-                        // The sync request for this batch has been canceled.
+                    Ok(Ok(response)) => {
+                        for batch in response.into_body().batches {
+                            // TODO: remove duplicate hashing of batch after primary-to-worker
+                            // communication is refactored to be synchronous.
+                            self.pending.remove(&batch.digest());
```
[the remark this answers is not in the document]
aschran: Yes, good point - updated

worker/src/synchronizer.rs
```diff
@@ -206,10 +190,12 @@ impl Synchronizer {
                                 }
                             };
 
-                            debug!("Sending BatchRequest message to {} for missing batches {:?}", worker_name, missing.clone());
+                            debug!("Sending WorkerBatchRequest message to {} for missing batches {:?}", worker_name, missing.clone());
 
-                            let message = WorkerMessage::BatchRequest(missing.into_iter().collect::<Vec<_>>(), self.name.clone());
-                            self.network.unreliable_send(worker_name, &message).await;
+                            // TODO: restore ability to cancel these requests when primary->worker RPCs are
+                            // made synchronous and this one becomes a child of those.
```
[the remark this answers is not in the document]
aschran: Yes, there may be internal timeouts driven locally, but the e2e process will be driven by primary

For example here
1. Primary calls `Synchronize`
2. This triggers a `WorkerBatchRequest` for the requested digests
3. Internally to the `Synchronize` handler, we may still keep the `sync_retry_delay` timeout currently implemented by the timer in this file and execute a second child RPC after that delay

worker/src/worker.rs
aschran: I actually made that change originally and then put it back, just in case we add another task again here later. If we think that will never happen, then this doesn't even need to be a separate function. I can inline the one single call to `Processor::spawn` in `Worker::spawn` just as easily - wdyt?

network/src/p2p.rs
```diff
@@ -156,11 +169,12 @@ impl Lucky for P2pNetwork {
 
 #[async_trait]
 impl UnreliableNetwork<PrimaryMessage> for P2pNetwork {
+    type Response = ();
     async fn unreliable_send(
         &mut self,
         peer: NetworkPublicKey,
         message: &PrimaryMessage,
-    ) -> JoinHandle<()> {
+    ) -> JoinHandle<anyhow::Result<anemo::Response<()>>> {
```
huitseeker, line 177: Nit: I'd love some imports in there!

network/src/p2p.rs
```diff
@@ -296,3 +317,43 @@ impl ReliableNetwork<WorkerMessage> for P2pNetwork {
...
+        &mut self,
+        peer: NetworkPublicKey,
+        message: &WorkerBatchRequest,
+    ) -> CancelOnDropHandler<anyhow::Result<anemo::Response<WorkerBatchResponse>>> {
+        let message = message.to_owned();
+        let f = move |peer| {
+            let message = message.clone();
+            async move {
+                WorkerToWorkerClient::new(peer)
+                    .request_batches(message)
```
huitseeker, line 352: I'm a bit uncomfortable with `request_batches` being the method that sends a generic network message between workers. Could we do a bit of renaming?

worker/src/helper.rs
```diff
@@ -1,115 +0,0 @@
-// Copyright (c) 2021, Facebook, Inc. and its affiliates
```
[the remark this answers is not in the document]
aschran: > This is great work @aschran 💯 and a pleasure to see us adopting a request/response pattern for such things. I am already thinking of other places where this would simplify things. Left a few comments. Could you also please run the benchmarks (apologies if you have done already) to ensure we don't see any significant change in performance - although I am not 100% sure the [current benchmark](https://github.com/MystenLabs/narwhal/tree/main/benchmark)s will reveal anything as this flow is not a frequently executed one unless we are under some kind of stress (ex network delays etc)

Thanks for the reminder. So it looks like I broke something with the benchmark (or else caused a big performance problem)? Lol. I'm not sure how consensus TPS could drop that much but e2e TPS stay the same? Maybe a log line the parser wants is no longer emitted? I'm also confused about the negative latency reports? Which happen even on main branch, apparently.

I'll start digging.

Before:
```
 + RESULTS:
 Consensus TPS: 48,849 tx/s
 Consensus BPS: 25,010,596 B/s
 Consensus latency: -77,277 ms

 End-to-end TPS: 48,515 tx/s
 End-to-end BPS: 24,839,538 B/s
 End-to-end latency: -65,391 ms
```

After:
```
+ RESULTS:
 Consensus TPS: 38 tx/s
 Consensus BPS: 19,641 B/s
 Consensus latency: -126,861 ms

 End-to-end TPS: 48,922 tx/s
 End-to-end BPS: 25,047,886 B/s
 End-to-end latency: -259,855 ms
```

network/src/p2p.rs
huitseeker: Traits have default methods: we can request the implementer to do less in the required method of the traits, e.g.implement the part where they return a `BoxFuture`.

worker/src/helper.rs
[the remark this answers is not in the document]
aschran: OK, I'm now realizing these results are inconsistent in a way that doesn't seem related to this PR.

I ran it a few more times, and some runs with my changes look fine:
```
 + RESULTS:
 Consensus TPS: 48,696 tx/s
 Consensus BPS: 24,932,343 B/s
 Consensus latency: -51,812 ms

 End-to-end TPS: 48,308 tx/s
 End-to-end BPS: 24,733,859 B/s
 End-to-end latency: -49,343 ms
```

Meanwhile on unmodified `main`, I can reproduce the bad results with similar frequency, with failures due to `BlockWaiter` timeouts and consensus TPS numbers of 38tx/s.

Hopefully continuing on with this refactoring makes exactly that sort of issue easier to debug? Right now it's hard for me to tell the cause and if this change makes it worse or has no effect.

worker/src/tests/synchronizer_tests.rs
```diff
@@ -14,6 +14,7 @@ use tokio::time::timeout;
 async fn synchronize() {
```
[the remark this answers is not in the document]
aschran: Currently 1 and 2 are pretty much the same case, since errors to the `WorkerBatchRequest` are treated the same as timeouts/non-responses--the existing timer-based retry mechanism (which remains unchanged) activates and sends a `lucky_broadcast` request to a few other random nodes for the same batches.

Given that the retry mechanism will soon change in the next PR when I refactor the `Synchronize` call coming from the primary, would it be fine by you if I hold of adding this test until then? I think otherwise some of it will have to be immediately rewritten.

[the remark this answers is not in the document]
aschran: (Note that one benefit of the next step in refactoring is that we can immediately retry requests for batches that the initial target worker does not have, instead of having to wait for a timeout)

aschran:
Planning to merge this in a few hours. Fire away if there are any last comments, or I'm happy to address any after that point in a follow-up PR

worker/src/worker.rs
mwtian: I don't have strong opinion either way. Since this has less delta I think we can definitely keep the current way.

worker/src/worker.rs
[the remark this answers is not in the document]
mwtian: @huitseeker

mwtian:
Great work. Let's wait a bit to see if @huitseeker or other folks have remaining comments?

worker/src/worker.rs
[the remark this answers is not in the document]
aschran: I put a TODO for this next to the one about defending against malicious workers using excessive resources

mwtian:
Since we may simplify this interface soon, it seems worthwhile to merge this for now and keep iterating on it.

test_utils/src/lib.rs
```diff
@@ -314,10 +324,23 @@ impl WorkerToWorker for WorkerToWorkerMockServer {
...
+    ) -> Result<anemo::Response<WorkerBatchResponse>, anemo::rpc::Status> {
+        let message = request.into_body();
+
+        self.batch_request_sender.send(message).await.unwrap();
+
+        // For testing stub, just always reply with no batches.
+        Ok(anemo::Response::new(WorkerBatchResponse {
+            batches: vec![],
+        }))
+    }
```
bmwill, line 343: With this change to request/response pairs we'll probably end up needing to have more tests for the actual handlers themselves since they're now expected to actually respond with something.

types/build.rs
```diff
@@ -103,6 +103,15 @@ fn build_anemo_services(out_dir: &Path) {
                 .codec_path("anemo::rpc::codec::BincodeCodec")
```
bmwill, line 103: I can't comment on that line, but is the old send_message rpc un-needed after this change?

worker/src/worker.rs
```diff
@@ -332,21 +330,8 @@ impl Worker {
...
-        // The `Helper` is dedicated to reply to batch requests from other workers.
-        let helper_handle = Helper::spawn(
-            self.id,
-            (*(*(*self.committee).load()).clone()).clone(),
-            self.worker_cache.clone(),
-            self.store.clone(),
-            tx_reconfigure.subscribe(),
-            rx_worker_request,
-            P2pNetwork::new(network),
-        );
```
bmwill, line 348: Awesome that we're able to completely remove this extra task

worker/src/worker.rs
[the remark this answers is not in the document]
bmwill: Do we have an idea of the upper-limit on size of a batch in terms of bytes?

bmwill (approved):
LGTM! I have a couple of clarifying questions but nothing that would block this landing.

test_utils/src/lib.rs
aschran: Agreed on more tests for the handlers. I think a good pattern for the mock servers is also TBD. Do you have thoughts on this? Curious an interface similar to gMock extensions for gRPC is possible here - is there any Rust mocking library we use?

For what test client code looks like in gRPC/C++, see the bottom of
https://grpc.github.io/grpc/cpp/md_doc_unit_testing.html

It's very intuitive IMO

aschran:
Can't figure out how to reply to your batch size comment. FWIW I think it's set to 500K in the benchmarks, though I'm not sure if that is the same in prod

PR #993, opened 2022-09-15
fix: Multiplex messages to the BlockSynchronizer on a single channel

huitseeker:
Today, the BlockSynchronizer receives several types of messages from the PrimaryReceiverHandler, which go on a different channel depending on their nature. This does not scale, especially since we are about to add more messages types traveling the same component edge in #848.

We put the messages under an enum and transmit that enum along the channel, which simplified BlockSynchronizer setup and channel patterns overall.

### DIFF
--- a/primary/src/block_synchronizer/mod.rs
+++ b/primary/src/block_synchronizer/mod.rs
@@ -40,6 +40,8 @@ use types::{
     ReconfigureNotification,
 };
 
+use self::responses::AvailabilityResponse;
+
 #[cfg(test)]
 #[path = "tests/block_synchronizer_tests.rs"]
 mod block_synchronizer_tests;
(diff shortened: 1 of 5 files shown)


mwtian (approved):
Thanks a lot @huitseeker! I will rebase #848 on this. Probably I should have done this inside #848 ...

akichidis (approved):
Totally makes sense, thanks @huitseeker for reducing this! My small concern is that if we experience a big influx of messages of one type (ex the `AvailabilityResponse::Payload`) it will force the component consume all those messages before any other one and let the protocol advance. Having 2 separate channels would do some load balancing here (as branches would be picked at random + readiness from tokio select) . @huitseeker if you think the above concern is not valid (or far fetched) feel free to ignore.

mwtian:
Sorry didn't see @akichidis 's reply earlier. I hope we can reduce the places where we have to consider backpressure / blockage soon, e.g. by pushing these concerns to the "top level" request handler and network layer.

huitseeker:
@akichidis I think load balancing works great once we are sure the processing of messages is not blocking :)

PR #1004, opened 2022-09-16
fix: improve the span arguments at the primary

huitseeker:
This will allow a better span analysis (especially the UIDs in the core)


velvia (approved):
You beat me to it, here is my version where I added the first ID when the API calls for a Vec....  

https://github.com/MystenLabs/narwhal/compare/main...velvia:narwhal:ec/decorate-spans-more

PR #1005, opened 2022-09-16
[subscriber] convert to sequential approach

akichidis:
Given the recent issues we've seen I am converting the Subscriber to work in a sequential / head of line blocking approach. No attempt will be made to download the payload of the next certificate, unless we have finished downloading the preceding one. The assumption is that we are willing to sacrifice part of our performance for a more straightforward approach that will prevent us from monopolizing our resources especially in cases of stress.

The PR is also refactoring the recovery method to follow the same approach - comments have been added with the decisions. Also metrics have been introduced for better visibility.

### DIFF
--- a/executor/src/metrics.rs
+++ b/executor/src/metrics.rs
@@ -1,13 +1,19 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
-use prometheus::{default_registry, register_int_gauge_with_registry, IntGauge, Registry};
+use prometheus::{
+    default_registry, register_histogram_with_registry, register_int_counter_with_registry,
+    register_int_gauge_with_registry, Histogram, IntCounter, IntGauge, Registry,
+};
 
 #[derive(Clone, Debug)]
 pub struct ExecutorMetrics {
     /// occupancy of the channel from the `Subscriber` to `Core`
     pub tx_executor: IntGauge,
-    /// Number of elements in the waiting (ready-to-deliver) list of subscriber
-    pub waiting_elements_subscriber: IntGauge,
+    /// Time it takes to download a payload on the Subscriber
+    pub subscriber_download_payload_latency: Histogram,
+    /// The number of certificates processed by Subscriber
+    /// during the recovery period to fetch their payloads.
+    pub subscriber_recovered_certificates_count: IntCounter,
 }
 
 impl ExecutorMetrics {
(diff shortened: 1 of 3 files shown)


executor/src/subscriber.rs
```diff
@@ -157,17 +187,29 @@ impl Subscriber {
     /// sequenced we will not quit this method until we have successfully
     /// fetched the payload.
     async fn wait_on_payload(
```
mwtian, line 189: It seems possible that parents of `certificate_id` may not exist in the certificate store, and this function triggers block waiter to fetch the parent certificates and the certificate itself as well, before fetching their payloads. If my understanding is correct, maybe we can update the function comment?

mwtian (commented):
Definitely worth a try to see if this stablize things. Hope we can have a way to reproduce issues in integration or unit test. For future work, do you think it will be worthwhile to make fetching certificates sequential, but allow parallelism for fetching payloads?

executor/src/subscriber.rs
asonnino: I don't think this is possible. Consensus would fail if we don't have all the ancestors of a certificate that we try to sequence. The role of the `certificate-waiter` is exactly to prevent that.

asonnino (approved):

executor/src/metrics.rs
```diff
@@ -19,12 +28,30 @@ impl ExecutorMetrics {
...
+                ],
                 registry
             )
             .unwrap(),
+            subscriber_recovered_certificates_count: register_int_counter_with_registry!(
+                "subscriber_recovered_certificates_count",
+                "The number of certificates processed by Subscriber during the recovery period to fetch their payloads",
+                registry
+            ).unwrap(),
+            subscriber_download_payload_attempts: register_histogram_with_registry!(
```
huitseeker, line 47: Nit: the name of the metric is `_retries` (i.e. attempts of order >1) and the documentation is attempts.

huitseeker (approved):
Thanks a lot @akichidis

executor/src/subscriber.rs
mwtian: I see. If the `GetBlock` message here never triggers causal completion (because the parents must already be in the certificate store), which logic triggers the causal completion?

mwtian: In `certificate waiter` I only see logic to wait for parents to become available, but not the logic to trigger fetching parents. Maybe I'm missing something ..

mwtian (approved):

huitseeker:
Actually @akichidis I think before merging this we should run benchmarks locally at least for 20, 40 and 60s

executor/src/subscriber.rs
akichidis: @mwtian if this is still unclear to you let's catch up offline, I can explain the process

mwtian: Great, thanks! We can discuss offline.

huitseeker:
Screen Shot 2022-09-24 at 12 21 18 PM

PR #1019, opened 2022-09-20
Fix `test_restore_from_disk`

mwtian:
After #976, it seems stopping a primary via test `Cluster` can take >= 2s to release the primary's port. If restarting the primary within 2s, there will be a port conflict.

I'm not sure if the delayed release of port is an expected behavior. For now, just mitigate the test failure with a longer wait (same as the wait in `test_read_causal_signed_certificates` below) between restarting the primary.


bmwill (approved):
Yeah the way the network is gracefully shutdown is by dropping all of the handles to it.

PR #1024, opened 2022-09-23
network: don't block on unreliable_send

bmwill:
Re-introduce the fix from #940 that was unintentionally reverted while migrating to anemo.

### DIFF
--- a/primary/src/block_remover.rs
+++ b/primary/src/block_remover.rs
@@ -557,7 +557,8 @@ impl BlockRemover {
             let message = PrimaryWorkerMessage::DeleteBatches(batch_ids.clone());
 
             // send the request
-            self.worker_network
+            let _ = self
+                .worker_network
                 .unreliable_send(worker_name.clone(), &message)
                 .await;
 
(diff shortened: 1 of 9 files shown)


primary/src/block_remover.rs
```diff
@@ -557,9 +557,9 @@ impl BlockRemover {
             let message = PrimaryWorkerMessage::DeleteBatches(batch_ids.clone());
 
             // send the request
-            self.worker_network
-                .unreliable_send(worker_name.clone(), &message)
-                .await;
+            let _ = self
```
aschran, line 560: what does the `let _ =` do? Asking mostly for my own edification :)

aschran (approved):

primary/src/block_remover.rs
bmwill: It immediately drops the response and in this case silences the warning well now get for ignoring a Result type (since most of these call sites don't presently care about the result anyway)

aschran: Cool makes sense! Didn't realize that without that it'd emit a warning

bmwill: you can annotate types with `#[must_use]` to have the compiler throw a warning if a type isn't used. its super helpful generally.

PR #1025, opened 2022-09-23
Add `narwhal-` prefix to narwhal package names

### LINKED ISSUE #757: [crates] rename Narwhal crates
We should rename the Narwhal crates to use the `narwhal_` prefix, which will allow fuzzy matching of all the Narwhal crates when tuning log levels.

mwtian:
Closes #757

This makes filtering by narwhal package names easier. Also if we place Narwhal packages along with other packages, we would like to have non-ambiguous package and matching directory names.

Adding `narwhal-` instead of `narwhal_` prefix to match the style in Sui.


primary/tests/integration_tests_validator_api.rs
```diff
@@ -1,11 +1,12 @@
-use arc_swap::ArcSwap;
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
+use arc_swap::ArcSwap;
```
kchalkias, line 3: Why do we need it in this PR?

kchalkias (approved):
Sounds reasonable to me. Thanks, and please ensure we update Sui accordingly if this is a breaking change.

consensus/Cargo.toml
```diff
@@ -19,13 +19,13 @@ thiserror = "1.0.35"
...
 fastcrypto = "0.1.2"
-crypto = { path = "../crypto" }
-storage = { path = "../storage" }
-dag = { path = "../dag" }
+crypto = { path = "../crypto", package = "narwhal-crypto" }
+storage = { path = "../storage", package = "narwhal-storage" }
+dag = { path = "../dag", package = "narwhal-dag" }
 prometheus = "0.13.2"
-types = { path = "../types" }
+types = { path = "../types", package = "narwhal-types" }
```
bmwill, line 28: my preference would be to probably import the crates with the name being "narwhal-*" instead of renaming them to "*" as we're doing in this case but i suppose that would result in quite a bit more churn atm

mwtian: Yeah the churn will be larger if we do this additional renaming. It can be a separate change in future.

mwtian:
> Sounds reasonable to me. Thanks, and please ensure we update Sui accordingly if this is a breaking change.

Yes Sui will be updated.

akichidis (commented):
Thanks @mwtian for this!

PR #1031, opened 2022-09-27
Support range synchronization of missing certificates

mwtian:
Part 2/n of MystenLabs/sui#5268

This PR continues from MystenLabs/narwhal#848, by adding the request handler for range sync, and the logic to kick start range sync when more than 2 rounds of certificates are missing. This number of missing rounds threshold is a constant.

Pseudo code for whether to start range sync:
```
for each certificate received by primary
    if there are too many missing rounds between local store and the received certificate
        kick start the range sync process, without blocking additional processing at primary
```

Range sync logic is now placed into a few different components:
- `Synchronizer`: checks if range sync should be started, and sends a message to block waiter.
- `BlockWaiter`: runs range sync without blocking other operations.
- `BlockSynchronizerHandler`: synchronize certificate digests in a range of rounds, then synchronize certificates for the received digests.
- `BlockSynchronizer`: (implemented in MystenLabs/narwhal#848) drives the protocol for getting certificate digests in a range of rounds
- `Helper`: handles range sync requests, fetches and returns certificate digests.

For now, range synchronization is disabled. It will be enabled in `process_certificate()` later once we finalize the plan to refactor synchronization logic.

### DIFF
--- a/primary/src/block_synchronizer/handler.rs
+++ b/primary/src/block_synchronizer/handler.rs
@@ -5,14 +5,16 @@ use crate::{
         handler::Error::{BlockDeliveryTimeout, BlockNotFound, Internal, PayloadSyncError},
         BlockSynchronizeResult, Command, SyncError,
     },
+    metrics::PrimaryMetrics,
     BlockHeader,
 };
 use async_trait::async_trait;
+use config::Committee;
 use fastcrypto::Hash;
 use futures::future::join_all;
 #[cfg(test)]
 use mockall::*;
-use std::time::Duration;
+use std::{sync::Arc, time::Duration};
 use storage::CertificateStore;
 use thiserror::Error;
 use tokio::{sync::mpsc::channel, time::timeout};
@@ -66,6 +68,10 @@ impl Error {
 #[cfg_attr(test, automock)]
 #[async_trait]
 pub trait Handler {
+    /// Synchronizes missing certificates via the block_synchronizer, from the next round
+    /// where the local certificate store does not have a certificate.
+    async fn synchronize_range(&self) -> Result<(), Error>;
+
     /// It retrieves the requested blocks via the block_synchronizer making
     /// sure though that they are fully validated. The certificates will only
     /// be returned when they have properly processed via the core module
@@ -109,6 +115,12 @@ pub struct BlockSynchronizerHandler {
     /// The timeout while waiting for a certificate to become available
     /// after submitting for processing to core.
     certificate_deliver_timeout: Duration,
+
+    /// Info about the validator committee.
+    committee: Committee,
+
+    /// Exported metrics.
+    metrics: Arc<PrimaryMetrics>,
 }
 
 impl BlockSynchronizerHandler {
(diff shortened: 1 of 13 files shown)


primary/src/block_synchronizer/handler.rs
```diff
@@ -151,6 +167,40 @@ impl BlockSynchronizerHandler {
 
 #[async_trait]
 impl Handler for BlockSynchronizerHandler {
+    /// This method kicks start a process to fetch digests of missing certificates after the latest local round.
```
asonnino, line 170:
```suggestion
    /// This method kick starts a process to fetch digests of missing certificates after the latest local round.
```

primary/src/block_synchronizer/handler.rs
```diff
@@ -151,6 +167,40 @@ impl BlockSynchronizerHandler {
...
+            .with_label_values(&[&self.committee.epoch.to_string()])
+            .inc();
+
+        let (tx_certificate_ids, mut rx_certificate_ids) = channel(1);
+        self.tx_block_synchronizer
+            .send(Command::SynchronizeRange {
+                respond_to: tx_certificate_ids,
+            })
+            .await
+            .expect("Couldn't send SynchronizeRange message to block synchronizer");
```
asonnino, line 187: Are we sure this won't panic upon shutdown or restart?

primary/src/block_synchronizer/handler.rs
```diff
@@ -151,6 +167,40 @@ impl BlockSynchronizerHandler {
...
+        let certificate_ids = rx_certificate_ids.recv().await.unwrap_or_default();
```
asonnino, line 188: What is the default case here? I am referring to `unwrap_or_default()`. Simply an empty list right?

primary/src/helper.rs
```diff
@@ -86,6 +87,11 @@ impl Helper {
                     PrimaryMessage::CertificatesRequest(digests, origin) => {
                         let _ = self.process_certificates(digests, origin, false).await;
                     }
+                    // The CertificatesRangeRequest will collect certificate digests from range_start,
+                    // for at most max_rounds number of rounds. All digests will be returned in one response.
+                    PrimaryMessage::CertificatesRangeRequest {range_start, max_rounds, requestor: origin} => {
+                        let _ = self.handle_certificates_range_request(range_start, max_rounds, origin).await;
```
asonnino, line 93: I know you are following the executing pattern, but why is it ok to ignore the result of this function?

asonnino: Perhaps we want to log it in case it is a store error?

primary/src/synchronizer.rs
```diff
@@ -164,6 +173,23 @@ impl Synchronizer {
...
+        if round.saturating_sub(latest_round) <= MISSING_ROUNDS_CATCH_UP_THRESHOLD {
+            // No catch up is needed.
+            return false;
+        }
+        let cmd = BlockCommand::GetRange {};
+        self.tx_get_block_commands
+            .send(cmd)
+            .await
+            .expect("Failed to send get range request");
+        info!("Missing rounds detected: received certificate at {}, local latest at {}. Starting to catch up ...", round, latest_round);
```
asonnino, line 189: are you sure this log won't spam us?

asonnino (approved):

mwtian (commented):
Thanks for taking a look!

mwtian:
Closing for the new iteration at https://github.com/MystenLabs/sui/pull/5058

PR #1033, opened 2022-09-28
Increase max task concurrency to 500K

mwtian:
Before we make a decision on whether and how to remove the `BoundedExecutor`, it seems useful to increase the max concurrency limit significantly so we are not running into it as often. Blocking on `BoundedExecutor` is potentially one source of deadlock we have observed.

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -52,7 +52,7 @@ impl<T> std::future::Future for CancelOnDropHandler<T> {
 //
 // The exact number here probably isn't important, the key things is that it should be finite so
 // that we don't create unbounded numbers of tasks.
-pub const MAX_TASK_CONCURRENCY: usize = 500;
+pub const MAX_TASK_CONCURRENCY: usize = 500_000;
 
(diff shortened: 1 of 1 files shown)


arun-koshy (approved):

akichidis:
@mwtian The reason why we introduced the BoundedExecutor in the past was because we were seeing some resource exhaustion in our nodes and the other nodes as well from which we were requesting entities. This was particularly visible when the `header_waiter` was asking stuff from the other nodes . Other correction measures were taken as well (ex make sure `header_waiter` is doing more proper backoff in retries) but BoundedExecutor was an action point as well. Unless there is something else downstream (ex via Anemo) which can limit the concurrency, a high threshold here might also get us into similar issues. Let's keep that under our radar.

mwtian:
I will definitely keep an eye out for resource usage spikes. Also I'm not sure if adding outgoing RPC limit is the most safe thing to do. Since this limit does not differentiate between destinations or methods, it is easy to envision a case where hitting the overall limit leads to deadlock.

PR #1034, opened 2022-09-28
Rework executor

andll:
This pull requests simplifies narwhal executor - it changes both how executor works internally and how narwhal executor interacts with consumer.

### Changes to sui interface:
* Pass transaction as `Vec<u8>`. This pushes serialization-deserialization fully to consumer. Previously serialization was done on consumer and de-serialization was done in narwhal, which does not make sense really
* Remove _executor->consumer_ channel. Currently both channel and callback existed as part of narwhal executor interface, this is redundancy that is not needed and callback can support execution channel if needed.
* Do not return `Outcome` from callback - this is not used once channel above is removed
* Do not return error from `load_execution_indices` - there is nothing to do with this error on nw side except to panic, so we can as well panic inside sui handler
* Do not return error from `handle_consensus_transaction`- entirety of error handling can be done in callback.
* Remove `ask|release_consensus_write_lock`  - this can be simply replaced by rust owner semantics

### How new executor works:
New executor only have 2 simple tasks.
* **Subscriber** task:
   * Download payload in memory(see below)
      * Buffer up to N downloads in parallel via FuturesOrdered
      * Send downloaded batch to notifier task
      * Propagate back-pressure to consensus if more then N downloads are pending
* **Notifier** task:
   * For each payload, execute consumer callback
   * Propagate back pressure from callback to subscriber

### Payload download 
is done in following manner:
* Local worker is checked first, in case payload is present locally
* If not found in local worker, start issuing parallel requests to 2f+1 remote workers signed the certificate
* Each request starts with staggering delay and performs infinite retries with exponential backoff
* Staggering delay allows not to have too many requests in case of good network conditions
* Exponential backoff allows not to spam too many request - after ramp up we don't issue more then 1 request per minute to remote worker (per download, so multiply by N max pending downloads for total)
* When one of requests succeeds all pending requests are cancelled

To recap, payload download is done as following formula:

```
Fetch: Select[exponential_backoff_with_stagger(get_payload_rpc()) for each authority signed certificate]
```

### Restartability
New executor only uses certificate index in order to support restart. Consumer must be idempotent in handling repeated transactions if execution was aborted in the middle of the certificate/payload.

### Main differences
between new and old executor

* New executor does not issue availability network requests to remote authorities, leveraging certificate signatures instead
* No temporary storage, we only hold up to N batches in memory after download. This reduces IO and complexity
* Not using batch_(loader|synchronizer) and uses simple network requests instead. No nested network calls and nested timeouts.

### Bounded resource usage
* Total number of pending downloads is limited by `MAX_PENDING_PAYLOADS`
* Total number of batches in memory is limited by `subscriber->notifier channel size` + `MAX_PENDING_PAYLOADS`
* Total number of pending RPC requests from this component: `MAX_PENDING_PAYLOADS`*`(2f+1)`
* Total number of pending RPC requests to a given node: `MAX_PENDING_PAYLOADS`

### TODO
 - [x] Rename GetPayload -> BatchRequest
 - [x] Rollback log level change
 - [x] Tests
 - [x] Randomize authorities
 - [x] Verify payload match digest
 - [x] Metrics
 - [x] Tracing


node/src/lib.rs
```diff
@@ -9,25 +9,21 @@ use consensus::{
 };
```
mwtian, line 9: Are we able to delete `temp_batch_store`, `temp_batch_map` and `TEMP_BATCH_CF` from this file?

executor/src/subscriber.rs
```diff
@@ -120,144 +153,155 @@ impl Subscriber {
...
-                attempts_count += 1;
-                attempts_count
-            };
-
-            let span = debug_span!("get_block", attempt = attempts);
-            async move {
-                let (sender, receiver) = oneshot::channel();
-
-                tx_get_block
-                    .send(BlockCommand::GetBlock {
```
mwtian, line 217: I think this was the only usage of `BlockCommand::GetBlock`. We can add a comment above `BlockCommand::GetBlock` to mention that it is now unused and see if we can delete unnecessary code.

mwtian (approved):
Great cleanup!

executor/src/lib.rs
```diff
@@ -45,122 +38,86 @@ pub type SerializedTransactionDigest = u64;
...
 pub struct Executor;
 
 impl Executor {
     /// Spawn a new client subscriber.
-    pub async fn spawn<State>(
-        store: Store<(CertificateDigest, BatchDigest), Batch>,
-        execution_state: Arc<State>,
+    pub fn spawn<State>(
+        name: PublicKey,
+        network: oneshot::Receiver<P2pNetwork>,
```
bmwill, line 65: Why isn't the network able to be directly passed in and instead is a oneshot?

bmwill: Ah, ok I see. The executor is started outside the primary and the network is constructed inside the primary which makes the dependency difficult to have at the time of construction.

executor/src/subscriber.rs
```diff
@@ -120,144 +153,155 @@ impl Subscriber {
...
+        worker_id: WorkerId,
+        authorities: Vec<PublicKey>,
+    ) -> Batch {
+        if let Some(payload) = self.try_fetch_locally(digest, worker_id).await {
+            return payload;
+        }
+        let _timer = self.metrics.subscriber_remote_fetch_latency.start_timer();
+        let mut stagger = Duration::from_secs(0);
+        let mut futures = vec![];
+        for authority in &authorities {
```
mwtian, line 209: Do we need to randomize the order of authorities to avoid overloading the authority that appears first? Also the order of authority to query may be relevant depending on how we handle response from byzantine workers.

executor/src/subscriber.rs
```diff
@@ -120,144 +153,155 @@ impl Subscriber {
...
+        let mut timeout = Duration::from_secs(10);
+        loop {
+            let deadline = Instant::now() + timeout;
+            let get_payload_guard =
+                PendingGuard::make_inc(&self.metrics.pending_remote_get_payload);
+            let payload =
+                tokio::time::timeout_at(deadline, self.network.get_payload(&worker, digest)).await;
+            drop(get_payload_guard);
+            match payload {
+                Ok(Ok(Some(payload))) => return payload,
```
mwtian, line 273: Should we verify the payload actually corresponds to the digest here? I'm not sure if this verification is done at a higher or lower level.

executor/src/subscriber.rs
```diff
@@ -120,144 +153,155 @@ impl Subscriber {
...
+                Ok(Ok(None)) => error!("[Protocol violation] Payload {} was not found at worker {} while authority {} signed certificate", digest, worker, authority),
```
mwtian, line 274: Maybe we don't need to keep retrying on a known byzantine worker, and just return and add a loop to the higher level `select_all()`.

mwtian (commented):
Btw, if we continue on this strategy, maybe we can optimize / simplify [this](https://github.com/mwtian/narwhal/blob/40f33bdc251b1e81b0194a559fa0a54ad1bc64dc/primary/src/header_waiter.rs#L177-L218) as a follow up.

andll:
Updated - added tests, randomizing authorities, verifying response digest

executor/src/subscriber.rs
andll: I think it is ok to try again - maybe there was some deployment issue on the validator and it got fixed. We have exponential backoff in any case, so it won't be too many retries

executor/src/subscriber.rs
andll: I will take a look if block waiter can be removed after this PR (also will need to sync with @akichidis about it)

node/src/lib.rs
andll: I'll take a look at that, I think right now there is still some code in block synchronizer that might use it(but yes, we should delete it)

executor/src/lib.rs
```diff
@@ -183,3 +141,21 @@ where
...
+    async fn handle_consensus_transaction(
+        &self,
+        consensus_output: &ConsensusOutput,
+        execution_indices: ExecutionIndices,
+        transaction: Vec<u8>,
+    ) {
+        self.as_ref()
+            .handle_consensus_transaction(consensus_output, execution_indices, transaction)
+            .await
+    }
```
akichidis, line 156: @andll out of curiosity why is this necessary? Am I getting it right that is mainly for convenience reasons or am I missing something?

executor/src/state.rs
```diff
@@ -3,10 +3,6 @@
 use serde::{Deserialize, Serialize};
 use types::SequenceNumber;
 
-#[cfg(test)]
-#[path = "tests/state_tests.rs"]
-pub mod state_tests;
-
```
akichidis, line 9: Since you removed the tests I would recommend to remove the implementation methods of `ExecutionIndices` as well since they don't seem to be used anymore.

akichidis (commented):
Great work @andll !! Left some comments. I'll wait for the rest of the tests your are working on - but that looks good!

executor/src/subscriber.rs
```diff
@@ -120,144 +156,313 @@ impl Subscriber {
...
+            return payload;
+        }
+        let _timer = self.metrics.subscriber_remote_fetch_latency.start_timer();
+        let mut stagger = Duration::from_secs(0);
+        let mut futures = vec![];
+        for worker in workers {
+            let future = self.fetch_from_worker(stagger, worker, digest);
+            futures.push(future.boxed());
+            stagger += Duration::from_secs(1);
+        }
```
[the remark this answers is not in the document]
akichidis: I like the staggering though, great approach!

executor/src/lib.rs
andll: Yeah, idea is that some places will use Arc<X> as executor state, and some(perhaps unit tests) might want to pass type without Arc

andll:
Re 'why warn vs debug' - there is separate commit that changes bunch of debug to warn, i'll revert it before merging once we have debug logs enabled for executor in private testnet

andll:
@akichidis I think I addressed rest of your comments

executor/src/subscriber.rs
```diff
@@ -120,144 +156,313 @@ impl Subscriber {
...
+                tokio::time::timeout_at(deadline, self.safe_request_batch(digest, &worker)).await;
+            drop(request_batch_guard);
+            match payload {
+                Ok(Ok(Some(payload))) => return payload,
+                Ok(Ok(None)) => error!("[Protocol violation] Payload {} was not found at worker {} while authority signed certificate", digest, worker),
+                Ok(Err(err)) => debug!(
+                    "Error retrieving payload {} from {}: {}",
+                    digest, worker, err
+                ),
+                Err(_elapsed) => debug!("Timeout retrieving payload {} from {}",
```
asonnino, line 268: If this happens it seems we have a serious problem, don't we? Should we log a warn / error?

akichidis:
> @akichidis I think I addressed rest of your comments, but let me know if something is missing

Great @andll, reviewing now

executor/src/subscriber.rs
```diff
@@ -1,112 +1,148 @@
...
-                store,
-                rx_reconfigure,
-                rx_consensus,
-                tx_executor,
-                tx_get_block_commands,
-                get_block_retry_policy,
-                metrics,
-            }
+#[must_use]
+pub fn spawn_subscriber(
```
akichidis, line 57: @andll are you going to add any test our the subscriber it self? I think it would be great if we could have some unit test coverage for that as well to ensure we don't accidentally break the behaviour

executor/src/subscriber.rs
akichidis: Well I guess not necessarily - workers of other validators might not respond in time (for whatever reasons)

executor/src/subscriber.rs
```diff
@@ -120,144 +156,313 @@ impl Subscriber {
...
+        }
+        let (batch, _, _) = futures::future::select_all(futures).await;
+        batch
+    }
+
+    #[instrument(level = "debug", skip_all, fields(digest = % digest, worker_id = % worker_id))]
+    async fn try_fetch_locally(&self, digest: BatchDigest, worker_id: WorkerId) -> Option<Batch> {
+        let _timer = self.metrics.subscriber_local_fetch_latency.start_timer();
+        let worker = self.network.my_worker(&worker_id);
+        let payload = self.network.request_batch(digest, &worker).await;
```
akichidis, line 228: Out of curiosity, will this eventually timeout at some point if no response is received? Not sure how Anemo handles requests (maybe @bmwill can provide some pointers?)

akichidis (approved):
Thanks for the changes @andll ! This PR looks great 💯  

I am approving to make sure is not getting blocked, but I would highly suggest to increase the test coverage in `Subscriber` & the `Notifier`. Right now there are methods & scenarios (ex retries) which are not covered and I would feel safer if we had something in place (especially to ensure that we don't break things as the codebase evolves)

executor/src/subscriber.rs
bmwill: If the connection is broken then yes it'll eventually timeout, otherwise if you need a particular timeout for this rpc you'll want to wrap it in tokio::time::timeout().

andll:
@akichidis @asonnino - I'll merge this PR to unblock nw<->sui merge. Will reply to your comments and/or make follow up PR after that

executor/src/subscriber.rs
andll: I don't have strong opinion on log level here - do you know how we typically log timeouts in other components?

executor/src/subscriber.rs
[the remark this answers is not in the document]
andll: So in ideal scenario we want to issue one RPC, and if node is healthy get result and move on (and if staggering step is chosen well this is what will happen with current approach)
In less ideal scenario we do want to have some overlapping attempts, because waiting exponential backoff before trying next node will be sub-optimal.

I think overall what you suggesting is very similar to how staggering delay and exponential back off actually end up working (except that we will have some overlapping requests).

PR #1035, opened 2022-09-28
Add primary->worker Synchronize RPC

aschran:
Moves implementation to a linear async function in the handler. Gets rid of all parts of the synchronizer that previously existed only to track pending batch requests to other workers.

NOTE: this changes retries of worker->worker block requests to happen only once instead of perpetually. Additional retries beyond that point will have to be driven by the primary.

### DIFF
--- a/network/src/p2p.rs
+++ b/network/src/p2p.rs
@@ -16,7 +16,7 @@ use tokio::{runtime::Handle, task::JoinHandle};
 use types::{
     PrimaryMessage, PrimaryToPrimaryClient, PrimaryToWorkerClient, PrimaryWorkerMessage,
     WorkerBatchRequest, WorkerBatchResponse, WorkerMessage, WorkerPrimaryMessage,
-    WorkerToPrimaryClient, WorkerToWorkerClient,
+    WorkerSynchronizeMessage, WorkerToPrimaryClient, WorkerToWorkerClient,
 };
 
 fn default_executor() -> BoundedExecutor {
@@ -237,6 +237,20 @@ impl ReliableNetwork<PrimaryWorkerMessage> for P2pNetwork {
         };
 
         self.send(peer, f).await
+    }
+}
+
+impl UnreliableNetwork<WorkerSynchronizeMessage> for P2pNetwork {
+    type Response = ();
+    fn unreliable_send(
+        &mut self,
+        peer: NetworkPublicKey,
+        message: &WorkerSynchronizeMessage,
+    ) -> Result<JoinHandle<Result<anemo::Response<()>>>> {
+        let message = message.to_owned();
+        let f =
+            move |peer| async move { PrimaryToWorkerClient::new(peer).synchronize(message).await };
+        self.unreliable_send(peer, f)
     }
 }
 
--- a/primary/src/block_synchronizer/mod.rs
+++ b/primary/src/block_synchronizer/mod.rs
@@ -36,8 +36,8 @@ use tokio::{
 };
 use tracing::{debug, error, info, instrument, trace, warn};
 use types::{
-    metered_channel, BatchDigest, Certificate, CertificateDigest, PrimaryWorkerMessage,
-    ReconfigureNotification, Round,
+    metered_channel, BatchDigest, Certificate, CertificateDigest, ReconfigureNotification, Round,
+    WorkerSynchronizeMessage,
 };
 
 use self::responses::{AvailabilityResponse, CertificateDigestsResponse};
@@ -896,8 +896,10 @@ impl BlockSynchronizer {
                 .expect("Worker id not found")
                 .name;
 
-            let message =
-                PrimaryWorkerMessage::Synchronize(batch_ids.clone(), primary_peer_name.clone());
+            let message = WorkerSynchronizeMessage {
+                digests: batch_ids.clone(),
+                target: primary_peer_name.clone(),
+            };
             let _ = self.network.unreliable_send(worker_name, &message);
 
             debug!(
--- a/primary/src/block_synchronizer/mod.rs
+++ b/primary/src/block_synchronizer/mod.rs
@@ -897,14 +897,14 @@ impl BlockSynchronizer {
                 .name;
 
             let message = WorkerSynchronizeMessage {
-                digests: batch_ids.clone(),
+                digests: batch_ids,
                 target: primary_peer_name.clone(),
             };
             let _ = self.network.unreliable_send(worker_name, &message);
 
             debug!(
                 "Sent request for batch ids {:?} to worker id {}",
-                batch_ids, worker_id
+                message.digests, worker_id
             );
         }
     }
--- a/primary/src/header_waiter.rs
+++ b/primary/src/header_waiter.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 use crate::{
     metrics::PrimaryMetrics,
-    primary::{PayloadToken, PrimaryMessage, PrimaryWorkerMessage},
+    primary::{PayloadToken, PrimaryMessage},
 };
 use config::{Committee, SharedWorkerCache, WorkerId};
 use crypto::PublicKey;
@@ -28,7 +28,7 @@ use types::{
     error::{DagError, DagResult},
     metered_channel::{Receiver, Sender},
     try_fut_and_permit, BatchDigest, CertificateDigest, Header, HeaderDigest,
-    ReconfigureNotification, Round,
+    ReconfigureNotification, Round, WorkerSynchronizeMessage,
 };
 
 #[cfg(test)]
@@ -231,7 +231,7 @@ impl HeaderWaiter {
                                     .name;
 
                                 // TODO [issue #423]: This network transmission needs to be reliable: the worker may crash-recover.
-                                let message = PrimaryWorkerMessage::Synchronize(digests, author.clone());
+                                let message = WorkerSynchronizeMessage{digests, target: author.clone()};
                                 let _ = self.network.unreliable_send(worker_name, &message);
                             }
                         }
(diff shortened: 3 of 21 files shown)


primary/src/block_synchronizer/mod.rs
```diff
@@ -896,8 +896,10 @@ impl BlockSynchronizer {
                 .expect("Worker id not found")
                 .name;
```
gdanezis, line 898: Put the debug! log here and no need to clone the batch_ids and primary_peer_name any more.

network/src/p2p.rs
```diff
@@ -240,6 +240,20 @@ impl ReliableNetwork<PrimaryWorkerMessage> for P2pNetwork {
...
 
+impl UnreliableNetwork<WorkerSynchronizeMessage> for P2pNetwork {
+    type Response = ();
+    fn unreliable_send(
+        &mut self,
+        peer: NetworkPublicKey,
+        message: &WorkerSynchronizeMessage,
+    ) -> Result<JoinHandle<Result<anemo::Response<()>>>> {
+        let message = message.to_owned();
+        let f =
```
gdanezis, line 251: Not super familiar with the anemo interface: is thif f fut triggered upon send, upon receive?

primary/src/header_waiter.rs
```diff
@@ -231,7 +231,7 @@ impl HeaderWaiter {
                                     .name;
 
                                 // TODO [issue #423]: This network transmission needs to be reliable: the worker may crash-recover.
-                                let message = PrimaryWorkerMessage::Synchronize(digests, author.clone());
+                                let message = WorkerSynchronizeMessage{digests, target: author.clone()};
```
gdanezis, line 234: Lets make a constructor function for `WorkerSynchronizeMessage` and use it here to give us flexibility.

types/src/primary.rs
```diff
@@ -714,6 +712,13 @@ pub enum PrimaryWorkerMessage {
     DeleteBatches(Vec<BatchDigest>),
 }
 
+/// Used by the primary to request that the worker sync the target missing batches.
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
gdanezis, line 716: Do we need `Clone`?

worker/src/handlers.rs
```diff
@@ -1,13 +1,163 @@
...
+        response_channel: oneshot::Sender<Result<anemo::Response<WorkerBatchResponse>>>,
+    ) {
+        let mut results: FuturesUnordered<_> = responses.into_iter().collect();
+        let mut last_result: Result<anemo::Response<WorkerBatchResponse>> = Err(anyhow::anyhow!(
+            "no WorkerBatchResponse received (or request(s) timed out)"
+        ));
+        while let Some(result) = results.next().await {
+            match result {
+                Ok(Ok(Ok(result))) => {
+                    let _ = response_channel.send(Ok(result));
```
gdanezis, line 99: So here it seems we respond with the first result. But what happens to the other Tasks? I supose they keep running since they are not cancelled. Is that what we want or not? Since we have their join handle we can call https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.abort is this desirable?

worker/src/handlers.rs
```diff
@@ -1,13 +1,163 @@
...
+                                let _ = response_channel.send(Err(e));
+                            },
+                        }
+                    } else if let Some(num_nodes) = num_nodes {
+                        let names = self.worker_cache.load()
+                            .others_workers(&self.name, &self.id)
+                            .into_iter()
+                            .map(|(_, info)| info.name)
+                            .collect();
+                        let handles: Vec<_> = self.network.lucky_broadcast(names, &request, num_nodes)
```
gdanezis, line 133: Do we want to give an int `num_nodes`, or a list of them (since we usually request a batch given a cert with a list of signers).

worker/src/handlers.rs
```diff
@@ -73,4 +240,140 @@ impl PrimaryToWorker for PrimaryReceiverHandler {
...
+                }
+                Err(e) => {
+                    error!("{e}");
+                    continue;
+                }
+            };
+        }
+
+        // reply back immediately for the available ones
+        if !available.is_empty() {
```
gdanezis, line 272: Do we need to check this condition? the `for` loop will just do nothing if available is empty right?

worker/src/handlers.rs
```diff
@@ -73,4 +240,140 @@ impl PrimaryToWorker for PrimaryReceiverHandler {
...
+            digests: missing.iter().cloned().collect(),
+        };
+        self.tx_request_batches_rpc
+            .send((Some(worker_name), None, message, tx_batch_response))
+            .await
+            .map_err(|e| anemo::rpc::Status::internal(e.to_string()))?;
+        let response = rx_batch_response
+            .await
+            .map_err(|e| anemo::rpc::Status::internal(e.to_string()))?;
+        if let Ok(response) = response {
```
gdanezis, line 320: Should we at least log an Err response?

gdanezis (commented):
Great stuff -- I am not super familiar (not at all!) with the Anemo interface, so I will leave others to give this a final stamp.

network/src/p2p.rs
aschran: This is the send side

primary/src/block_synchronizer/mod.rs
aschran: I kept it in the same location but made it reference the field in `message` so that we could stil log "Sent" instead of "Sending"

`primary_peer_name` still needs the clone because it's a function argument that gets reused multiple times in the loop

primary/src/header_waiter.rs
aschran: I think we should have the same convention for all of the RPC request/response messages. Right now none of them have constructors. I don't have a strong opinion either way but maybe worth a discussion and then we could change them all together in a separate PR if desired. @gdanezis what do you think the benefits are? @bmwill thoughts?

types/src/primary.rs
aschran: Yes we do, because of https://github.com/MystenLabs/narwhal/blob/aschran/nwproto/network/src/traits.rs#L10

worker/src/handlers.rs
aschran: Good point, this is code moved without modification from synchronizer but maybe the original reason for this check no longer applies

worker/src/handlers.rs
aschran: So this RPC should be expected to occasionally fail since it's sent unreliably, other nodes may be byzantine, etc. RPC errors will at least be counted in the new anemo rpc prometheus metrics. If the whole `Synchronize` fails, the primary will eventually log an error about that. Given all that do you think an additional error log here is still useful or too spammy? I am on the fence myself, so if you ask me to put it I will happily do it but at least wanted to explain the context

worker/src/handlers.rs
aschran: This preserves the existing behavior from synchronizer, which chooses randomly among all other workers. I tried to keep same behavior as much as possible since this PR is already huge; hoping it'll be easier to make additional business logic tweaks separate from the big refactoring of the call pattern

worker/src/handlers.rs
```diff
@@ -1,13 +1,163 @@
...
+    // Timeout on RequestBatches RPC.
+    request_batches_timeout: Duration,
+    // Incoming RequestBatches requests to be sent out. Uses `unreliable_send` if a target is
+    // provided, or `lucky_broadcast` if a target node count is provided.
+    rx_request_batches_rpc: Receiver<(
+        Option<NetworkPublicKey>,
+        Option<usize>,
+        WorkerBatchRequest,
+        oneshot::Sender<Result<anemo::Response<WorkerBatchResponse>>>,
+    )>,
```
bmwill, line 50: nit (feel free to ignore): I think it would be beneficial to take this tuple and convert it to some named struct to make this a bit easier to wield.

worker/src/handlers.rs
```diff
@@ -1,13 +1,163 @@
...
+        let mut inflight_requests = FuturesUnordered::new();
+        loop {
+            tokio::select! {
+                Some((target, num_nodes, request, response_channel)) = self.rx_request_batches_rpc.recv() => {
+                    if target.is_some() && num_nodes.is_some() {
+                        panic!("cannot set both target and num_nodes");
+                    }
+                    if let Some(target) = target {
+                        match self.network.unreliable_send(target, &request) {
+                            Ok(handle) => {
```
bmwill, line 119: Could we break this handler arm out into a function for readability? (there's lots of indentation as well as rustfmt tends to break inside of macros)

worker/src/handlers.rs
aschran: Yes good point. I avoided getting too fancy here since we want to delete `ChildRpcHandler` as soon as anemo supports child rpcs sent from handler functions

worker/src/handlers.rs
aschran: Per chat, agree it'd be better that way, but maybe skip this work if we'll be able to delete `ChildRpcSender` soonish?

worker/src/handlers.rs
aschran: Per chat, agree it'd be better that way, but maybe skip this work if we'll be able to delete `ChildRpcSender` soonish?

primary/src/header_waiter.rs
gdanezis: No strong feelings, just a suggestion. In my experience its harder to refactor when we init structures directly, but its a case by case thing.

worker/src/handlers.rs
gdanezis: I initially err on the side of logging more (trace level if you prefer), and then we can cut down if we want.

bmwill (approved):
There's a few things that'll we'll be able to improve once anemo supports child rpcs, till then I'm happy with this and I didn't see any major blockers from George.

worker/src/handlers.rs
aschran: Ok added!

PR #1036, opened 2022-09-28
network: add metrics and logging layers

bmwill:
Add metrics and logging layers to both inbound and outbound network requests.

### DIFF
--- a/network/src/metrics.rs
+++ b/network/src/metrics.rs
@@ -1,155 +1,180 @@
 // Copyright (c) 2022, Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
-use prometheus::{default_registry, register_int_gauge_vec_with_registry, IntGaugeVec, Registry};
+use anemo_tower::callback::MakeCallbackHandler;
+use anemo_tower::callback::ResponseHandler;
+use prometheus::HistogramTimer;
+use prometheus::{
+    register_histogram_vec_with_registry, register_int_counter_vec_with_registry,
+    register_int_gauge_vec_with_registry, IntGaugeVec, Registry,
+};
+use prometheus::{HistogramVec, IntCounterVec};
 use std::sync::Arc;
 
-pub trait NetworkMetrics {
-    fn network_available_tasks(&self) -> &IntGaugeVec;
+#[derive(Clone)]
+pub struct NetworkMetrics {
+    /// Counter of requests by route
+    requests: IntCounterVec,
+    /// Request latency by route
+    request_latency: HistogramVec,
+    /// Request size by route
+    request_size: HistogramVec,
+    /// Response size by route
+    response_size: HistogramVec,
+    /// Gauge of the number of inflight requests at any given time by route
+    inflight_requests: IntGaugeVec,
+    /// Failed requests by route
+    errors: IntCounterVec,
 }
 
-#[derive(Clone, Debug)]
-pub struct PrimaryNetworkMetrics {
-    /// The number of executor available tasks
-    pub network_available_tasks: IntGaugeVec,
-}
+const LATENCY_SEC_BUCKETS: &[f64] = &[
+    0.001, 0.005, 0.01, 0.05, 0.1, 0.25, 0.5, 1., 2.5, 5., 10., 20., 30., 60., 90.,
+];
 
-impl PrimaryNetworkMetrics {
-    pub fn new(registry: &Registry) -> Self {
+impl NetworkMetrics {
+    pub fn new(node: &'static str, direction: &'static str, registry: &Registry) -> Self {
+        // Buckets from 1kb to 8mb by powers of 2
+        let size_byte_buckets = prometheus::exponential_buckets(1024.0, 2.0, 15).unwrap();
+
+        let requests = register_int_counter_vec_with_registry!(
+            format!("{node}_{direction}_requests"),
+            "The number of requests made on the network",
+            &["route"],
+            registry
+        )
+        .unwrap();
+
+        let request_latency = register_histogram_vec_with_registry!(
+            format!("{node}_{direction}_request_latency"),
+            "Latency of a request by route",
+            &["route"],
+            LATENCY_SEC_BUCKETS.to_vec(),
+            registry,
+        )
+        .unwrap();
+
+        let request_size = register_histogram_vec_with_registry!(
+            format!("{node}_{direction}_request_size"),
+            "Size of a request by route",
+            &["route"],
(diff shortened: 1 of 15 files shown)


network/src/metrics.rs
```diff
@@ -1,155 +1,150 @@
...
+};
+use prometheus::{HistogramVec, IntCounterVec};
 use std::sync::Arc;
-
-pub trait NetworkMetrics {
-    fn network_available_tasks(&self) -> &IntGaugeVec;
+use std::time::Instant;
+
+#[derive(Clone)]
+pub struct NetworkMetrics1 {
```
aschran, line 14: why the 1 suffix?

network/src/metrics.rs
```diff
@@ -1,155 +1,150 @@
...
+    /// Counter of requests, route is a label (ie separate timeseries per route)
+    requests: IntCounterVec,
+    /// Request latency, route is a label
+    request_latency: HistogramVec,
+    concurrent_requests: IntGaugeVec,
```
aschran, line 19: Instead of the name `concurrent`, how about `inflight` or `pending`?

the meaning of `concurrent` is less obvious to me. Like, does that mean the maximum number that the stack will allow to be concurrent? The highest concurrent number we have ever seen? etc.

network/src/metrics.rs
```diff
@@ -1,155 +1,150 @@
...
+const LATENCY_SEC_BUCKETS: &[f64] = &[
+    0.001, 0.005, 0.01, 0.05, 0.1, 0.25, 0.5, 1., 2.5, 5., 10., 20., 30., 60., 90.,
+];
+
+impl NetworkMetrics1 {
+    pub fn new(node: &'static str, registry: &Registry) -> Self {
+        let requests = register_int_counter_vec_with_registry!(
+            format!("{node}_requests"),
+            "The number of anemo requests made on the network",
+            &["direction", "route"],
```
aschran, line 33: Instead of making direction a label for these metrics, I think I'd sooner have entirely separate metrics for inbound vs outbound? I can't imagine a case where you would want to aggregate these metrics regardless of direction, and in fact it might not even be comparable

For example, with request latency: outbound latency includes the network latency, while inbound latency doesn't. So these are almost like two different units, and it would not only be not-useful but actually incorrect to apply aggregation functions to the union of those

network/src/metrics.rs
```diff
@@ -1,155 +1,150 @@
...
-
-pub trait NetworkMetrics {
-    fn network_available_tasks(&self) -> &IntGaugeVec;
+use std::time::Instant;
+
+#[derive(Clone)]
+pub struct NetworkMetrics1 {
+    /// Counter of requests, route is a label (ie separate timeseries per route)
+    requests: IntCounterVec,
+    /// Request latency, route is a label
```
aschran, line 17: Maybe harmonize the format/descriptiveness of these comments? Or just delete them? The metrics are self documenting in the descriptions below anyway, so partially duplicating those descriptions here is maybe liable to get outdated/misleading

network/src/metrics.rs
```diff
@@ -1,155 +1,150 @@
...
+        let concurrent_requests = register_int_gauge_vec_with_registry!(
+            format!("{node}_concurrent_requests"),
+            "The number of concurrent network requests",
+            &["direction", "route"],
+            registry
+        )
+        .unwrap();
+
+        let errors = register_int_counter_vec_with_registry!(
+            format!("{node}_errors"),
```
aschran, line 56: `{node}_request_errors`? Just cause I can imagine that name being too generic and confusing with respect to other error metrics we might have?

network/src/metrics.rs
```diff
@@ -1,155 +1,150 @@
...
+#[derive(Clone)]
+pub struct NetworkMetrics1 {
+    /// Counter of requests, route is a label (ie separate timeseries per route)
+    requests: IntCounterVec,
+    /// Request latency, route is a label
+    request_latency: HistogramVec,
+    concurrent_requests: IntGaugeVec,
+    /// Failed requests by route
+    errors: IntCounterVec,
 }
```
aschran, line 22: doesn't have to be in this PR but I think request and response sizes in bytes would also be pretty useful

primary/src/primary.rs
```diff
@@ -207,10 +208,28 @@ impl Primary {
         let routes = anemo::Router::new()
             .add_rpc_service(primary_service)
             .add_rpc_service(worker_service);
+
+        let service = ServiceBuilder::new()
+            .layer(TraceLayer::new())
```
aschran, line 213: Is it possible/do you think it's worthwhile to add a `MystenNetworkLayer` type of thing that itself includes all of: `TraceLayer`, the metrics `CallbackLayer`, and ownership of the actual metrics struct? This means future changes/additions to our standard suite of layers can go just in that one place instead of many places, plus reduces some boilerplate in these main functions.

aschran (changes requested):
Thanks, so excited to have this!

network/src/metrics.rs
bmwill: agreed, either of those are better, more descriptive names

network/src/metrics.rs
bmwill: I was trying to avoid needing to define two completely different sets of metrics for inbound/outbound for code duplications sake since they'd be easy enough to separate out on the dashboards after the fact

network/src/metrics.rs
bmwill: This is specifically rustdoc comment and always prefer more doc comments

network/src/metrics.rs
bmwill: good catch

network/src/metrics.rs
bmwill: what sort of dimensionality are you interested in? Count of bytes sent/received?

primary/src/primary.rs
bmwill: Yes we could probably look at creating a simple function that returns the layer stack we need, see here for an example of how we could do that: https://docs.rs/tower/latest/tower/util/struct.BoxLayer.html

network/src/metrics.rs
aschran: For posterity: histogram per our slack discussion

bmwill: discussed offline that this should be a hist

primary/src/primary.rs
bmwill: But i'm going to leave this to be done at a future point in time.

aschran (approved):

PR #1037, opened 2022-09-29
recover primary core on startup

lanvidr:
Replaces #817 
In progress: needs testing and metrics


mwtian:
I believe we have merged the completed PR in the sui repo.

